PR #33, opened 2021-09-09
Add missing router methods

fkettelhoit:
This PR mostly adds sync and async router fns for the missing HTTP methods HEAD, PUT, PATCH, DELETE, OPTIONS, CONNECT and TRACE.

I originally added a `with_status_code` setter to `Response` in my fork, but later saw that you have since added a nearly identical method, so I merged the two commits and kept `with_status` as the name from the main branch (though perhaps `with_status_code` should be used to be consistent with the existing getter). Let me know if you would prefer to have a separate PR for the `with_status` return value type change.

### DIFF
--- a/worker/src/response.rs
+++ b/worker/src/response.rs
@@ -107,7 +107,7 @@ impl Response {
         })
     }
 
-    /// Set the HTTP Status code on this `Response`.
+    /// Get the HTTP Status code of this `Response`.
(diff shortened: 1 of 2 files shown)


nilslice:
Thanks @fkettelhoit - I've merged this into a testing branch and will get back to you if there are any changes needed. I think it's a great addition and appreciate the help! This will likely be out with the next release.

PR #34, opened 2021-09-09
Router avoid method conflicts

fkettelhoit:
Depends on #33 

I noticed that I could not add a catchall route `/*whatever` without causing a conflict with an existing route `/something`, even if the 2 routes used different HTTP methods. I am not sure whether this is expected behavior, but I thought that it might make sense to check conflicts only for routes with the same HTTP methods, which is what this PR implements.

For example, with this PR it is possible to define a `/*whatever` route for OPTIONS requests without causing conflicts with any existing and more specific GET route.

### DIFF
--- a/worker/src/response.rs
+++ b/worker/src/response.rs
@@ -107,7 +107,7 @@ impl Response {
         })
     }
 
-    /// Set the HTTP Status code on this `Response`.
+    /// Get the HTTP Status code of this `Response`.
     pub fn status_code(&self) -> u16 {
         self.status_code
     }
--- a/worker/src/response.rs
+++ b/worker/src/response.rs
@@ -154,9 +154,17 @@ impl Response {
     }
 
     /// Set this response's status code.
-    pub fn with_status(mut self, status_code: u16) -> Self {
+    ///
+    /// Will return Err if the status code provided is outside the valid HTTP
+    /// error range of 100-599.
+    pub fn with_status(mut self, status_code: u16) -> Result<Self> {
+        if !(100..=599).contains(&status_code) {
+            return Err(Error::Internal(
+                "provided error status code is invalid".into(),
+            ));
+        }
         self.status_code = status_code;
-        self
+        Ok(self)
     }
 
     /// Read the `Headers` on this response.
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -36,11 +36,9 @@ impl<D> Clone for Handler<'_, D> {
     }
 }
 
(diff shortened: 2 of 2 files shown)


nilslice:
Thanks @fkettelhoit! this is great, and nice catch with the route issue. I'll give this and your other PR a review ASAP.

nilslice:
@fkettelhoit  - 

testing this on branch `merge/pr-34`, I observe this panic: 

```
panicked at 'failed to register Options route for /async pattern: Conflict { with: "/async" }', ~/workers-rs/worker/src/router.rs:271:18
```

Mind taking a look?

fkettelhoit:
I haven't deployed the sandbox worker yet, but shouldn't this be the expected behavior? In [line 235](https://github.com/cloudflare/workers-rs/blob/a21ef017622c344c1476978d7b1868233faed4f6/worker-sandbox/src/lib.rs#L235) the worker uses an `on_async`, which will match all methods, including OPTIONS and thus clash with the `options_async` in [line 347](https://github.com/cloudflare/workers-rs/blob/a21ef017622c344c1476978d7b1868233faed4f6/worker-sandbox/src/lib.rs#L347).

(Btw I also noticed that the `with_status` method in this branch uses a `Result<Self>` signature again, despite your previous commit. Sorry about the signature change btw, you are right that the `Self` return type makes more sense for mapping responses.)

nilslice:
> I haven't deployed the sandbox worker yet, but shouldn't this be the expected behavior? In line 235 the worker uses an on_async, which will match all methods, including OPTIONS and thus clash with the options_async in line 347.

Ah, yes! thank you. Definitely working as expected 😆 

> (Btw I also noticed that the with_status method in this branch uses a Result<Self> signature again, despite your previous commit. Sorry about the signature change btw, you are right that the Self return type makes more sense for mapping responses.)

Yea I need to rebase this. Thanks! No worries about the signature change.. it's really the _right_ thing to do so we can catch invalid codes.. Do you think it's worth the trade-off? My concern is that it's pretty unlikely that the end-user will defensively attempt to check this error, since `Response` is generally at the very end of the block... very open to input on it though!

fkettelhoit:
I was initially a bit torn myself, but I agree that choosing `Self` is the better trade-off. Situations where handling the error instead of panicking makes a difference should be pretty rare and `Result<Self>` is definitely much more inconvenient to use.

nilslice:
Ok, cool. Thank you for the feedback. I'll leave it as-is then and we can reconsider later on!

PR #42, opened 2021-09-12
Fix two compiler errors about iterating arrays and remove one unneeded unsafe block

gnp:
While trying the Getting Started steps I had to locally fix these issues to continue because the latest Rust compiler reports errors on building worker-build. The fixes are simple and after making them I was able to successfully run the example, make changes to the worker and see those changes applied.

### DIFF
--- a/worker-build/src/main.rs
+++ b/worker-build/src/main.rs
@@ -151,7 +151,7 @@ export default {{ fetch }};
         (shim_content, shim_path),
     ].iter() {
         let mut file = File::create(path)?;
-        let buf = unsafe {
+        let buf = {
             // SAFETY: this string contains valid UTF-8, and will continue to do so
             // until the write is complete. So this operation is safe.
             content.as_bytes()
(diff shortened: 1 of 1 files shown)


nilslice:
@gnp - thanks for this! The build program is valid in recent versions of Rust. Out of curiosity, are you running on the latest version of Rust when you encounter these errors? If not, which version are you using? (`rustc --version`)

We still have not decided on the minimum supported version of `rustc` for this crate, but my guess is you're using an older version.

gnp:
This is the version I'm using:

```rustc --version
rustc 1.52.1 (9bc8c42bb 2021-05-09)```

jyn514:
`IntoIterator` for arrays was stabilized in 1.53: https://github.com/rust-lang/rust/blob/master/RELEASES.md#libraries-2

nilslice:
Thanks, @jyn514! 

So, @gnp, I think we will at least start with 1.53 as a minimum supported version. 

I will review the unsafe usage and get back to you. In the meantime, if you would like to still keep this PR active, please upgrade your Rust version and push a new commit without the addition of the `.iter()` call on the arrays. Much appreciated :)

nilslice:
Actually, I will go ahead and try to modify your fork. Currently working on this now and can probably handle it. I'll push to your branch so you maintain contributor credit here.

gnp:
Thanks!

PR #44, opened 2021-09-13
remove unsafe from worker-build command

nilslice:


caass (approved):

worker-build/src/main.rs
```diff
@@ -104,10 +104,7 @@ fn copy_generated_code_to_worker_dir() -> Result<()> {
 }
 
 fn write_worker_shims_to_worker_dir() -> Result<()> {
-    // write some shims to make our wasm worker-compatible
-    // this is handled by higher-level crates automatically
```
jyn514, line 108: Why remove this comment?

worker-build/src/main.rs
```diff
@@ -178,22 +169,14 @@ fn replace_generated_import_with_custom_impl() -> Result<()> {
...
     let mut file = File::open(path)?;
-    let mut contents = String::with_capacity(file_size);
-    let buf = unsafe {
-        // SAFETY: these bytes are valid UTF-8, and we don't mutate them
-        // while reading into them.
-        contents.as_mut_vec()
-    };
-    file.read_to_end(buf)?;
-    Ok(contents)
+    let mut buf: Vec<u8> = Vec::with_capacity(file_size);
```
jyn514, line 172:
```suggestion
    let mut buf = Vec::with_capacity(file_size);
```

jyn514 (commented):
LGTM other than the nits :)

worker-build/src/main.rs
nilslice: it could probably be expanded upon rather than removed.. @caass care to add some detail? 

primarily, the first comment is about the same as the fn name, but definitely could go further into detail about what compatibility means, and what the shim needs to do in order to bridge the js/wasm code. 

the second comment might be more helpful if it's expanded to contain the names of the crates, or what "this" is. are we talking about the Rust -> wasm compilation, or the codegen from wasm_bindgen to do memory stuff, or js module stuff?

nilslice: going to merge as-is and if we want to clarify later we can open a new PR.

PR #46, opened 2021-09-14
Merge PR for #34, add or_else_any_method explicit handlers

nilslice:
This PR builds on #34, adding explicit route handler for non-match patterns using the `not_found` and `not_found_async` methods.

It also modifies the `worker-sandbox` Worker to narrow its match handlers to more specific HTTP methods to allow for better testing of wildcard patterns and `not_found` handler.

### DIFF
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -39,6 +39,7 @@ impl<D> Clone for Handler<'_, D> {
 /// A path-based HTTP router supporting exact-match or wildcard placeholders and shared data.
 pub struct Router<'a, D> {
     handlers: HashMap<Method, Node<Handler<'a, D>>>,
+    not_found_handler: Option<Handler<'a, D>>,
     data: Option<D>,
 }
 
@@ -94,6 +95,7 @@ impl<'a, D: 'static> Router<'a, D> {
     pub fn new(data: D) -> Self {
         Self {
             handlers: HashMap::new(),
+            not_found_handler: None,
             data: Some(data),
         }
     }
@@ -143,6 +145,13 @@ impl<'a, D: 'static> Router<'a, D> {
     /// Register an HTTP handler that will respond to any requests.
     pub fn on(mut self, pattern: &str, func: HandlerFn<D>) -> Self {
         self.add_handler(pattern, Handler::Sync(func), Method::all());
+        self
+    }
+
+    /// Register an HTTP handler that will catch any route that has no matched pattern for any
+    /// method. Any wildcard route will override this.
+    pub fn not_found(mut self, func: HandlerFn<D>) -> Self {
+        self.not_found_handler = Some(Handler::Sync(func));
         self
     }
 
@@ -262,6 +271,19 @@ impl<'a, D: 'static> Router<'a, D> {
         self
     }
 
+    /// Register an HTTP handler that will catch any route that has no matched pattern for any
+    /// method. Any wildcard route will override this. Enables the use of `async/await` syntax in
+    /// the callback.
+    pub fn not_found_async<T>(mut self, func: fn(Request, RouteContext<D>) -> T) -> Self
+    where
+        T: Future<Output = Result<Response>> + 'static,
+    {
+        self.not_found_handler = Some(Handler::Async(Rc::new(move |req, route| {
+            Box::pin(func(req, route))
+        })));
+        self
+    }
+
     fn add_handler(&mut self, pattern: &str, func: Handler<'a, D>, methods: Vec<Method>) {
         for method in methods {
             self.handlers
@@ -277,10 +299,10 @@ impl<'a, D: 'static> Router<'a, D> {
 
     /// Handle the request provided to the `Router` and return a `Future`.
     pub async fn run(self, req: Request, env: Env) -> Result<Response> {
-        let (handlers, data) = self.split();
+        let (handlers, data, not_found_handler) = self.split();
 
         if let Some(handlers) = handlers.get(&req.method()) {
-            if let Ok(Match { value, params }) = handlers.at(&req.path()) {
+            if let Ok(Match { value, params }) = &handlers.at(&req.path()) {
                 let mut par: RouteParams = HashMap::new();
                 for (ident, value) in params.iter() {
                     par.insert(ident.into(), value.into());
@@ -296,6 +318,7 @@ impl<'a, D: 'static> Router<'a, D> {
                 };
             }
         }
+
         for method in Method::all() {
             if method == Method::Head || method == Method::Options || method == Method::Trace {
                 continue;
@@ -306,6 +329,21 @@ impl<'a, D: 'static> Router<'a, D> {
                 }
             }
         }
+
+        if let Some(handler) = not_found_handler {
+            let route_info = RouteContext {
+                data,
+                env,
+                params: HashMap::new(),
+            };
+            return match handler {
+                Handler::Sync(func) => (func)(req, route_info).map(|resp| resp.with_status(404)),
+                Handler::Async(func) => (func)(req, route_info)
+                    .await
+                    .map(|resp| resp.with_status(404)),
+            };
+        }
+
         Response::error("Not Found", 404)
     }
 }
@@ -313,16 +351,13 @@ impl<'a, D: 'static> Router<'a, D> {
 type NodeWithHandlers<'a, D> = Node<Handler<'a, D>>;
 
 impl<'a, D: 'static> Router<'a, D> {
-    fn split(self) -> (HashMap<Method, NodeWithHandlers<'a, D>>, Option<D>) {
-        (self.handlers, self.data)
-    }
-}
-
-impl<D> Default for Router<'_, D> {
-    fn default() -> Self {
-        Self {
-            handlers: HashMap::new(),
-            data: None,
-        }
+    fn split(
+        self,
+    ) -> (
+        HashMap<Method, NodeWithHandlers<'a, D>>,
+        Option<D>,
+        Option<Handler<'a, D>>,
+    ) {
+        (self.handlers, self.data, self.not_found_handler)
     }
 }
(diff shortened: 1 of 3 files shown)


nilslice:
@fkettelhoit - if possible, it would be great to get your review on this to make sure it captures all of what you were aiming to have supported with your PR.

worker/src/router.rs
```diff
@@ -264,67 +271,93 @@ impl<'a, D: 'static> Router<'a, D> {
...
-            ));
+        for method in methods {
+            self.handlers
+                .entry(method.clone())
+                .or_insert_with(Node::new)
+                .insert(pattern, func.clone())
+                .expect(&format!(
+                    "failed to register {:?} route for {} pattern",
+                    method, pattern
+                ));
```
jyn514, line 296: This allocates on each iteration of the loop. I'd suggest one of the following:
1. Move the allocation outside the loop; this would also allow you to remove `method.clone()`.
2. Use `unwrap_or_else(|| panic!(...))`.

2 is probably more efficient since `Method` is really cheap to clone.

worker/src/router.rs
```diff
@@ -264,67 +271,93 @@ impl<'a, D: 'static> Router<'a, D> {
...
         }
+
         Response::error("Not Found", 404)
     }
 }
 
-type NodeWithHandlers<'a, D> = Node<[Option<Handler<'a, D>>; 9]>;
+type NodeWithHandlers<'a, D> = Node<Handler<'a, D>>;
 
 impl<'a, D: 'static> Router<'a, D> {
```
jyn514, line 353: Hmm, why does this have a `'static` bound?

nilslice: The `Handler::Async` enum variant keeps hold of a `fn(Request, RouteContext<D>` and otherwise complains about `D` maybe not living long enough. Feel free to take a closer look and see if there is a way around it!

nilslice: maybe it could just be `'a` though... will test it out

nilslice: nope. still yelled at :) not my top priority to change this up right now, but again feel free to re-work it if you have time and want to think about a different approach

jyn514: Hmm, how are you testing this? I ran `cargo check --all-targets` with this diff and it seemed to compile ok:
```
diff --git a/worker/src/router.rs b/worker/src/router.rs
index c26ba54..67957cd 100644
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -350,7 +350,7 @@ impl<'a, D: 'static> Router<'a, D> {
 
 type NodeWithHandlers<'a, D> = Node<Handler<'a, D>>;
 
-impl<'a, D: 'static> Router<'a, D> {
+impl<'a, D> Router<'a, D> {
     fn split(
         self,
     ) -> (
```

jyn514: In fact the rest of the API doesn't need to require 'static either:
```diff
diff --git a/worker/src/router.rs b/worker/src/router.rs
index c26ba54..6bbdb82 100644
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -16,7 +16,7 @@ use crate::{
 
 type HandlerFn<D> = fn(Request, RouteContext<D>) -> Result<Response>;
 type AsyncHandlerFn<'a, D> =
-    Rc<dyn Fn(Request, RouteContext<D>) -> LocalBoxFuture<'a, Result<Response>>>;
+    Rc<dyn 'a + Fn(Request, RouteContext<D>) -> LocalBoxFuture<'a, Result<Response>>>;
 
 /// Represents the URL parameters parsed from the path, e.g. a route with "/user/:id" pattern would
 /// contain a single "id" key.
@@ -89,7 +89,7 @@ impl<D> RouteContext<D> {
     }
 }
 
-impl<'a, D: 'static> Router<'a, D> {
+impl<'a, D: 'a> Router<'a, D> {
     /// Construct a new `Router`, with arbitrary data that will be available to your various routes.
     /// If no data is needed, provide any valid data. The unit type `()` is a good option.
     pub fn new(data: D) -> Self {
@@ -159,7 +159,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// `async/await` syntax in the callback.
     pub fn head_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -173,7 +173,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// `async/await` syntax in the callback.
     pub fn get_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -187,7 +187,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// `async/await` syntax in the callback.
     pub fn post_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -201,7 +201,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// `async/await` syntax in the callback.
     pub fn put_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -215,7 +215,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// `async/await` syntax in the callback.
     pub fn patch_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -229,7 +229,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// of `async/await` syntax in the callback.
     pub fn delete_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -247,7 +247,7 @@ impl<'a, D: 'static> Router<'a, D> {
         func: fn(Request, RouteContext<D>) -> T,
     ) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -261,7 +261,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// syntax in the callback.
     pub fn on_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.add_handler(
             pattern,
@@ -276,7 +276,7 @@ impl<'a, D: 'static> Router<'a, D> {
     /// the callback.
     pub fn not_found_async<T>(mut self, func: fn(Request, RouteContext<D>) -> T) -> Self
     where
-        T: Future<Output = Result<Response>> + 'static,
+        T: Future<Output = Result<Response>> + 'a,
     {
         self.not_found_handler = Some(Handler::Async(Rc::new(move |req, route| {
             Box::pin(func(req, route))
@@ -350,7 +350,7 @@ impl<'a, D: 'static> Router<'a, D> {
 
 type NodeWithHandlers<'a, D> = Node<Handler<'a, D>>;
 
-impl<'a, D: 'static> Router<'a, D> {
+impl<'a, D> Router<'a, D> {
     fn split(
         self,
     ) -> (
```

nilslice: Don't you still have `'static` on `@@ -261,7 +261,7 @@ impl<'a, D: 'static> Router<'a, D> {` etc.?

rust-analyzer is yelling if I remove / change `'static` to `'a`

nilslice: and cargo check giving the same errors:

```
...
...
error[E0310]: the parameter type `D` may not live long enough
   --> worker/src/router.rs:281:54
    |
92  |   impl<'a, D: 'a> Router<'a, D> {
    |            -- help: consider adding an explicit lifetime bound...: `D: 'static +`
...
281 |           self.not_found_handler = Some(Handler::Async(Rc::new(move |req, route| {
    |  ______________________________________________________^
282 | |             Box::pin(func(req, route))
283 | |         })));
    | |__________^ ...so that the type `[closure@worker/src/router.rs:281:62: 283:10]` will meet its required lifetime bounds

For more information about this error, try `rustc --explain E0310`.
error: could not compile `worker` due to 18 previous errors
➜  worker git:(merge/pr-34) ✗ rustc --version
rustc 1.55.0 (c8dfcfe04 2021-09-06)
```

fkettelhoit:
> @fkettelhoit - if possible, it would be great to get your review on this to make sure it captures all of what you were aiming to have supported with your PR.

Looks good! I was thinking a bit about how the `not_found` case might be generalized, the result is PR #47. But I am honestly not sure whether that would add significant value over the more specific version implemented here, in my opinion either would be fine.

(I think there is a larger issue at play here: How much should be handled directly in a single router vs. composing multiple routers? Thanks to the functional nature of the router, it is easy to map the result of one router and use another router for any 404 response, for example, which can already solve most catchall cases. But there is still value in providing an API that solves most common situations in a single router, in my opinion, because compositions of different routers are not always the most obvious solution. I would say that all in all the current PR strikes a good balance there.)

worker/src/router.rs
nilslice: weird -- this is working now 👻

PR #47, opened 2021-09-15
Generalize `not_found` handler to handle arbitrary patterns

fkettelhoit:
(I am honestly not totally sure whether this PR adds any value compared to a more specific `not_found` handler.)

I think of the `not_found` case as a kind of "fall-through" pattern in the pattern matching logic. The current implementation of `not_found` is specific to the 404 case, however, because it assumes a catchall route and always maps the response status to a 404. This is certainly the most common case for a fall-through route, so it might make sense to prioritize this situation, but perhaps it would be useful to generalize the logic and allow for arbitrary patterns? (One example I was thinking of is a "proxy router", which only handles a few specific methods and routes, but wants to pass on all other method + route combinations to a different router or even call another worker.)

The more specific `not_found` case can still be implemented pretty easily:

```rust
        .or_else_any_method("/*catchall", |_req, _ctx| {
            let resp = ...;
            resp.map(|resp| resp.with_status(404))
        })
```

I'm not totally happy with the naming, `or_else_any_method` is pretty verbose, but I personally like the `or_else_...` prefix because it emphasizes that the handler will be called only if no other pattern matches, even if it is defined in the code before a "normal" handler.

(I also included @jyn514 's lifetime suggestions for convenience here.)

### DIFF
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -16,7 +16,7 @@ use crate::{
 
 type HandlerFn<D> = fn(Request, RouteContext<D>) -> Result<Response>;
 type AsyncHandlerFn<'a, D> =
-    Rc<dyn Fn(Request, RouteContext<D>) -> LocalBoxFuture<'a, Result<Response>>>;
+    Rc<dyn 'a + Fn(Request, RouteContext<D>) -> LocalBoxFuture<'a, Result<Response>>>;
 
 /// Represents the URL parameters parsed from the path, e.g. a route with "/user/:id" pattern would
 /// contain a single "id" key.
@@ -89,7 +89,7 @@ impl<D> RouteContext<D> {
     }
 }
 
-impl<'a, D: 'static> Router<'a, D> {
+impl<'a, D: 'a> Router<'a, D> {
     /// Construct a new `Router`, with arbitrary data that will be available to your various routes.
     /// If no data is needed, provide any valid data. The unit type `()` is a good option.
     pub fn new(data: D) -> Self {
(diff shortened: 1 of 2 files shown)


nilslice:
haha why is my rust so mad at me about these lifetimes! need to do a cargo clean, restart, and maybe wipe my whole disk 😆

worker/src/router.rs
```diff
@@ -332,17 +340,15 @@ impl<'a, D: 'static> Router<'a, D> {
             }
         }
 
-        if let Some(handler) = not_found_handler {
+        if let Ok(Match { value, .. }) = or_else_any_method_handler.at(&req.path()) {
             let route_info = RouteContext {
                 data,
                 env,
                 params: HashMap::new(),
```
nilslice, line 347: I think we'd still want to collect the params from the `Match` into the `RouteContext` so that a user can easily access them from the `ctx.param` func - right? 

I could see a scenario where someone might want to handle the `or_else_any_method` case differently based on the route params. 

Curious to know if you had other thoughts and specific reasoning to not pass them through and create an empty `HashMap` instead?

nilslice:
@fkettelhoit - I like this! I may add that minor change to include the `params` as mentioned in my comment above, and we can revert later. I don't expect you to respond beforehand given the time difference. FWIW, I agree about the naming on this, but it is pretty clear about the intention there, and does follow some familiar Rust naming.. so let's keep it for now . 

Thank you!

nilslice (approved):

worker/src/router.rs
fkettelhoit: You're absolutely right, that was just an oversight on my part. I wanted everything to compile with minimal changes first and then completely forgot to pass the params!

I fixed it on my branch in case you want to take the commit from there, but of course feel free to just do it here otherwise.

nilslice: Ok cool - yea, I sort of figured as much :) 

I have a `From` impl on `matchit::Params` for the `RouteContext` which I also changed to be a newtype over the HashMap .. so I'll probably just carry on with the changes I have, but thank you for updating so quickly!

PR #56, opened 2021-09-21
area(worker/response): add support for redirect() method

### LINKED ISSUE #55: No Implementation of `redirect()` for `Response`
Hey There! :wave: 

Was just exploring the use of worker-rs where i eventually figured out that there is no `redirect()` method implementation for the `Response` struct, which means that i cannot redirect to a URL (302,308,etc) for an response recieved. It also becomes inconsistent with the documentation found at https://developers.cloudflare.com/workers/runtime-apis/response#instance-methods where it specifies an instance method for `redirect()`. 

**Proposal:**
Implement a function in the `Impl` of Response struct with the apt headers, and struct fields which becomes available for the user to use

Would be happy to raise a PR for the same :smiley:

DarthBenro008:
- Fixes #55 

This PR adds support for the redirect of responses and adds two new methods.

- The `Response::redirect(url)` method that takes in an URL and redirects with a default status code of `302`
```rust
Response::redirect("https://github.com")
```

- The `Response::redirect_with_status_code(url, status_code)` method that redirects with a custom status code
```rust
Response::redirect_with_status_code("https://github.com", 308)
```

The relevant FFI binding can be found here:

https://github.com/cloudflare/workers-rs/blob/1bc536dcf1d397a0dc3087a021a28d60c9c85d66/worker-sys/src/response.rs#L88-L94

- [x] The PR has been tested in a playground project generated by wrangler
- [x] Code has been linted
- [x] Doc strings have been added


nilslice:
@DarthBenro008 that's great, thank you for taking this on! Couple of things... 

1. Since the `redirect` method in the runtime expects an absolute URL (as far as I can tell relative paths won't work.. e.g. `/admin/logout`), I think we should consider changing the Rust helper method on `Response` to accept a `url::Url` param rather than the `impl Into<String>`. This way the Rust type system with enforce the requirement that the input to the method is a valid URL.  What do you think? 

- if you think we _should_ accept relative paths (since this is more of a "framework" afterall..) then we'd need to do some Request inspection and find the URL host & scheme to reconstruct the new URL. 

2. Since we already have a `with_status` method on `Response`, I am wondering if we should omit the `redirect_with_status_code` option... I totally see it being helpful for common redirects though, so I am a bit torn. The single method call is definitely less verbose than: 

```rust
Response::redirect("https://cloudflare.com").map(|resp| resp.with_status(302))
```

Which requires the user to know to `map` the `Result<Response, E>` (which would need to be tested, I don't see why it _wouldn't_ work, but you never know until you test 😉 )

3. If we leave `_redirect_with_status_code`, let's rename it `redirect_with_status`, but also return `Err` if the status code is outside the `300-399` range, since that is invalid and the runtime might already throw an error. There is some example code in [here](https://github.com/cloudflare/workers-rs/blob/main/worker/src/response.rs#L97-L101) that might help you tackle this. 

Thank you again for this PR, and sorry for the slow response.. I got some computer time on vacation, but heading back to relaxation for a bit 😆 

cc/ @caass, @ocsfrank, @ObsidianMinor, @jyn514 or @jspspike who may have some additional input/thoughts on the above.

DarthBenro008:
@nilslice Thanks for the detailed insights!

1. I completely agree that the helper method should take in an `url::Url` to enforce valid redirects, the boon in doing this is that the Implementation of `url::Url` already splits and has helper methods to gain host, relative paths, etc that would make life easier in this case

2. I had thought of this, but then I realised that implementing `redirect_with_status_code` would be more aligned with the JS functions one would typically find in a Browser, hence I thought to implement the same, in case we use the `map | resp |` there is a possibility that the user can provide any `status_code` they want (other than in the range of 3XX). I think the `Result<Response, E>` should be only used when they are creating an empty response (`Response::empty()`) for their custom needs (like a manual override). What do you think about this?

3. Sure! I'll add the range check so that we only support 3XX based custom codes 😃

jyn514:
There are only a fixed number of possible redirect status codes, right? If you want to allow choosing between them, we could have an enum of possible redirect codes, which wouldn't allow passing anything outside the 300-399 range by accident.

Hmm, I suppose that doesn't allow you to use redirect codes that are between 300-399 but not standardized. I don't know how common that is in practice - I think requiring you to use `with_status` for those might be an ok tradeoff.

jyn514:
> 
    If we leave _redirect_with_status_code, let's rename it redirect_with_status, but also return Err if the status code is outside the 300-399 range, since that is invalid and the runtime might already throw an error. There is some example code in here that might help you tackle this.

Why check it twice? If the runtime is already checking the status I don't think we need to check it a second time; especially since Workers are perf-sensitive.

nilslice:
@jyn514 for the same reason "Response::error is a footgun" 😆 

Is there a difference or are you just asking questions?

jyn514:
@nilslice I didn't realize it was validated by the runtime - my understanding is that `Response::error` really would have worked for any status before, the method name just would have been odd. If it's checked by the runtime I don't know that it needs an explicit Result.

nilslice:
Yea, if we handle this range check elsewhere we should be consistent where applicable. The runtime just throws and exception which could be avoided if someone cares, though there isn't a whole lot to do. 

I'm not concerned about perf here.

DarthBenro008:
What I would suggest is, let's create one single function `Response::redirect(url: url::Url, status_code: Option<Set of Redirect Codes>)`

If the `status_code` is `None` we automatically wrap them with a `302`. If the user wants to pass a status code, they would be only able to pass redirect codes that's specified by the `Enum`

How does this sound? @nilslice @jyn514

jyn514:
@DarthBenro008 hmm, why make it optional? I think it's useful to make the developer choose whether it's a permanent or temporary redirect; and that way there's only one way to specify the redirect instead of two.

nilslice:
@DarthBenro008 why don't we start with `redirect_with_status` which takes the url::Url and a u16 and does the range check for valid status codes. 

Then we can consider adding the base `redirect` method later, which by default will add a 302 status code. 

This way we get your PR merged without so much deliberation and can revisit this later on 🙂

DarthBenro008:
@nilslice @jyn514 , I realised if we are expecting the user to provide `url::Url` we probably need to re-export the create so that it's consistent for the user to use the same `url::Url` as `worker-rs`. Should i include the same in this PR?

DarthBenro008:
Implemented the suggested changes, I have not included specific enums for redirect code, but a range of `3xx` to be consistent with the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) :smiley:

nilslice:
Hi @DarthBenro008 -

Thank you! Re-exporting the url::Url is a great idea. Please feel free to do so.

DarthBenro008:
@nilslice , I have re-exported `url::Url`!

worker/src/response.rs
```diff
@@ -107,6 +107,27 @@ impl Response {
...
+            Ok(edge_response) => Ok(Response::from(edge_response)),
+            Err(err) => Err(Error::from(err)),
+        }
+    }
+
+    /// Create a `Response` which redirects to the specified URL with a custom status_code
+    pub fn redirect_with_status_code(url: url::Url, status_code: u16) -> Result<Self> {
+        if !(300..=399).contains(&status_code) {
+            return Err(Error::Internal(
+                "provided error status code is invalid".into(),
```
caass, line 122: super minor nit: could this be a little more specific and say something like `"redirect status codes must be in the 300 range! see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages for more"`?

caass (changes requested):
Looks great! Thanks so much for this :) One minor nit and then I think we're ready to merge.

worker/src/response.rs
nilslice: This might be the same error string used in another "invalid status" scenario in `Response::error`, in which case we should probably change both to be more descriptive. Adding the specific range to both errors would be great.

nilslice: can we also change the name to `redirect_with_status` ? thank you!

DarthBenro008:
@caass @nilslice , Have amended the requested changes!

PR #63, opened 2021-10-04
chore: fix wasm-bindgen snippets not getting moved

### LINKED ISSUE #62: Snippets directory not moved into workers directory in build tool
While trying to use [wasm-streams](https://github.com/MattiasBuelens/wasm-streams) to implement streaming support for the [worker-kv](https://github.com/zebp/worker-kv) crate I ran across a bug where generated JS snippets aren't moved into the worker sub directory, causing all function calls that use that snippet to crash the worker. [This](https://github.com/MattiasBuelens/wasm-streams/blob/6a5cc653f862353cc1a06c79ea6685d2e145e2d2/src/readable/into_underlying_byte_source.rs#L34) is the snippet in wasm-streams that needs to be generated due to a Chromium bug.

[Here](https://github.com/zebp/inline-js-example) is an example repo with this bug.

zebp:
`wasm-bindgen` supports having JavaScript [snippets](https://rustwasm.github.io/wasm-bindgen/reference/js-snippets.html) with a library but they
were not getting moved in to the expected sub directory. This made some libraries such as `wasm-streams` unusable.

Fixes #62

### DIFF
--- a/worker-build/src/main.rs
+++ b/worker-build/src/main.rs
@@ -96,7 +96,16 @@ fn copy_generated_code_to_worker_dir() -> Result<()> {
         .join(WORKER_SUBDIR)
         .join(format!("{}_bg.wasm", OUT_NAME));
 
-    for (src, dest) in [(glue_src, glue_dest), (wasm_src, wasm_dest)] {
+    // wasm-bindgen supports adding arbitrary JavaScript for a library, so we need to move that as well.
+    // https://rustwasm.github.io/wasm-bindgen/reference/js-snippets.html
+    let snippets_src = PathBuf::from(OUT_DIR).join("snippets");
+    let snippets_dest = PathBuf::from(OUT_DIR).join(WORKER_SUBDIR).join("snippets");
+
+    for (src, dest) in [
+        (glue_src, glue_dest),
+        (wasm_src, wasm_dest),
+        (snippets_src, snippets_dest),
+    ] {
         fs::rename(src, dest)?;
     }
 


nilslice:
@zebp - this is a cool feature, and I was totally unaware of it. Would you be able to share a minimal example that I could try this out with? Any project you can publish a minimal example of that has some JS snippets mapped to Rust externs and can then run the worker-build script on?

zebp:
> @zebp - this is a cool feature, and I was totally unaware of it. Would you be able to share a minimal example that I could try this out with? Any project you can publish a minimal example of that has some JS snippets mapped to Rust externs and can then run the worker-build script on?

[This](https://github.com/zebp/inline-js-example) worker should do, requesting `/` should work with a `worker-build` compiled from the PR, but will 500 with the one from main. You will need to remove `cargo install -q worker-build` from the `build.command` in the `wrangler.toml` to prevent the `worker-build` from the PR getting overwritten.

nilslice:
Oh cool, I see. So in this case `wasm-bindgen` writes that inline JS to a file, and that file needs to be copied. What about cases like: 
```rust
#[wasm_bindgen(module = "/js/foo.js")]
extern "C" {
    fn add(a: u32, b: u32) -> u32;
}
```

does `wasm-bindgen` _also_ copy that file into the same place and thus is handled by your fix?

zebp:
> Oh cool, I see. So in this case `wasm-bindgen` writes that inline JS to a file, and that file needs to be copied. What about cases like:
> 
> ```rust
> #[wasm_bindgen(module = "/js/foo.js")]
> extern "C" {
>     fn add(a: u32, b: u32) -> u32;
> }
> ```
> 
> does `wasm-bindgen` _also_ copy that file into the same place and thus is handled by your fix?

Yes, all snippets are stored under the same snippets directory so they are copied to the correct directory. I believe this is the only other case outside of TypeScript type generation where it generates a required file for runtime.

nilslice:
@zebp Did you manage to get this working? I am testing with this simple inline snippet: 

```rust
#[wasm_bindgen(inline_js = "export function add(a, b) { return a + b; }")]
extern "C" {
    fn add(a: u32, b: u32) -> u32;
}
```
and am seeing this API error when running `wrangler dev` (after installing the new `wrangler-build` bin): 
```
Error: Something went wrong with the request to Cloudflare...
Uncaught SyntaxError: Unexpected token 'export'
  at line 1
 [API code: 10021]
```

Want to make sure you didn't get the same error before I go too deep into debug-mode 😆

PR #72, opened 2021-10-21
fix: don't rename to mjs

zebp:
Previously `worker-build` emitted JavaScript files using the `esmodule` syntax in a `.mjs` file so they could be loaded without being explicitly marked as `esmodule` in the wrangler file. This caused issues in #63 where snippets weren't renamed because import statements would have to change, so instead now we don't rename anything to `.mjs` and explicitly mark all JavaScript files in the `wrangler.toml` file as `ESModule` as shown in [this example project](https://github.com/zebp/inline-js-example/blob/esmodule/wrangler.toml#L23).

One problem is that if this version is ever published under `worker-build` any project using an outdated version of the workers crate would install the latest version when compiling, emitting an invalid output for their wrangler config. The update `worked-build` tool would already require users to change the config, so at that point we should recommend them install a pinned version of `worker-build`.

Fixes issues introduced by #63.


nilslice (approved):
thanks, testing this on `pr-63`!

PR #81, opened 2021-11-05
fix: response body access converting body as string

### LINKED ISSUE #80: Accessing response body forces string conversion
`Response::bytes()` calls `text()` when inner data is `ResponseBody::Streaming` so data gets corrupted with error message.
https://github.com/cloudflare/workers-rs/blob/1be6eaa0ee5c66431a71cbfdd839779ccf29e4c2/worker/src/response.rs#L164-L168

kiwiyou:
This fixes #80.

### DIFF
--- a/worker/src/response.rs
+++ b/worker/src/response.rs
@@ -161,9 +161,10 @@ impl Response {
         match &self.body {
             ResponseBody::Body(bytes) => Ok(bytes.clone()),
             ResponseBody::Empty => Ok(Vec::new()),
-            ResponseBody::Stream(response) => JsFuture::from(response.text()?)
+            ResponseBody::Stream(response) => JsFuture::from(response.array_buffer()?)
                 .await
-                .map(|value| value.as_string().unwrap().into_bytes())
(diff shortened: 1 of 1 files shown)


nilslice:
Hi @kiwiyou - thanks for this! Would you mind sharing how you are constructing the request which triggers this branch? using `curl` would be ideal, but if you have another way already that would be great too.

kiwiyou:
When I sent a request to https://static.solved.ac/misc/default_profile.png (CDN server of other person) worker-rs retrieves the body well. However, when I used `CfProperties` to set `cache_everything` on, `Called .text() on an HTTP body which does not appear to be text.` message appeared and the image did not render properly.

nilslice:
Hi @kiwiyou  - thank you for the follow-up. If you are ready for this to be reviewed, please update from Draft to normal PR. No rush, I just didn't want you to think it's blocked on anything else. 

I haven't taken a close look, thinking you are still working on some aspects of it. But if it's ready, I'm happy to put it in my review rotation. Thanks!

nilslice:
Thank you!

PR #87, opened 2021-11-16
Uses wasm-pack as a library

armfazh:
This makes use of the wasm-pack as a library rather than as a command.
So there is no need to check whether is already installed, since it will be as is listed as a dependency.


nilslice:
This is awesome, thank you! I will review ASAP.

nilslice:
Hi @armfazh - sorry for the delay here. Currently blocked on some decisions on how to handle further integration of wasm-pack, or the possibility of removing it altogether in favor of more direct usage of wasm-bindgen, which is wrapped by wasm-pack. Just giving you an update!

PR #88, opened 2021-11-19
Allow to debug-print requests

mlafeldt:
Which already works for responses


nilslice:
Thanks! I will add a note to add more Debug across the codebase.

Issue #89, opened 2021-11-19
Mutable Request/Response in place


nmoutschen:
Hi! 👋 

I'm currently trying to port my Cloudflare worker from TypeScript to Rust. I'm using a middleware pattern where I want pass a `worker::Request` to various functions that will transform them (e.g. transform the URL, strip certain headers, etc.) before sending to the origin. Similarly, I want to transform the `worker::Response` through multiple middlewares before sending it back to the end-user.

As far as I can, as the fields of both objects are private, I can only access those values through functions (e.g. `req.path()`, `req.headers()`), which means I can't mutate those values in place.

Is there any plan to support this type of pattern in Rust? Or is there anything I'm missing that would allow me to do that?

nilslice:
Hey @nmoutschen - 

Good question.. you can use `headers_mut(&mut self)` on the `Request` itself, a la: https://docs.rs/worker/0.0.7/worker/struct.Request.html#method.headers_mut

For the `path`, there is no current way to do this in-place; you would have to create a new Request. If you think `path_mut(&mut self)` is something you would consider implementing (use the `headers_mut` [implementation](https://docs.rs/worker/0.0.7/src/worker/request.rs.html#153-162) as reference), and try out in your project - I would be happy to consider merging a PR!

nmoutschen:
Hey @nilslice !

Happy to contribute there. I've created a PR that adds `path_mut()` to Requests: https://github.com/cloudflare/workers-rs/pull/90

nilslice:
Closed via #90. Thank you!

PR #90, opened 2021-11-19
feat: support mutable Request paths

nmoutschen:
This PR adds support for mutating a path in-place for a `Request`s using `req.path_mut()`.

See https://github.com/cloudflare/workers-rs/issues/89 for more information.


nilslice (approved):
Thanks! LGTM.

Issue #15, opened 2021-08-24
Use `log` instead of `console_log!`


jyn514:
This will mean all the logging from libraries that developers don't control will still show up in the JS console. We could implement it by registering a log handler which outputs to the JS console.

nilslice:
leaving a reminder to take a look at https://docs.rs/console_log/0.2.0/console_log/

PR #91, opened 2021-11-25
Add `console.debug`, `console.error` & `console.error`

equt:
Maybe it's better to wrap them in a `console` mod, e.g., `console::error!`

### DIFF
--- a/worker-sys/src/global.rs
+++ b/worker-sys/src/global.rs
@@ -48,5 +48,14 @@ extern "C" {
     ) -> ::js_sys::Promise;
 
     #[wasm_bindgen(js_namespace = console)]
+    pub fn debug(s: &str);
+
+    #[wasm_bindgen(js_namespace = console)]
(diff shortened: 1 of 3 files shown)


equt:
c.c. #15

nilslice (approved):
Thank you!

nilslice:
I like the idea of wrapping them in a `console` module... if you'd like to make that change, please do. Along with the change, it would be great to ensure the `console` mod is pub from `worker`, and the `worker-sandbox` & README references are updated to use the new module. 

I can do this as well, but didn't want to take it from you :) just lmk!

equt:
After some thought, I found out this is not doable, since Rust's macro always gets moved to crate root, right? So something like `console::log` will require a `console` crate.

nilslice:
Ah, yes, good catch! Ok, then I will merge this as-is, and I'll take a note to consider other interfaces for these logging calls. Thank you!

PR #95, opened 2021-11-26
refactor: Make data field of RouteContext public

gmorenz:
I specifically need this to call state.put in a durable object, I could also use a mutable getter, but there doesn't seem to be any reason why this field isn't public. It's just user supplied data after all, the framework doesn't have any expectations about how it can change. If for some reason a mutable getter (or something else) would be preferable, I'm happy to change this PR to do that instead.

I left the old getter in place for backwards compatibility, but with a deprecation warning, since using a getter when the field is directly available is unnecessary.

### DIFF
--- a/worker-sandbox/src/lib.rs
+++ b/worker-sandbox/src/lib.rs
@@ -85,7 +85,7 @@ pub async fn main(req: Request, env: Env) -> Result<Response> {
         .get_async("/async-request", handle_async_request)
         .get("/test-data", |_, ctx| {
             // just here to test data works
-            if ctx.data().regex.is_match("2014-01-01") {
+            if ctx.data.regex.is_match("2014-01-01") {
                 Response::ok("data ok")
             } else {
                 Response::error("bad match", 500)
@@ -256,7 +256,7 @@ pub async fn main(req: Request, env: Env) -> Result<Response> {
         })
         .get_async("/cloudflare-api", |_req, ctx| async move {
             let resp = ctx
-                .data()
+                .data
                 .cloudflare_api_client
                 .request_handle(&cloudflare::endpoints::user::GetUserDetails {})
                 .await
@@ -310,10 +310,7 @@ pub async fn main(req: Request, env: Env) -> Result<Response> {
             Response::from_bytes(serde_json::to_vec(&todo)?)
         })
         .post_async("/nonsense-repeat", |_, ctx| async move {
-            //  just here to test data works, and verify borrow
-            let _d = ctx.data();
-
-            if ctx.data().regex.is_match("2014-01-01") {
+            if ctx.data.regex.is_match("2014-01-01") {
                 Response::ok("data ok")
             } else {
                 Response::error("bad match", 500)
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -52,13 +52,14 @@ pub struct Router<'a, D> {
 /// Container for a route's parsed parameters, data, and environment bindings from the Runtime (such
 /// as KV Stores, Durable Objects, Variables, and Secrets).
 pub struct RouteContext<D> {
-    data: D,
+    pub data: D,
     env: Env,
     params: RouteParams,
 }
 
 impl<D> RouteContext<D> {
     /// Get a reference to the generic associated data provided to the `Router`.
+    #[deprecated(since = "0.0.8", note="please use the `data` field directly")]
     pub fn data(&self) -> &D {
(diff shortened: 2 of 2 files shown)


nilslice:
Hi @gmorenz - thanks for the PR! Could you provide a bit more context as to how this is related to `state.put` in your Worker? Even a stripped down code example would do. Just curious how you're using `data` here. I'm in favor of accepting the PR as-is, but would be great to know what you've run into. 

Thanks!

gmorenz:
Ah, this is embarrassing, I don't technically need it for that API and misread the docs to think that I did when writing up the justification for the commit (specifically I saw that storage.put takes an &mut self, and looked over the fact that state.storage returns an owned copy...).

I do need it because I wrapped that API in a caching layer, and that caching layer assumes mutable access to data. More generally I need it for mutable access to data stored in memory used by multiple endpoint handlers at all (see the comments in `async fn fetch` for the data access pattern)

I also copy and pasted the caching layer, and put an example of it being used in, though it's really just serving as an example of "and I want mutable access to in memory data". I left some TODOs in the implementation of the caching layer about potential improvements to this library because they might be of interest, but they're not related to this issue.

Will fix the commit message to be more accurate.

```rust
#[durable_object]
pub struct Chronicle {
    state: State
    some_counter: DurableValue<u32, { DurableKey("") }>,
}

#[durable_object]
impl DurableObject for Chronicle {
    fn new(state: State, env: Env)->Self {
        set_panic_hook();
        Chronicle {
            state,
            some_counter: Default::default(),
        }
    }
    
    async fn fetch(&mut self, req: Request) -> Result<Response> {
        let router = Router::with_data(self);
        router
            .get_async("/history/:idx", |_req, ctx| async move {
                 // Need mutable access to self here,  for example
                 
                let mut storage = self.state.storage();
                let pv = self.some_counter.get(&storage).await? + 1;
                self.some_counter.set(&mut storage, pv).await?;
            })
            .get_async("/", |_req, ctx| async move {
                // And also need mutable access to self here, so I can't just capture it.
            })
            .run(req, env)
            .await
    }
}

// Caching layer

/// Workaround for rustc bug in the incomplete feature
/// where bare lifetimes in the const generic break things.
#[derive(PartialEq, Eq)]
pub struct DurableKey(pub &'static str);

#[derive(Default)]
pub struct DurableValue<T, const KEY: DurableKey>
where T: Default + for <'a> Deserialize<'a> + Serialize {
    value: Option<T>
}

impl<T, const KEY: DurableKey> DurableValue<T, KEY>
where T: Default + for <'a> Deserialize<'a> + Serialize {
    pub fn get<'this, 'store>(&'this mut self, storage: &'store Storage) -> impl Future<Output=Result<&'this T>> + 'store
    where 'this: 'store {
        async move {
            // Try and use stored value
            if let Some(ref v) = self.value {
                return Ok(v);
            }

            // First run of this in-memory instance, get the pv from storage
            // TODO: The JS api has an options field, which allows disabling cloudflares cache.
            // Since we cache internally, we should pass that option.
            let val = match storage.get(KEY.0).await {
                Ok(v) => v,
                Err(Error::JsError(msg)) if msg == "No such value in storage." => {
                    // First use overall, it's the default value
                    T::default()
                },
                Err(e) => return Err(e)
            };

            // Store and return reference to value
            self.value = Some(val);
            Ok(self.value.as_ref().unwrap()) // Not None because we just set it
        }
    }

    pub async fn set(&mut self, storage: &mut Storage, val: T) -> Result<()> {
        // TODO: The JS api has an options field, which allows disabling cloudflares cache.
        // Since we cache internally, we should pass that option.

        // TODO: By ignoring the JS promise, we wou8ld avoid waiting on it to write to disk until
        // the *next* time we call an await. That might be a pretty big performance gain,
        // but TODO: We should benchmark if this matters.
        //
        // We can't do that right now, because the rust API will only call the JS api if we
        // await on the future, and it will then at the same time await the promise.
        //
        // On the other hand, too many outstanding puts is "an issue", in that it could make
        // us exceed our 128MB memory limit and crash the whole worker. This seems like a
        // pretty theoretical concern.
        //
        // While we're at it, does this even need to return a Result, is it possible for us
        // to have an error returned?
        storage.put(KEY.0, &val).await?;
        self.value = Some(val);
        Ok(())
    }
}
```

nilslice:
Ok, I don't have any issue with making the field public. Thanks for the PR and your description - your project looks interesting! Would like to see more if/when you're ready to share :)

PR #100, opened 2021-12-18
Update matchit to version 0.4.5.

asgpng:
[matchit v0.4.4](https://github.com/ibraheemdev/matchit/releases/tag/v0.4.4) adds support for fallback routes with overlapping patterns, as specified in https://github.com/ibraheemdev/matchit/issues/4.

This PR updates the dependency version used in `worker` to 0.4.4.

I've tested this on a personal cloudflare worker site and am getting expected behavior. Let me know if there's any specific testing or things I'm missing from this PR. Thanks!


nilslice:
Hi @asgpng - 

Thanks for this! Looks like we could probably re-implement the routing to take advantage of this fallback feature instead of the workarounds we have currently. I'll probably hold off on merging this until after the holidays, when I can spend enough time considering the impacts. 

If you are interested, take a look at the Router implementation in `workers-rs` and lmk if you see anywhere specific that the updates in `matchit` could benefit us. No pressure though, thanks again for bringing this to my attention!

asgpng:
Hi @nilslice - thanks for the response! Circling back after the holidays sounds great.

> Looks like we could probably re-implement the routing to take advantage of this fallback feature instead of the workarounds we have currently.

By 'workarounds', are you referring to the [or_else_any_method_handler](https://github.com/cloudflare/workers-rs/blob/main/worker/src/router.rs#L353-L363)? That could be a pretty reasonable build, as it would reduce the surface area of the `Router` API while still providing all the existing functionality.

E.g., instead of:
```rust
Router::new()
    .post_async("/foo/specific", ...)
    .or_else_any_method("/foo/:other", ...)
```

you could do:

```rust
Router::new()
    .post_async("/foo/specific", ...)
    .post_async("/foo/:other", ...)
```

If we feel this the direction to go, I'd be happy to augment the PR with removal of the `or_else_any` logic, which seems redundant and would probably be confusing to users if offered as part of the API in addition to the native `matchit` behavior.

nilslice:
Yes, exactly! If you want to, please go ahead. Feel free to ping here with any questions / issues if you run into any.

asgpng (commented):
@nilslice, after a bit of a delay on my end, this PR should be ready for review.

kflansburg:
Superseded by #450

PR #104, opened 2021-12-28
Add support for cron triggers

### LINKED ISSUE #53: scheduled event failed
`lib.rs`:

```rust
mod utils;
use worker::*;
// could not find `Schedule` in `worker`rustc(E0433)
#[event(scheduled)]
pub async fn main(ty: String, schedule: u64, cron: String) -> Result<()> {
    println!("test");
    Ok(())
}
```

Is there something I missed?

Randoooom:
Adds support for cron triggers (resolves #53).

I currently need some inspiration for handling an undefined scheduled event in the `shim.mjs`

### DIFF
--- a/worker/src/schedule.rs
+++ b/worker/src/schedule.rs
@@ -1,0 +1,57 @@
+use std::fmt::Error;
+use std::future::Future;
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::{future_to_promise};
+use worker_sys::{Schedule as EdgeSchedule};
+
+/// [Schedule](https://developers.cloudflare.com/workers/runtime-apis/scheduled-event)
+#[derive(Debug, Clone)]
+pub struct Schedule {
+    cron: String,
+    schedule: u64,
+    ty: String,
+    pub edge: EdgeSchedule,
+    // env: Env,
+}
+
+impl From<EdgeSchedule> for Schedule {
+    fn from(schedule: EdgeSchedule) -> Self {
+        Self {
+            cron: schedule.cron(),
+            schedule: schedule.schedule(),
+            ty: schedule.ty(),
+            edge: schedule,
+        }
+    }
+}
+
+impl From<Schedule> for EdgeSchedule {
+    fn from(schedule: Schedule) -> Self {
+        schedule.edge.into()
+    }
+}
+
+impl Schedule {
+    /// get cron trigger
+    pub fn cron(&self) -> String {
+        self.cron.clone()
+    }
+
+    /// get type
+    pub fn ty(&self) -> String {
+        self.ty.clone()
+    }
+
+    /// get trigger time as u64
+    pub fn schedule(&self) -> u64 {
+        self.schedule.clone()
+    }
+
+    /// wrap for waitUntil
+    pub fn wait_until<T>(&self, future: T)
+        where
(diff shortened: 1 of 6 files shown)


nilslice:
@Randoooom - this is great, thank you for working on adding scheduled event triggers. 

I think we should still include `waitUntil` in the API, but it will require updating more code to include the `ctx` object into the function signature. I'd suggest taking a look at the docs for this function: https://developers.cloudflare.com/workers/runtime-apis/scheduled-event#syntax-module-worker

It might seem more complicated at first glance with the old implementation in `workers-rs` since it was leftover from a prior implementation where the `EventListener` format vs. modules 

Let me know if you have any specific request for guidance or commentary on this while its in review. You can also join us on our Discord: https://discord.com/invite/cloudflaredev, where the team can more continuous conversation with you.

Randoooom:
Oh yea, now I see I might be very stupid: I already implemented waitUntil but instead of using `ctx` I used the root object. I will keep working on it

nilslice:
Not stupid at all - this is impressive work! Keep at it, and let me know how we can help.

Randoooom:
Just resolved the generation of the `shim.mjs`. 
I'm not sure if there is a better variant, but this is working at the moment:
```rust
    let shim_content = format!(
        r#"
import * as imports from "./{0}.mjs";

const fetch = imports.fetch;
const scheduled = imports.scheduled;

export * from "./{0}.mjs";
export default {{ fetch, scheduled }};
"#,
        bg_name
    );
```

worker/src/schedule.rs
```diff
@@ -0,0 +1,69 @@
...
+impl From<EdgeScheduleContext> for ScheduleContext {
+    fn from(edge: EdgeScheduleContext) -> Self {
+        Self {
+            edge
+        }
+    }
+}
+
+impl ScheduleContext {
+    pub fn wait_until<T>(&self, handler: T)
```
zebp, line 56: We should try to match the signature of [this PR](https://github.com/cloudflare/workers-rs/pull/109/files#diff-e2d4e1abd1defdb178fd60c050a700788731a1e9971e3188cc996e573b28c649R29), or we should make the output of the future `worker::Result<T>` and pass that error to JavaScript.

worker/src/schedule.rs
```diff
@@ -0,0 +1,69 @@
...
+        where
+            T: Future<Output=Result<(), ()>> + 'static,
+    {
+        self.edge.wait_until(
+            future_to_promise(
+                async {
+                    handler.await;
```
zebp, line 63: This gives a warning because the result is ignored, we should either not use a result or throw it up to JavaScript.

zebp (changes requested):
Error handling for `until should be improved`.

worker/src/schedule.rs
Randoooom: Oh yea, seems like I missed the result. Regarding [the PR](https://github.com/cloudflare/workers-rs/pull/109/files#diff-e2d4e1abd1defdb178fd60c050a700788731a1e9971e3188cc996e573b28c649R29) I removed it completely.

zebp:
During my process of fixing this merge conflict, I accidentally managed to squash your commits. But this has been merged, thank you @Randoooom.

PR #106, opened 2022-01-06
Fixes module references in event macro

Dav1dde:
This fixes the event macro to not depend on global symbols, instead use the absolute module path of the worker crate. This simplifies the import and basically makes it possible to not *-use the workers crate.

This might be considered a breaking change, if anyone is depending on the generated wasm wrapper function.

Before:
```rust
use worker::*;

#[event(fetch)]
pub async main(req: Request, env: Env) -> Result<Response> {
    Response::empty()
}
```

After:
```rust
use worker::{event, Env, Request, Response};

#[event(fetch)]
pub async main(req: Request, env: Env) -> Result<Response> {
    Response::empty()
}
```

I also updated the "scheduled", but there seems to be more broken (e.g. worker::Schedule doesnt seem to exist in the first place).

### DIFF
--- a/worker-macros/src/event.rs
+++ b/worker-macros/src/event.rs
@@ -58,9 +58,9 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
             // create a new "main" function that takes the worker_sys::Request, and calls the
             // original attributed function, passing in a converted worker::Request
             let wrapper_fn = quote! {
-                pub async fn #wrapper_fn_ident(req: worker_sys::Request, env: ::worker::Env) -> worker_sys::Response {
+                pub async fn #wrapper_fn_ident(req: ::worker::worker_sys::Request, env: ::worker::Env) -> ::worker::worker_sys::Response {
                     // get the worker::Result<worker::Response> by calling the original fn
-                    match #input_fn_ident(worker::Request::from(req), env).await.map(worker_sys::Response::from) {
+                    match #input_fn_ident(::worker::Request::from(req), env).await.map(::worker::worker_sys::Response::from) {
                         Ok(res) => res,
                         Err(e) => {
                             ::worker::console_log!("{}", &e);
@@ -76,7 +76,11 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
             let output = quote! {
                 #input_fn
 
-                #wasm_bindgen_code
+                mod _worker_fetch {
+                    use ::worker::{wasm_bindgen, wasm_bindgen_futures};
+                    use super::#input_fn_ident;
+                    #wasm_bindgen_code
+                }
             };
 
             TokenStream::from(output)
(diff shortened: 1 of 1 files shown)


worker-macros/src/event.rs
```diff
@@ -58,12 +58,12 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
             // create a new "main" function that takes the worker_sys::Request, and calls the
             // original attributed function, passing in a converted worker::Request
             let wrapper_fn = quote! {
-                pub async fn #wrapper_fn_ident(req: worker_sys::Request, env: ::worker::Env) -> worker_sys::Response {
```
futursolo, line 61: Macros should use absolute paths (`::worker`)?

So if somebody defines a module named `worker` Rust still knows to resolve `::worker` to worker crate, not the user defined worker module.

nilslice:
Hi @Dav1dde - thank you for your PR! I will take a look asap. With regards to your comment about the `Scheduled` type events, they are indeed broken. However, maybe we should wait for #104 to land before attempting to merge this?

Also, I agree with @futursolo's comment:
> Macros should use absolute paths (::worker)?
> So if somebody defines a module named worker Rust still knows to resolve ::worker to worker crate, not the user defined worker module.

I will need to test your changes, but I do have some concerns that the change may lead to unintended consequences. 

It is also interesting that you mention part of the inspiration here is to avoid a `worker::*` import, but that _should_ work already - since the macro should not rely on your code's `use` statements bringing anything into scope. I'll double check this though!

worker-macros/src/event.rs
Dav1dde: I thought about that and wasn't sure, what I should use. I'll change it to `::worker` thinking about it again, it makes more sense.

Dav1dde:
> Hi @Dav1dde - thank you for your PR! I will take a look asap. With regards to your comment about the `Scheduled` type events, they are indeed broken. However, maybe we should wait for #104 to land before attempting to merge this?

I don't think this is necessary, the PR's are unrelated, I updated the broken code, but it's just as broken as before and another PR fixing the scheduled events would not be negatively affected by this.

> Also, I agree with @futursolo's comment: [...]

I'll change it.

> It is also interesting that you mention part of the inspiration here is to avoid a `worker::*` import, but that _should_ work already - since the macro should not rely on your code's `use` statements bringing anything into scope. I'll double check this though!

I did not state my intent properly, yes it works already without the *-use, but in order to make it work without star, you have to import/use a lot of symbols, with the PR you just have to use the macro (and the symbols you use in the handler). You no longer have to import symbols that the macro requires, like `wasm_bindgen` and `wasm_bindgen_futures` (which was a bit surprising for me at first).

zebp (approved):
Looks good to me!

> This might be considered a breaking change, if anyone is depending on the generated wasm wrapper function.

Hopefully no one is trying to rely on this. But if that becomes a problem, it should be easy enough to re-export it from the generated module.

PR #109, opened 2022-01-09
Add fetch context

dbw9580:
Add support for fetch context as described here: https://developers.cloudflare.com/workers/runtime-apis/fetch-event#syntax-module-worker

This enables a key use case that is using `waitUntil` to manipulate cache asynchronously.


dbw9580:
I tested with my owner worker app and it's working correctly. Not sure how to add proper tests for it in the `worker-sandbox` app.

nilslice:
@dbw9580 - thank you! I'll get this tested ASAP and will reach out here if I have any questions.

dbw9580:
Any updates?

nilslice:
@dbw9580 - sorry for the delay. I'm onboarding a new team member this week who will be working through current PRs with me (including this one). It should be included as part of the next release, likely at the end of the week. Thank you!

zebp (commented):
Looks good to me, managed to test this with some really hacky JS interop to call the cache api and it worked as expected. I don't see much of a reason to merge this until we have a painless way for us to interact with the cache. 

IMO we should wait for #67 to be merged for this. What do you think @nilslice and @dbw9580?

nilslice:
I think it depends on where #67 is.. if we're close to done, I agree let's wait. If not, maybe it can borrow some of this `Context` code.. @caass, can you chime in here on progress?

dbw9580:
I see this and the cache api as two separate things, although the key use case for fetch context is to manipulate cache with `waitUntil`.  Can we test this without involving cache? Say, set up a counter in a durable object, and increment that counter in `waitUntil`, then after a period of time, check if the counter has been updated.

nilslice:
So, let's leave #67 out of consideration for this PR. We may just need to start the Cache API work separately. 

@zebp - let's chat about if this should be the final PR to include in the `0.0.8` release tomorrow 👍

PR #110, opened 2022-01-10
Add size & type for FormData

tyhi:
This will expose the size & type functions to match how you would use them natively in js/ts workers.

### DIFF
--- a/worker/src/formdata.rs
+++ b/worker/src/formdata.rs
@@ -132,6 +132,16 @@ impl File {
         self.0.name()
     }
 
+    /// Get the file size.
+    pub fn size(&self) -> f64 {
+        self.0.size()
+    }
+
+    /// Get the file type.
+    pub fn type_(&self) -> String {
+        self.0.type_()
+    }
+
     /// Read the file from an internal buffer and get the resulting bytes.
     pub async fn bytes(&self) -> Result<Vec<u8>> {
         JsFuture::from(self.0.array_buffer())


zebp (changes requested):
My one and only issue with this PR is the return type of `size`, in the real world we're almost always going to want this as a `usize` and I don't think we'll ever run into a situation where size isn't a positive integer.

(sorry this isn't inline, github won't let me add a comment there for some reason)

tyhi:
> My one and only issue with this PR is the return type of `size`, in the real world we're almost always going to want this as a `usize` and I don't think we'll ever run into a situation where size isn't a positive integer.
> 
> (sorry this isn't inline, github won't let me add a comment there for some reason)

Yeah this makes sense to me, my question is just casting it with `as usize` the best way to do this? Like you said there shouldn't be a situation where it would be negative, I just want to make sure I do this the correct way.

Thanks!

zebp:
> > My one and only issue with this PR is the return type of `size`, in the real world we're almost always going to want this as a `usize` and I don't think we'll ever run into a situation where size isn't a positive integer.
> > (sorry this isn't inline, github won't let me add a comment there for some reason)
> 
> Yeah this makes sense to me, my question is just casting it with `as usize` the best way to do this? Like you said there shouldn't be a situation where it would be negative, I just want to make sure I do this the correct way.
> 
> Thanks!

Casting should work fine, worst case is any invalid values turn into a `0`. This should be fine:
```rust
/// Get the file size.
pub fn size(&self) -> usize {
    self.0.size() as usize
}
```

tyhi:
> > > My one and only issue with this PR is the return type of `size`, in the real world we're almost always going to want this as a `usize` and I don't think we'll ever run into a situation where size isn't a positive integer.
> > > (sorry this isn't inline, github won't let me add a comment there for some reason)
> > 
> > 
> > Yeah this makes sense to me, my question is just casting it with `as usize` the best way to do this? Like you said there shouldn't be a situation where it would be negative, I just want to make sure I do this the correct way.
> > Thanks!
> 
> Casting should work fine, worst case is any invalid values turn into a `0`. This should be fine:
> 
> ```rust
> /// Get the file size.
> pub fn size(&self) -> usize {
>     self.0.size() as usize
> }
> ```

Gotcha thanks for the help!

Sorry for the noise, may or may not have delete the fork by accident. #

PR #114, opened 2022-01-19
feat: add size and type methods to formdata

tyhi:
This will expose the size & type functions to match how you would use them natively in js/ts workers.

continue of #110 with changes requested.


zebp (approved):
Looks good to me, thanks!

PR #115, opened 2022-01-21
fix: responses occasionally corrupted

### LINKED ISSUE #64: Responses being mangled
I’m experimenting with using the `async-graphql` crate inside a worker. When I use `wee_alloc` as the global allocator, the first few bytes of each response from the worker are being consistently ‘mangled’ for some reason. Here is a repository with a minimal worker which replicates the issue: https://github.com/thomasfoster96/wealloc-issue and here is the worker in action: https://wealloc-issue.thomasfoster.workers.dev/.

Here I what I would expect my worker to produce:
```json
{"data":null,"errors":[{"message":"Bad Request"}]}
```
Instead, this is what I get:
```json
��ta":null,"errors":[{"message":"Bad Request"}]}
```

The bytes are mangled the same way on every request. Calling `serde_json::to_string` on the GraphQL request and then logging it with `console_log!` produces the correct output — this is only happening when I use `Response::from_json` or similar methods.

Creating a response body in other ways produces similar results. For example, using `serde_json::to_string` produces the correct output but using `Response::from_html` then mangles it, as does using `serde_json::to_vec` and then `Response::from_bytes`.

(Apologies if any of this is unclear, I’m very much a beginner with Rust and Cloudflare Workers)

zebp:
Fixes issue where responses can have their bodies corrupted. This is caused by us giving JS the body's bytes as a reference to a byte slice, which is then dropped causing a use after free. Most of the time this is fine, but occasionally the allocator will allocate something that overlaps with the response body.

To fix this, we instead give the [Response constructor](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response) an owned `Uint8Array` that doesn't have any references into the wasm memory.

Probably fixes #64 or at least parts of it.

### DIFF
--- a/worker-sys/src/response.rs
+++ b/worker-sys/src/response.rs
@@ -1,5 +1,7 @@
 use crate::FormData;
 use crate::ResponseInit;
+
+use js_sys::Uint8Array;
 use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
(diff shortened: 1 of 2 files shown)


ByteAlex:
Hi!

Thanks for the fix.

It works perfectly fine for my case.

nilslice (approved):
🎉🎉🎉🎉🎉 nice find, @zebp !

PR #116, opened 2022-01-23
fix: headers are not applied to stream response

### LINKED ISSUE #112: `with_headers` on `Response` does nothing when inner is `ResponseBody::Stream`
A response from `Fetch` can't be modified:

```rust
Fetch::Request(request).send().await?.with_headers(Headers::new())
```

The code compiles but doesn't actually update the headers.

My workaround:

```rust
use wasm_bindgen::{prelude::*, JsCast};
use worker::worker_sys;

#[wasm_bindgen]
extern "C" {
    pub type Response;

    #[wasm_bindgen(catch, constructor, js_class=Response)]
    pub fn new_with_opt_stream_and_init(
        body: Option<web_sys::ReadableStream>,
        init: &web_sys::ResponseInit,
    ) -> std::result::Result<Response, JsValue>;
}

impl Response {
    pub fn dup(response: worker::Response, headers: &mut worker::Headers) -> crate::Result<worker::Response> {
        let mut response_init = web_sys::ResponseInit::new();
        response_init.headers(&mut headers.0);

        let response: worker_sys::Response = response.into();
        let response = Response::new_with_opt_stream_and_init(response.body(), &response_init)?;
        let response = response.unchecked_into::<worker_sys::Response>();
        let body = worker::ResponseBody::Stream(response);

        Ok(worker::Response::from_body(body)?)
    }
}
```

Dav1dde:
I was also thinking about cloning the headers in the other `From` implementation:

```rust
impl From<EdgeResponse> for Response {
    fn from(res: EdgeResponse) -> Self {
        Self {
            // CLONE HERE
            headers: Headers(res.headers() /* .clone() */),
            status_code: res.status(),
            body: match res.body() {
                Some(_) => ResponseBody::Stream(res),
                None => ResponseBody::Empty,
            },
        }
    }
}
```

A response coming from the `cache` will not have modifiable headers, but this might sometimes not be necessary, would help a lot with ergonomics though.

Edit: Forgot I opened an issue for that, this fixes #112

### DIFF
--- a/worker-sys/src/response.rs
+++ b/worker-sys/src/response.rs
@@ -77,6 +77,13 @@ extern "C" {
         init: &web_sys::ResponseInit,
     ) -> Result<Response, JsValue>;
 
+    #[wasm_bindgen(catch, constructor, js_class=Response)]
+    #[doc = "The `new Response(..)` constructor, creating a new instance of `Response`."]
+    pub fn new_with_opt_stream_and_init(
+        body: Option<web_sys::ReadableStream>,
+        init: &web_sys::ResponseInit,
+    ) -> Result<Response, JsValue>;
+
     #[wasm_bindgen(catch, method, structural, js_class=Response, js_name=clone)]
     #[doc = "The `clone()` method."]
     pub fn clone(this: &Response) -> Result<Response, JsValue>;
--- a/worker/src/response.rs
+++ b/worker/src/response.rs
@@ -236,7 +236,15 @@ impl From<Response> for EdgeResponse {
                 .into(),
             )
             .unwrap(),
-            ResponseBody::Stream(response) => response,
+            ResponseBody::Stream(response) => EdgeResponse::new_with_opt_stream_and_init(
+                response.body(),
+                &ResponseInit {
+                    status: res.status_code,
+                    headers: res.headers,
+                }
+                .into(),
+            )
+            .unwrap(),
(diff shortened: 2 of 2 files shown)


zebp (approved):
Looks good to me, thanks!

zebp:
> I was also thinking about cloning the headers in the other `From` implementation:
> 
> ```rust
> impl From<EdgeResponse> for Response {
>     fn from(res: EdgeResponse) -> Self {
>         Self {
>             // CLONE HERE
>             headers: Headers(res.headers() /* .clone() */),
>             status_code: res.status(),
>             body: match res.body() {
>                 Some(_) => ResponseBody::Stream(res),
>                 None => ResponseBody::Empty,
>             },
>         }
>     }
> }
> ```

I don't really see a point in the cloning, the headers at that point are still owned by JavaScript so we'd just be cloning essentially a reference into the wasm-bindgen object table. If this becomes a problem feel free to open an issue about it.

Dav1dde:
@zebp the clone does more than just copy the reference:

```rust
impl Clone for Headers {
    fn clone(&self) -> Self {
        // Headers constructor doesn't throw an error
        Headers(EdgeHeaders::new_with_headers(&self.0).unwrap())
    }
}
```

It actually creates a new JS Headers object, which is **not** immutable anymore (Headers returned from the Cache etc. are readonly).

PR #117, opened 2022-01-25
Add Github Actions

### LINKED ISSUE #58: add GitHub Actions to make it easier for contributors and maintainers
Would suggest adding GitHub Actions (clippy and build check) which runs on every PR/merge to `main` that makes it easier for maintainers to review Pull Requests as well as for contributors to improve their coding style when they contribute to this repository.

Would be happy to raise a PR for the same :smile:

zebp:
Adds Github Actions to automatically test pull requests by running `worker-sandbox` through miniflare. Initially this used `wrangler publish` and `wrangler dev` but I ran into some unresolved issues in wrangler, so it might be a good idea to switch to [wrangler2](https://github.com/cloudflare/wrangler2) once it's stabilized.

This PR also goes through and fixes all the places clippy complained as I thought it'd be good to add a linter test, but a recent wasm-bindgen [bug](https://github.com/rust-lang/rust-clippy/issues/8314) has caused it to generate code that emits a clippy error so it is disabled for now.

Fixes #58


nilslice:
This is great. Maybe can you split this up into 2 PRs though? One containing the fixes, and then another with the Actions? That way we can merge the fixes, and then the Actions (maybe just the Linter first, so the checks pass, and then the integration tests once we figure out what's going wrong).

zebp:
Closing to break it up into multiple PRs.

Issue #76, opened 2021-10-23
Usage of Router inside a DurableObject: need to move env, which is required to be a reference


CRogers:
Hi,

Apologies in advance if I'm missing something obvious here, I'm new to Rust. I'm trying to use `Router` in a `#[durable_object]`s `fetch` method like so:

```rust
use worker::*;

#[durable_object]
pub struct Chatroom {
    messages: Vec<String>,
    state: State,
    env: Env,
}

#[durable_object]
impl DurableObject for Chatroom {
    fn new(state: State, env: Env) -> Self {
        Self {
            messages: vec![],
            state,
            env,
        }
    }

    async fn fetch(&mut self, req: Request) -> worker::Result<Response> {
        let router = Router::with_data(&self.messages);

        router
            .get("/", |_req, context| {
                Response::ok(&format!("{} messages", context.data().len()))
            })
            .run(req, self.env)
            .await
    }
}
```

This fails with the error:

```
error[E0507]: cannot move out of `self.env` which is behind a mutable reference
  --> src/chatroom.rs:27:23
   |
27 |             .run(req, self.env)
   |                       ^^^^^^^^ move occurs because `self.env` has type `worker::Env`, which does not implement the `Copy` trait
```

This makes sense to me, as `Router#run` takes ownership of it's `env` argument, and since I'm taking a `&mut self` as a reference I can't move `self.env`. The `#[durable_object]` macro forces `fetch` to be `&mut self` so I can _only_ give references of `env`. I can't work out how you would use Router inside it?

Maurimura7:
Hey, @nilslice I tried to derive the `Copy` trait on the `Env` definition but with no luck. So I decided to continue with the workaround until I get something working this way I could see the benefits from the current implementation, for this, I've created a wishlist-like project using Durable Objects as my persistent storage.

Everything went well at first, I was able to get data from storage pretty quick, but when I tried to put in some data problems started to appear. 

I was passing `self` to the router like this
```rust
let router = Router::with_data(self);
```
But when I tried to do
```rust
router
    .post_async("/item", |mut req, ctx| async move {
        let mut _self = ctx.data();
        let mut storage = _self.state.storage();
        storage.put("items", "new item").await?;
        Response::ok("Created")
```
this wasn't compiling because the `put` method expects a `&mut self` and `ctx.with_data()` give us a `&self`. So, my next workaround was wrapping `self` like this: 
```rust
let router = Router::with_data(Rc::new(RefCell::new(self)));
[...]

PR #118, opened 2022-01-25
Direct field access for RouteContext#env

ByteAlex:
Hi,

since data just got updated to use direct field access instead, but env is facinig the same limitiations, I thought a PR to fix that would be a good idea.

### DIFF
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -53,7 +53,7 @@ pub struct Router<'a, D> {
 /// as KV Stores, Durable Objects, Variables, and Secrets).
 pub struct RouteContext<D> {
     pub data: D,
-    env: Env,
+    pub env: Env,
     params: RouteParams,
 }
 
@@ -66,6 +66,7 @@ impl<D> RouteContext<D> {
 
     /// Get the `Env` for this Worker. Typically users should opt for the `secret`, `var`, `kv` and
     /// `durable_object` methods on the `RouteContext` instead.
+    #[deprecated(since = "0.0.8", note="please use the `env` field directly")]
     pub fn get_env(self) -> Env {
         self.env
     }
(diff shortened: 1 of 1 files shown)


nilslice:
I'm not opposed to this, but would like to know what @zebp thinks. I can see where it would make some things easier and I like that it's consistent with the `data` change. Just want to double-check that there aren't any edge cases where this could cause issue. 

It could also make #76 less of an issue. I don't recall where we left off, but I think there was discussion of adding `Clone` to `Env` if it isn't already.

worker/src/router.rs
```diff
@@ -66,6 +66,7 @@ impl<D> RouteContext<D> {
 
     /// Get the `Env` for this Worker. Typically users should opt for the `secret`, `var`, `kv` and
     /// `durable_object` methods on the `RouteContext` instead.
+    #[deprecated(since = "0.0.8", note="please use the `env` field directly")]
```
zebp, line 69: The version should probably be `0.0.9` as this correct for `0.0.8` and the formatting for the `note` argument isn't correct.

zebp (approved):
I think this looks good, I can't think of any situation off the top of my head where this change would break.

worker/src/router.rs
nilslice: We can probably keep it as `0.0.8`, unless you specifically want to wait to include this in a future release (not the upcoming `0.0.8`)?

zebp: Woops! I misremembered what version we were on. Never mind the version then.

nilslice (approved):
LGTM, thank you! I couldn't find any example code that references this which would need updating.. but happy to have someone take a second look.

ByteAlex:
Hi, sorry I just looked into this again - I copied the formatting from the above "data" field, which apparently has wrong formatting too. Eventually you want to fix it there too?

zebp:
This PR looks good for merge, the CI failure seems irrelevant and a problem on our end. I also couldn't find any out of date example code.

> Hi, sorry I just looked into this again - I copied the formatting from the above "data" field, which apparently has wrong formatting too. Eventually you want to fix it there too?

No that's fine, it's already fixed on `main`.

PR #121, opened 2022-01-25
Integration testing action

zebp:
Fixes last part of #58 and completes the breaking up of #117.

As noted in #117 this currently use [miniflare](https://github.com/cloudflare/miniflare) for testing instead of [wrangler](https://github.com/cloudflare/wrangler) or [wrangler2](https://github.com/cloudflare/wrangler2). In the future once wrangler2 is stabilized we should look into using that for integration testing but this should do for now.


nilslice:
Re-ran the job after adding a new secret for the Action, all looks good 👍

nilslice (approved):

PR #123, opened 2022-01-26
Extend the Response with_cors!

ByteAlex:
This PR is a pure QoL PR, which aims to make it easier to configure a CORS policy and apply it to responses using `with_cors` or `with_cors_unsafe` (for convenience).

### DIFF
--- a/worker/src/cors.rs
+++ b/worker/src/cors.rs
@@ -1,34 +1,5 @@
 use crate::{Error, Headers, Method, Result};
 
-/// Utility macro which applies cors to a response while properly handling the error results
-/// produced by the `with_cors` call.
-///
-/// This macro requires the calling method to have a `Result<T, From<worker::Error>>` return type!
-///
-/// Example code
-/// ```
-/// pub fn fetch() -> Result<worker::Response, worker::Error> {
-///     let cors = worker::Cors::default();
-///     let response = worker::Response::empty();
-///     return with_cors!(response, cors);
-/// }
-/// ```
-#[macro_export]
-macro_rules! with_cors {
-    ($response:expr, $cors:ident) => {
-        match $response {
-            Ok(response) => Ok(response.with_cors(&$cors)?),
-            err => err,
-        }
-    };
-    ($response:ident, $cors:ident) => {
-        match $response {
-            Ok(response) => Ok(response.with_cors(&$cors)?),
-            err => err,
-        }
-    };
-}
-
 /// Cors struct, holding cors configuration
 #[derive(Debug, Clone)]
 pub struct Cors {
--- a/worker/src/cors.rs
+++ b/worker/src/cors.rs
@@ -1,0 +1,111 @@
+use crate::{Error, Headers, Method, Result};
+
+/// Cors struct, holding cors configuration
+#[derive(Debug, Clone)]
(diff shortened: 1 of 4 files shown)


nilslice:
Thanks, @ByteAlex - I like the idea of having some utility macros. @zebp - what do you think about adding a new crate for things like this? `worker-ext` or something like that? I'd prefer not to set a precedent of adding all sorts of things to `worker` crate directly, unless they are necessary for _most_ projects. CORS is definitely closer to a main requirement for lots of API-like projects, but I think this could be a nice first for an extension/utility crate.

haze:
:D

zebp:
>  like the idea of having some utility macros. @zebp - what do you think about adding a new crate for things like this? `worker-ext` or something like that?

I think putting utility macros in `worker-macro` isn't a bad idea, but I'm unsure about this PR's `with_cors!`. The type you get is also a `worker::Result<Response>` so I don't know why we should add a macro instead of letting users use `response?.with_cors(whatever)` or `response.and_then(|r| r.with_cors)`.

But other than that macro, I think this PR is pretty good.

worker/src/response.rs
```diff
@@ -193,6 +194,19 @@ impl Response {
...
 
+    /// Sets this response's cors headers from the `Cors` struct.
+    pub fn with_cors(mut self, cors: &Cors) -> Result<Self> {
+        cors.apply_headers(self.headers_mut())?;
+        Ok(self)
+    }
+
+    /// Sets this response's cors headers from the `Cors` struct without checking for the
+    /// `apply_headers` `Result`.
+    pub fn with_cors_unsafe(mut self, cors: &Cors) -> Self {
```
zebp, line 205: I generally dislike functions that encourage you to not handle your errors and I don't see many scenarios where someone wouldn't care about the cors headers being applied. I think it'd be better if we got rid of this method and users could ignore the result themselves with:
```rust
let cors = whatever;
cors.apply_headers(&mut response).ok();
```

zebp (changes requested):
Looks good for the most part, I have two concerns that I'd like your opinion on.

worker/src/cors.rs
```diff
@@ -0,0 +1,163 @@
...
+/// Example code
+/// ```
+/// pub fn fetch() -> Result<worker::Response, worker::Error> {
+///     let cors = worker::Cors::default();
+///     let response = worker::Response::empty();
+///     return with_cors!(response, cors);
+/// }
+/// ```
+#[macro_export]
+macro_rules! with_cors {
```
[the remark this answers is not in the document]
ByteAlex: True, I totally did not think of this, my mind was stuck in Result#map which was not quite handy.

worker/src/response.rs
ByteAlex: As above, this was driven by me being stuck with the Result#map approach, fair to remove that too!

ByteAlex:
Though for general knowledge approaching macros:
The worker-marco crate contains only proc-macros so far, if I ever decide to contribute another utility macro-rules macro, it should belong into that crate rather than the respecting file in th worker crate?

zebp (approved):
Looks good to me,t hanks!

PR #124, opened 2022-01-28
WebSocket support

ByteAlex:
My initial approach to supporting websockets in workers-rs.
Sending data works fine, though receiving data - for example through onmessage - does not work yet.

I'd appreciate some help figuring this out!

Also at maintainers, would this PR, if properly documented, get an OK from you?

### DIFF
--- a/worker/src/websocket.rs
+++ b/worker/src/websocket.rs
@@ -50,8 +50,9 @@ impl WebSocket {
 
     /// Sends raw binary data through the `WebSocket`.
     pub fn send_with_binary<D: AsRef<[u8]>>(&self, bytes: D) -> Result<()> {
-        let array = js_sys::Uint8Array::new_with_length(bytes.len() as u32);
-        array.copy_from(bytes.as_ref());
+        let slice = bytes.as_ref();
+        let array = js_sys::Uint8Array::new_with_length(slice.len() as u32);
+        array.copy_from(slice);
         self.socket.send_with_u8_array(array).map_err(Error::from)
     }
 
--- a/worker/src/websocket.rs
+++ b/worker/src/websocket.rs
@@ -1,9 +1,12 @@
 use crate::ws_events::{CloseEvent, ErrorEvent, MessageEvent};
 use crate::{Error, Result};
 use serde::Serialize;
+use std::future::Future;
 use wasm_bindgen::convert::FromWasmAbi;
 use wasm_bindgen::prelude::Closure;
 use wasm_bindgen::{JsCast, JsValue};
+use wasm_bindgen_futures::JsFuture;
+use worker_sys::Response;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct WebSocketPair {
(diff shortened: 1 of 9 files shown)


ByteAlex:
I did some initial investigation on why the `onmessage` does not work and it seems that the JS `onmessage` is not called at all.

```js
export function __wbg_setonmessage_8bf4fad923b5767c(arg0, arg1) {
    console.log(new WebSocketPair()[1]);
    let ws = getObject(arg0);
    console.log(ws);
    ws.onmessage = function (evt) {
        console.log(evt);
        let fn = getObject(arg1);
        console.log(fn);
        fn(evt);
    };
};
```
The first log was solely to figure how the WS should look.
The second log was too see whether the WS is correctly instantiated.

These logs are written once a ws-request is received.

then I have the ws.onmessage and it's contents are never logged. So it does not seem to be an issue with the closure, but it implies that the ResponseInit did not properly carry the webSocket to the browser, I think?

Taking a look into the workers-chat-demo, I see
```js
return new Response(null, { status: 101, webSocket: pair[0] });
```
Which should be the same as my rust implementation.
ResponseInit in worker-sys:
```rs
    #[doc = "Change the `webSocket` field of this object."]
    pub fn websocket(&mut self, val: &WebSocket) -> &mut Self {
        let r = ::js_sys::Reflect::set(self.as_ref(), &JsValue::from("webSocket"), val.as_ref());
        debug_assert!(
            r.is_ok(),
            "setting properties should never fail on our dictionary objects"
        );
        let _ = r;
        self
    }
    ```

**Comment by ByteAlex** [comment]
I intended to go to sleep when I was sending my investigation answer, but I wanted to get this done, so I kept searching.

The issue was more stupid than I thought; It's not "set_onmessage", as implied by the generated WebSocket bindings from wasm, but set_event_handler.
After fixing those, this PR should be fully functional.
I've tested sending, receiving and echoing data aswell as the close event, but I was not able to test an "error" case yet.

**Comment by nilslice** [comment]
@ByteAlex - wow, thank you for all of this. We will start taking a look, but please convert it to a "ready for review" PR once you think it is out of Draft. 

> Also at maintainers, would this PR, if properly documented, get an OK from you?

Yes, I don't see why not! @zebp 👀 🚀

**Comment by ByteAlex** [comment]
@nilslice Will do! Planning to complete documentation later today (this evening) and will turn it into "Ready for Review".
I do not expect any more changes to the actual implementation, as I've been using the current version in my other project already and it seems solid!

What probably is important: I removed the `web_sys::ResponseInit` implementation and did an own implementation in `worker_sys::response_init::ResponseInit`, due to the required additional `webSocket` field.

**Comment by zebp** [comment]
> > Also at maintainers, would this PR, if properly documented, get an OK from you?
> 
> Yes, I don't see why not! @zebp eyes rocket

Yeah this would be awesome @ByteAlex! You're making great progress and it looks pretty good so far. Just a heads up, run the PR through clippy when you're ready for review, we currently have clippy commented out of the action workflow because of rustwasm/wasm-bindgen#2774 but we'll be re-enabling it in the near future.

This is the clippy command we are going to use for the check `cargo clippy --all --all-features -- -D warnings`

**Comment by ByteAlex** [comment]
> Yeah this would be awesome @ByteAlex! You're making great progress and it looks pretty good so far.
> 
> Edit: Removed clippy reminder, forgot it isn't commented out anymore we just pinned wasm-bindgen instead.

Haha yeah, Linters / Formatter is always complaining about my code, but I think it should be all good now + it's ready for review! 🚀

File: worker/src/websocket.rs [hunk 1]
```diff
@@ -0,0 +1,267 @@
...
+
+    /// Sends a raw string through the `WebSocket`
+    pub fn send_with_str<S: AsRef<str>>(&self, data: S) -> Result<()> {
+        self.socket
+            .send_with_str(data.as_ref())
+            .map_err(Error::from)
+    }
+
+    /// Sends raw binary data through the `WebSocket`.
+    pub fn send_with_binary(&self, bytes: Vec<u8>) -> Result<()> {
```
zebp, line 52: We probably shouldn't require a `Vec<u8>` here, someone might only have a reference to a byte slice. I suggest changing `bytes` to either `&[u8]` or `impl AsRef<[u8]>`.

worker/src/websocket.rs
```diff
@@ -0,0 +1,267 @@
...
+        let js_callback = Closure::wrap(Box::new(fun) as Box<dyn FnMut(T)>);
+        let result = self
+            .socket
+            .add_event_listener(
+                JsValue::from_str(r#type),
+                Some(js_callback.as_ref().unchecked_ref()),
+            )
+            .map_err(Error::from);
+        // we need to leak this closure, or it will be invalidated at the end of the function.
+        js_callback.forget();
```
zebp, line 187: Instead of leaking this, could we possibly just clean it up when we close the websocket? I haven't entirely wrapped my brain around what wasm-bindgen would think of that lifetime wise so if this isn't feasible just mark as resolved.

zebp (changes requested):
Super minor changes, but overall this is awesome :)

worker/src/websocket.rs
ByteAlex: Wasn't that an issue with the response referencing to released memory addresses?
This was a slice before, but to avoid the same error, I've used an owned array there!

zebp: That was unrelated, because we gave javascript iself the reference, I'm talking about taking using a slice in rust and then copying that into the `Uint8Array`. So the code would look like
```rust
/// Sends raw binary data through the `WebSocket`.
pub fn send_with_binary(&self, bytes: impl AsRef<[u8]>) -> Result<()> {
    let bytes = bytes.as_ref();
    let array = js_sys::Uint8Array::new_with_length(bytes.len() as u32);
    array.copy_from(bytes);
    self.socket.send_with_u8_array(array).map_err(Error::from)
}
```

worker/src/websocket.rs
ByteAlex: I'm not 100% sure either. Without this call, the callback will not work in 100% of the cases.
Eventually you can hold that value somewhere in the depths of the code and release it once the websocket closes, but how do you have a guarantee to get that event?
1) It's the user's responsibility to call accept();
2) We'd need to always register an extra closure to catch that event internally as well.

All together, I think, the Closure is light enough, it doesn't really matter if it's static memory.
But again, I don't know how wasm keeps the rust program open or how it would evict that.

What would be your approach to caching the closures until... "something" happens?

zebp: What *could* work, is we keep a vec of owned closures in a reference counted refcell inside of the websocket itself and then drop them all once we close. I just prototyped this out myself and found it's rather clunky and would remove the `PartialEq` and `Eq` trait impls. I'll mark this one as resolved for now and I can come back later if this becomes a problem.

zebp (approved):

PR #125, opened 2022-01-28
Fix respond_with_error attribute

ByteAlex:
The macro has not been using an import or qualified path for the Response in the macro, thus it failed to compile.
This should fix it. (It fixed it for me at least.)


zebp:
Thanks, but I'm going to close this in favor of #126 which is the same fix but with a fully qualified path.

PR #128, opened 2022-01-31
Cleanup clippy lints

zebp:
Cleans up some lints in the integration tests and removes any unnecessary `#[allow]`s that are no longer required. Makes no changes to the API except adding two `Default` implementations on `ListOptions` and `Headers`.


nilslice (approved):
LGTM!

PR #131, opened 2022-02-02
CORS Immutable Headers

ByteAlex:
Depending on the code path how the Response is constructed, the headers will eventually be immutable.
This PR fixes this by cloning the headers into a new object, applying the changes there and overwrites the old headers after.

### DIFF
--- a/worker/src/response.rs
+++ b/worker/src/response.rs
@@ -223,9 +223,10 @@ impl Response {
     ///         .with_cors(&cors)
     /// }
     /// ```
-    pub fn with_cors(mut self, cors: &Cors) -> Result<Self> {
-        cors.apply_headers(self.headers_mut())?;
-        Ok(self)
+    pub fn with_cors(self, cors: &Cors) -> Result<Self> {
+        let mut headers = self.headers.clone();
+        cors.apply_headers(&mut headers)?;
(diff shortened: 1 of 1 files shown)


zebp:
I'm having a hard time understanding the path where the headers would become immutable, I may be missing something though. Do you mind explaining what is the problem this PR is addressing?

ByteAlex:
So it happend to me when using `redirect_with_status`, but I think it also happens if you proxy another fetch-response through.
As soon as the EdgeResponse has been created, mutable access to the headers aren't necessarily guaranteed to be working.
The JS runtime will throw an error like this: `TypeError: Can't modify immutable headers.`

The difference between the paths is, that `redirect_with_status` directly works on an `EdgeResponse` which, if I understand that correctly, is immutable, while the `Response` itself is not immutable.
This breaks down to the "code path" where it's either already on the JS runtime or not, while it will not work if it the response has been passed to the JS runtime or not.

zebp:
> The JS runtime will throw an error like this: `TypeError: Can't modify immutable headers.`

Ah okay, I thought you were talking about the mutability in Rust. That makes a lot more sense.

zebp (approved):

PR #133, opened 2022-02-03
refactor websockets to be more idiomatic

zebp:
The current WebSocket api is closely modeled after the JavaScript WebSocket api which is functional but seems like a strange fit inside of a rust program. This PR is meant to refactor the WebSocket api to feel more *rust-y*.

Inspiration was taken from several other rust http frameworks (primarily [axum](https://docs.rs/axum/latest/axum/extract/ws/struct.WebSocket.html) and [warp](https://docs.rs/warp/latest/warp/filters/ws/struct.WebSocket.html)) but some of the restrictions JavaScript places on us requires us to make some changes. The biggest different in my opinion is that we don't use a [Sink](https://docs.rs/futures/latest/futures/prelude/trait.Sink.html) for sending messages through the WebSocket, which is because [send](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/send) does not return a promise for when it has finished sending the message and I thought having an async trait that always acts synchronously is strange.

This PR is a draft so if you have opinions on what the WebSocket api should look like leave your opinion, I'd love some feedback on this.

### DIFF
--- a/worker-sandbox/tests/websocket.rs
+++ b/worker-sandbox/tests/websocket.rs
@@ -19,5 +19,17 @@ fn websocket() {
         .and_then(|msg| msg.into_text())
         .unwrap();
 
-    assert_eq!(&msg, "Hello, world!")
+    assert_eq!(&msg, "Hello, world!");
+
+    // Drop the socket, causing the connection to get closed.
+    drop(socket);
+
+    println!("Dropped socket");
+
+    let got_close_event: bool = util::get("got-close-event", |r| r)
+        .text()
+        .expect("could not deserialize as text")
+        .parse()
+        .expect("body was not boolean");
+    assert!(got_close_event)
 }
--- a/worker-sandbox/tests/websocket.rs
+++ b/worker-sandbox/tests/websocket.rs
@@ -1,0 +1,23 @@
+use reqwest::Url;
+use tungstenite::{connect, Message};
+
+mod util;
+
+#[test]
+fn websocket() {
+    util::expect_wrangler();
+
+    let (mut socket, _) =
+        connect(Url::parse("ws://localhost:8787/websocket").unwrap()).expect("Can't connect");
+
+    socket
+        .write_message(Message::Text("Hello, world!".into()))
+        .unwrap();
+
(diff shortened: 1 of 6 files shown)


worker/src/websocket.rs
```diff
@@ -225,7 +293,7 @@ pub mod ws_events {
 
     impl MessageEvent {
```
nilslice, line 294: for `MessageEvent`'s `get_*` methods, do you think being consistent with `Request` impl naming is helpful? 

e.g. 
```rust
    /// Access this request's body encoded as JSON.
    pub async fn json<B: DeserializeOwned>(&mut self) -> Result<B> { ... }
    /// Access this request's body as plaintext.
    pub async fn text(&mut self) -> Result<String> { ... }
    /// Access this request's body as raw bytes.
    pub async fn bytes(&mut self) -> Result<Vec<u8>> { ... }
```

worker-sandbox/tests/websocket.rs
```diff
@@ -0,0 +1,23 @@
...
+        .write_message(Message::Text("Hello, world!".into()))
+        .unwrap();
+
+    let msg = socket
+        .read_message()
+        .and_then(|msg| msg.into_text())
+        .unwrap();
+
+    assert_eq!(&msg, "Hello, world!")
+}
```
nilslice, line 23: Maybe also add a test for the `close` event, such that the client can close, and the server sees the close?

nilslice:
https://github.com/cloudflare/workers-rs/pull/133/files#diff-9429c18070f1b18618d23f33d40eb65a6e19ce99be5ffeaf5744cd5db2526de9R58

idk why I can't leave a comment in the PR inline there, but I'd change `send_with_binary` to `send_with_bytes`. not critical, but if you rename the `MessageEvent` methods to match `Request` ones, it makes more sense.

worker/src/websocket.rs
zebp: Yeah, that'd probably preferable.

nilslice (approved):
✅ ✅ ✅ ✅ 👍

PR #134, opened 2022-02-04
fix: uninitialized wasm imports

zebp:
Previously we generated a `export_wasm.mjs` that imported all of the glue generated by wasm bindgen and modified the glue to import from the generated file. This caused a circular import leading to some functions, specifically ones that were just const aliases of builtins, generated by wasm-bindgen to be uninitialized and cause a crash. This bug can be replicated by making a worker that tries to use [Math::random()](https://docs.rs/js-sys/latest/js_sys/Math/fn.random.html).

To fix this we move all of the code in the `export_wasm.mjs` to that glue file. I don't think this would cause any side effects but I'd like to know what you think @nilslice.

### DIFF
--- a/worker-build/src/main.rs
+++ b/worker-build/src/main.rs
@@ -210,3 +210,40 @@ fn find_exports(js: &str) -> Vec<&str> {
         })
         .collect::<Vec<_>>()
 }
+
+#[cfg(test)]
+mod tests {
+    use std::path::Path;
+
+    use crate::find_exports;
+    use wasm_bindgen_cli_support::Bindgen;
+
+    #[test]
+    fn all_exports_found() {
+        let input_path = Path::new(
+            "./bindgen-test-subject/target/wasm32-unknown-unknown/release/bindgen_test_subject.wasm",
+        );
+        assert!(
+            input_path.exists(),
+            "The bindgen-test-subject module is not present, did you forget to compile it?"
+        );
+
+        let bindings = Bindgen::new()
+            .input_path(input_path)
+            .typescript(false)
+            .generate_output()
+            .expect("could not generate wasm bindings");
+        let export_names = find_exports(bindings.js());
+
+        // Ensure that every import in the final wasm binary can be found in the names we collected
+        // from the generated JavaScript.
+        for import in bindings.wasm().imports.iter() {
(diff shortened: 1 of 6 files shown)


worker-build/src/main.rs
```diff
@@ -183,3 +188,25 @@ fn write_string_to_file<P: AsRef<Path>>(path: P, contents: String) -> Result<()>
 
     Ok(())
 }
+
+// TODO: Potentially rewrite this to run an actual JavaScript parser.
```
Electroid, line 192: If you can get it working in Rust, we use this in the dashboard: https://github.com/guybedford/es-module-lexer

zebp: This seems interesting, but after a quick glance I don't know if it'd be worth the time to either port it or make some necessary changes and link. I'm fairly confident the current inelegant approach is fine for any code that wasm-bindgen generates today, I mostly wrote that comment as a suggestion if wasm-bindgen makes any changes that would break this in the future.

What do you think @nilslice?

nilslice:
> This seems interesting, but after a quick glance I don't know if it'd be worth the time to either port it or make some necessary changes and link. I'm fairly confident the current inelegant approach is fine for any code that wasm-bindgen generates today, I mostly wrote that comment as a suggestion if wasm-bindgen makes any changes that would break this in the future.
What do you think @nilslice?

I agree, not worth the squeeze to get it either ported or compiled and linked. I think it could certainly work, but for what this is trying to accomplish it's likely more effort than anyone should dedicate to this check.

Would prefer a test that expects generated code to conform to specific syntax, and will catch anything changed upstream that would break this expectation.

zebp:
> Would prefer a test that expects generated code to conform to specific syntax, and will catch anything changed upstream that would break this expectation.

Latest commit adds a test that should prevent this from happening, LMK what you think.

zebp:
> @zebp - the test looks great, but correct me if I'm wrong here, I don't see it being run?
> 
> Might need to adjust the test Action, currently specifies `worker-sandbox` only: https://github.com/cloudflare/workers-rs/actions/runs/1814455374/workflow#L44

Good catch! Should be fixed now :)

nilslice (approved):
LGTM!

PR #137, opened 2022-02-10
Support for client sided WebSockets

zebp:


worker/src/websocket.rs
```diff
@@ -12,6 +13,53 @@ use wasm_bindgen::{JsCast, JsValue};
...
+    url.set_scheme(&scheme).unwrap();
+
+    let mut req = Request::new(url.as_str(), Method::Get)?;
+    req.headers_mut()?.set("upgrade", "websocket")?;
+
+    let res = Fetch::Request(req).send().await?;
+
+    match res.websocket() {
+        Some(ws) => Ok(ws),
+        None => Err(Error::RustError("server did not accepts".into())),
```
nilslice, line 59: maybe `s/accepts/accept`? or better yet, something a bit more descriptive might be helpful. 

"server failed to make WebSocket connection" or something like that.

worker-sandbox/src/lib.rs
```diff
@@ -119,6 +119,28 @@ pub async fn main(req: Request, env: Env, _ctx: worker::Context) -> Result<Respo
             // Let the integration tests have some way of knowing if we successfully received the closed event.
             Response::ok(got_close_event)
         })
+        .get_async("/ws-client", |_, _| async move {
+            let ws = worker::connect("wss://echo.zeb.workers.dev/".parse()?).await?;
```
nilslice, line 123: is `worker` the top-level crate? if so, I think we could be more explicit here and instead add this as a method to `WebSocket` maybe? adding a public `connect` method at the current level could be misconstrued with many other "connect"-like APIs, i.e. HTTP CONNECT (which we don't support), some future TCP socket API, etc.. 

was there an issue with associating this method with one of the WebSocket types?

worker/src/websocket.rs
```diff
@@ -12,6 +13,53 @@ use wasm_bindgen::{JsCast, JsValue};
...
+///     if let WebsocketEvent::Message(msg) = event {
+///         if let Some(text) = msg.text() {
+///             return Response::ok(text);
+///         }
+///     }
+/// }
+///
+/// Response::error("never got a message echoed back :(", 500)
+/// ```
+pub async fn connect(mut url: Url) -> Result<WebSocket> {
```
nilslice, line 41: (see comment in worker-sandbox)

nilslice (changes requested):
I like it! just little bit of re-packaging / organization here will help discoverability and clarity of the API.

worker-sandbox/src/lib.rs
zebp: Workers is the top-level crate, the reason I did this was to try to follow the API of other WebSocket libraries as close as possible. I agree that the name `connect` is vague, I think either a method like `WebSocket::connect` or a more descriptive path to the function like `worker::ws::connect`  would be better. I'll go ahead with `WebSocket::connect` if that's your preference.

nilslice (approved):
LGTM!

PR #139, opened 2022-02-11
chore: fix unused function warning

zebp:
With the addition of integration testing for WebSockets we unintentionally introduced a warning where some functions declared in the integration test's util module were unused. This is because all integration tests are compiled as separate binaries so for the websocket binary functions weren't used for that specific compile job.

This also adjust flags in the CI to catch warnings like this in the future.


nilslice (approved):
👍

PR #142, opened 2022-02-15
Support for JavaScript snippets

zebp:
Adds support for `wasm-bindgen`'s [snippet](https://rustwasm.github.io/wasm-bindgen/reference/js-snippets.html) feature by moving the snippets into the correct directory and then bundling them with [SWC](https://github.com/swc-project/swc).

This PR doesn't strictly require the SWC bundler step, instead of bundling we can rename all of the generated snippets and transform the javascript. But I believe that bundling is the preferred route as it opens up the possibility for us to build bindings around existing JavaScript libraries that aren't in the global scope if we were to allow the bundling of `node_modules` noted [here](https://github.com/cloudflare/workers-rs/commit/417083fa71306df66838576af56b8eaf9b0c6b3b#diff-72e505a9360efda6be70299bfcddfafd00f301b7ed8bae515c5316d25e78b4bbR89).

Closes #63

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,28 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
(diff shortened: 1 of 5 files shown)


nilslice:
So far so good, just wondering if there is a possible test here that 

- demonstrates the location of snippets
- how to do the bindings
- that the interop works 

Does this support the inline _and_ "external" path-based snippets?

zebp:
> So far so good, just wondering if there is a possible test here that
> 
> * demonstrates the location of snippets
> * how to do the bindings
> * that the interop works
> 
> Does this support the inline _and_ "external" path-based snippets?

[This repo](https://github.com/zebp/inline-js-example) should provide a good test, but I also added snippets to `worker-build/bindgen-test-subject` so that we could run testing on the bundling process.

zebp:
Currently causes an issue where durable objects aren't exported, please wait for me to fix that before merging.

nilslice:
Nice! LGTM 👍

nilslice (approved):

PR #148, opened 2022-02-23
Start event

zebp:
Adds support for a `start` event that is invoked when the worker is initialized. This is useful for things like setting panic hooks, setting up loggers, creating any initial state, etc.

Under the hood this uses wasm-bindgen's [start feature](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/start.html) and because of that we inherit it's limitations, like not being able to register multiple start events.

### DIFF
--- a/worker-build/src/main.rs
+++ b/worker-build/src/main.rs
@@ -175,6 +175,7 @@ let importsObject = {{
 }};
 
 wasm = new WebAssembly.Instance(_wasm, importsObject).exports;
+wasm.__wbindgen_start?.();
 "#,
         bg_name,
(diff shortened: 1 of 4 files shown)


jakubadamw:
@zebp is there any way this start handler could be extended to have access to the environment variables/secrets of the worker?

zebp:
> @zebp is there any way this start handler could be extended to have access to the environment variables/secrets of the worker?

I don't believe so. As far as I'm aware there isn't any way to access the environment object until an actual event is fired (fetch, scheduled, etc) whereas the start "event" is just called when the WASM is loaded.

PR #152, opened 2022-03-10
Fix typo in README

nihaals:


nihaals:
It looks like main is already failing CI, does this need to wait until it passes before it can be merged?

zebp:
Seems to be that there was an update for the linter that adds some new checks, I'll merge and deal with that on Monday.

nihaals:
> Seems to be that there was an update for the linter that adds some new checks, I'll merge and deal with that on Monday.

This should be handled by #153

PR #153, opened 2022-03-12
Run CI on push

nihaals:


nihaals:
Looks like an upstream issue is causing tests to fail now

nihaals:
Fixed in #155

PR #156, opened 2022-03-17
Update Cass' cache-api PR with the latest workers-rs commits on main

ByteAlex:
This updates and resolves the merge conflicts with #67


ByteAlex:
@zebp Sorry this kind of drowned lately. I fixed the checks and now this should be good? Any concerns?

PR #157, opened 2022-03-22
Improve error handling

njaremko:
Improves error handling by:
- Adding support for `anyhow`
- Implementing `From<Error> for Response`

Note: After making these changes, I get the following error when I run `wrangler dev`, which seems to suggest an issue with the bundler? Because I don't think I touched anything relevant to that error?
```
Error: Something went wrong with the request to Cloudflare...
Uncaught Error: No such module "snippets/wasm-streams-8c20110b5d812e48/inline0.js".
  at line 0
 [API code: 10021]
```

### DIFF
--- a/worker-macros/src/event.rs
+++ b/worker-macros/src/event.rs
@@ -48,12 +48,13 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
 
             let error_handling = match respond_with_errors {
                 true => {
-                    quote! {
-                        ::worker::Response::error(e.to_string(), 500).unwrap().into()
+                    quote! { 
+                        ::worker::console_log!("{}", e);
+                        e.into() 
                     }
                 }
                 false => {
-                    quote! { panic!("{}", e) }
+                    quote! { e.into() }
                 }
             };
 
--- a/worker-macros/src/event.rs
+++ b/worker-macros/src/event.rs
@@ -49,8 +49,7 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
             let error_handling = match respond_with_errors {
                 true => {
                     quote! { 
-                        ::worker::console_log!("{}", e);
-                        e.into() 
+                        ::worker::Response::error(e.to_string(), 500).unwrap().into()
                     }
                 }
                 false => {
--- a/worker/src/error.rs
+++ b/worker/src/error.rs
@@ -24,46 +24,25 @@ pub enum Error {
     RustError(#[from] anyhow::Error),
 }
 
-#[inline]
-pub fn unwrap_abort<T, E>(o: Result<T, E>) -> T {
-    use std::process;
-    match o {
-        Ok(t) => t,
-        Err(_) => process::abort(),
-    }
-}
-
 impl From<Error> for Response {
     fn from(e: Error) -> Self {
         match e {
             Error::ResponseError(resp) => resp,
-            Error::Json((msg, code)) => {
-                let res = Response::from_json(&json!({
-                    "msg": msg,
-                    "code": code
-                }))
-                .map(|r| r.with_status(code));
-                unwrap_abort(res)
-            }
-            _ => unwrap_abort(Response::error("An error occurred.", 500)),
+            Error::Json((msg, code)) => Response::from_json(&json!({
+                "msg": msg,
+                "code": code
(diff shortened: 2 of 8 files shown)


worker-macros/src/event.rs
```diff
@@ -48,12 +48,13 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
 
             let error_handling = match respond_with_errors {
                 true => {
-                    quote! {
-                        ::worker::Response::error(e.to_string(), 500).unwrap().into()
+                    quote! { 
+                        ::worker::console_log!("{}", e);
```
zebp, line 52: We should probably keep the 500 response.

worker/src/error.rs
```diff
@@ -1,19 +1,70 @@
...
     BindingError(String),
-    RouteInsertError(matchit::InsertError),
+    RouteInsertError(#[from] matchit::InsertError),
     RouteNoDataError,
-    RustError(String),
-    SerdeJsonError(serde_json::Error),
+    ParseError(#[from] url::ParseError),
+    StringError(String),
+    SerdeJsonError(#[from] serde_json::Error),
+    RustError(#[from] anyhow::Error),
```
zebp, line 24: This error is never used, it should probably be removed.

worker/src/error.rs
```diff
@@ -1,19 +1,70 @@
...
+}
+
+#[inline]
+pub fn unwrap_abort<T, E>(o: Result<T, E>) -> T {
```
zebp, line 28: I don't see why this needs to exist, I think just regular `expect` calls would be better. I might not be understanding correctly though, why this instead of an `expect`?

worker/src/error.rs
```diff
@@ -1,19 +1,70 @@
...
+                unwrap_abort(res)
+            }
+            _ => unwrap_abort(Response::error("An error occurred.", 500)),
+        }
+    }
+}
+
+impl From<Error> for worker_sys::Response {
+    fn from(e: Error) -> Self {
+        match e {
```
zebp, line 55: This seems like it should just convert the error to a `Response` then `.into()` it into a sys response.

zebp (changes requested):
Seems like an improvement, but unsure on if all of these changes are necessary. It also might be good if we refactor `StringError` and make concrete error types for it's uses, but that's something I need to think over.

worker/src/error.rs
njaremko: Agreed, fixed

njaremko (commented):
Sorry I took so long to address these, notification got buried with work stuff, and I forgot to check back 😅

worker/src/error.rs
zebp: Hmm, fair. But I think we should add `anyhow` as a dependency when we could just use `Box<dyn std::error::Error>`.

worker/src/error.rs
```diff
@@ -1,19 +1,49 @@
...
     RouteNoDataError,
-    RustError(String),
-    SerdeJsonError(serde_json::Error),
+    ParseError(#[from] url::ParseError),
+    StringError(String),
+    SerdeJsonError(#[from] serde_json::Error),
+    RustError(#[from] anyhow::Error),
+}
+
+impl From<Error> for Response {
```
zebp, line 27: I don't think we should be opinionated on what an error response looks like, I think we should remove these impls. For stuff like this I think it's better for the user to have some way to turn errors into response with some structure they would like.

worker-macros/src/event.rs
```diff
@@ -48,12 +48,12 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
...
             let error_handling = match respond_with_errors {
                 true => {
-                    quote! {
+                    quote! { 
                         ::worker::Response::error(e.to_string(), 500).unwrap().into()
                     }
                 }
                 false => {
-                    quote! { panic!("{}", e) }
+                    quote! { e.into() }
```
zebp, line 56: Because we should remove the `From<Error>` for `Response`, we should go back to panicking.

zebp (changes requested):
Hey sorry it took so long to review this, it also got lost in a sea of GitHub notifications. A few small changes and then this should be good to go! There's also a merge conflict and with a PR this small a rebase should be pretty easy.

PR #158, opened 2022-03-25
Bump wasm-bindgen and loose the version constraint

quangIO:
The constraint `wasm-bindgen = "=0.2.78"` make using wasm-bindgen with some other packages impossible. Perhaps, we could relax that constraint?


zebp:
Going to merge this onto a new branch and take it from here. Thanks @quangIO

Issue #166, opened 2022-04-09
A panic in the worker leaves it unable to process any subsequent requests for a period of time


jakubadamw:
When a Rust-based worker panics, it leaves the worker dangling and unable to process subsequent requests from the same client (client A), which manifests itself in consistent 500s returned by Cloudflare. Requests from another client (client B) work fine, which suggests the sticky routing mechanism causes the requests sent by the client A to be routed to a dangling non-functional instance of the worker.

I created a minimal test case in [this repository](https://github.com/jakubadamw/cloudflare-dangling-worker-bug-reproduction) with the exact steps to reproduce.

Here's the relevant worker exhibiting this behaviour.

```rust
use worker::*;

#[event(fetch)]
pub async fn main(req: Request, env: Env, _context: Context) -> Result<Response> {
    let router = Router::new();
    router
        .get("/helloworld", |_, _ctx| Response::ok("Hello World!"))
        .get("/crashtheworker", |_, _ctx| panic!("CRASH"))
        .run(req, env)
        .await
}
```

I am not sure if this is a bug in `workers-rs` – perhaps not – but hopefully it can be routed to the appropriate team.

Thank you!

zebp:
Hmm, this is certainly a problem we should fix. After a quick look it seems that re-creating the entire wasm instance and re-setting the JS glue isn't really possible with `wasm-pack` because of it's inability to use `reference-types` (related [PR](https://github.com/rustwasm/wasm-pack/pull/937)), meaning we have an "object heap" that we can't reset. But I was able to accomplish this by using `wasm-bindgen` manually, so if `worker-build` switches to that and uses `reference-types` we should be able to fix this.

Moving away from `wasm-pack` was discussed in the past but this might be the straw that broke the camel's back. I'll work on a fix for this soon.

jakubadamw:
@zebp thanks so much for taking a look at this! This has been one of the significant issues we've been hitting with Workers in building our product.

Do you know if a potential fix for this issue is considered to be included in the nearest team's roadmap?

Thanks again.

zebp:
Thankfully, this seems to have been fixed when we updated `wasm-bindgen` in [this commit](https://github.com/cloudflare/workers-rs/commit/1ca2fc294cd22fe12f1389f09b20c9bb173dae3e) which is now released in `0.0.10`. I'm not sure which commit to `wasm-bindgen` fixed this, so we'll need to be careful in the future to make sure we don't get a regression.


[...]

PR #178, opened 2022-05-09
Add call_unwind to  to prevent panics from crashing worker

jspspike:
Currently workers crash when a panic breaks the worker until restarted


zebp:
It seems that `catch_unwind` doesn't work in in WASM as it's default (and only?) behavior is `panic=abort` ([related issue](https://github.com/rust-lang/rust/issues/58874)), which will throw a JS error we can't easily catch because of `wasm-bindgen-futures` async interop. Tested this out and the panic was never caught, and always ended up with an error being thrown by the export for invoking a shared `WasmClosure`.

I think the best way to fix #166 would be if we had a panic hook (which do get invoked!) that could reset the wasm instance. But it feels like there has to be a better way to reset the wasm state as the somewhat hacky solution i suggested there.

I think until [this](https://github.com/rust-lang/rust/issues/58874) issue is fixed, we're blocked.

zebp:
Closing as this was fixed when we updated `wasm-bindgen`.

Issue #179, opened 2022-05-09
Method for easily getting query parameters


zebp:
Currently getting query parameters is a something similar to the code below:

```rust
let mut example = None;
for (k, v) in req.url()?.query_pairs() {
    if k != "task_id" {
        example = Some(v.as_ref());
    }
}
```

A `param` method on [Url](https://docs.rs/worker/latest/worker/struct.Url.html) would be a nice QOL improvement.

sengho66:
If you are still keen, have a look at this crate I recently worked on!
https://github.com/sengho66/worker-route/tree/main

PR #180, opened 2022-05-11
Method for easily getting query parameters

witchof0x20:
Thought I'd do a try of #179 

`Url` is actually a foreign type, so a trait is required to extend its functionality, unless this is sent upstream instead. 

Because a query may have multiple values for a key, the function returns an iterator. For assumed-unique parameters, perhaps a
```rust
fn first_value_of(&self, key: &str) -> Option<Cow<'a, str>>
```
 function might also be useful.

### DIFF
--- a/worker/src/request.rs
+++ b/worker/src/request.rs
@@ -3,6 +3,7 @@ use crate::{
 };
 
 use serde::de::DeserializeOwned;
+use std::borrow::Cow;
 use url::Url;
 use wasm_bindgen::JsCast;
 use wasm_bindgen_futures::JsFuture;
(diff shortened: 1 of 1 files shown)


worker/src/request.rs
```diff
@@ -228,3 +229,28 @@ impl Request {
         &self.edge_request
     }
 }
+
+/// Used to add additional helper functions to url::Url
+pub trait UrlExt {
+    /// Given a query parameter, returns an Iterator of values for that parameter in the url's
+    /// query string
+    fn param<'a>(&'a self, key: &'a str) -> Box<dyn Iterator<Item = Cow<'a, str>> + 'a>;
```
zebp, line 237: I think if we're going just for a QOL improvement it would make more sense for this to return the first value of that name. Wanting to get all values for a query param seems more niche to me and I'd rather make the more widely easier. I think we should change the method to:
```rust
fn param<'a>(&'a self, key: &'a str) -> Cow<'a, str>;
```

zebp (changes requested):

worker/src/request.rs
fisherdarling: I do have a use case for multiple parameters. Maybe having both `param` and `param_iter` would be fine? Though tbf, I could just use `Url::query_params`.

zebp (approved):
Made a small change, but this looks good now. Thanks for contributing!

PR #175, opened 2022-05-03
Update Cache API

ocsfrank:
This is a continuation of https://github.com/cloudflare/workers-rs/pull/67.

- Updated most parameters to use references instead of values.
- Added some endpoints and tests to verify behavior of the cache API.

It seems to be working so far, but would like to have more feedback!

PR #182, opened 2022-05-13
feat: support for the Cache api

zebp:
Finally adds support for the [Cache api](https://developers.cloudflare.com/workers/runtime-apis/cache/).

This is a continuation of #67 and #175


cometkim:
Can I get a published version for this?

zebp:
> Can I get a published version for this?

A release should be right around the corner, hoping to get #180 and maybe #183 merged before then but we'll do one without them if they'll take too long.

PR #183, opened 2022-05-14
Add Service Binding Support

ImSingee:
The PR adds support for [Service Bindings](https://developers.cloudflare.com/workers/platform/bindings/about-service-bindings/)

### DIFF
--- a/worker-sys/src/remote_service.rs
+++ b/worker-sys/src/remote_service.rs
@@ -1,0 +1,18 @@
+use crate::Request as EdgeRequest;
+
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen (extends = ::js_sys::Object, js_name = Fetcher)]
+    pub type RemoteService;
+
+    #[wasm_bindgen (method, js_class = "Fetcher", js_name = fetch)]
+    pub fn fetch_with_request_internal(
+        this: &RemoteService,
+        req: &EdgeRequest,
+    ) -> ::js_sys::Promise;
+
+    #[wasm_bindgen (method, js_class = "Fetcher", js_name = fetch)]
+    pub fn fetch_with_str_internal(this: &RemoteService, url: &str) -> ::js_sys::Promise;
+}
--- a/worker/src/env.rs
+++ b/worker/src/env.rs
@@ -1,5 +1,6 @@
 use crate::durable::ObjectNamespace;
 use crate::error::Error;
+use crate::remote_service::RemoteService;
 use crate::Result;
 
 use js_sys::Object;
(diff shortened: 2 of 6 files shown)


worker-sys/src/remote_service.rs
```diff
@@ -0,0 +1,18 @@
+use crate::Request as EdgeRequest;
+
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen (extends = ::js_sys::Object, js_name = Fetcher)]
+    pub type RemoteService;
```
zebp, line 8: We should try to either match the [workers-types](https://github.com/cloudflare/workers-types) name `Fetcher` or just use `Service`. I think we should replace all occurences of `RemoteService` with `Service`.

worker/src/env.rs
```diff
@@ -46,6 +47,11 @@ impl Env {
     pub fn durable_object(&self, binding: &str) -> Result<ObjectNamespace> {
         self.get_binding(binding)
     }
+
+    /// Access Service bindings by the binding name configured in your wrangler.toml file.
+    pub fn remote_service(&self, binding: &str) -> Result<RemoteService> {
```
zebp, line 52: Should be named `service`

zebp (changes requested):
Looks good but needs some minor tweaks. I think we need to use the name `Service` instead of `Remote service` and we need to add testing in `worker-sandbox/`.

Service bindings in Miniflare seem a little rough right now, it seems that we'll need to update the GH action to add a service binding test.

ImSingee:
@zebp For the name, I think now (in the dashboard) the previous `Worker Script` is now totally renamed to `Service`, So I think a `Service Binding` implies the `Remote Service Binding`, which is difference from the current local binding.

I don't think my name is right, but I don't think just name it`service` is good either.

ImSingee:
For test, I think we must use [wrangler2](https://github.com/cloudflare/wrangler2), and it seems that the `wrangler.toml` should be changed too, maybe like this (remove `type` and change `build.upload` to `main`):

```
name = "testing-rust-worker"
workers_dev = true
compatibility_date = "2021-11-27" # required
main = "build/worker/shim.js"

kv_namespaces = [
    { binding = "SOME_NAMESPACE", id = "", preview_id = "" },
    { binding = "FILE_SIZES", id = "", preview_id = "" },
]

vars = { SOME_VARIABLE = "some value" }

[durable_objects]
bindings = [{ name = "COUNTER", class_name = "Counter" }]

[build]
command = "worker-build --release"

```

And after the migration to wrangler2 is done, I think I can add a test easily.

BTW, the current support for service binding in wrangler2 is merged (in https://github.com/cloudflare/wrangler2/pull/906) but isn't release now, so maybe we need to wait their release?

cjpatton:
miniflare 2.6.0 allegedly has bineings support: https://github.com/cloudflare/miniflare/issues/301

sergey-melnychuk:
Friendly ping @zebp @ImSingee 
Any chance making some progress here anytime soon?
If not, any objections if I pick it put and push to an acceptable form?

zebp:
Closing in favor of #243

PR #186, opened 2022-05-16
Fix `README.md` to use wrangler 2 instead of wrangler 1

ocsfrank:
This is a patch to fix the getting started portion of the `README.md` to use [Wrangler 2.0](https://github.com/cloudflare/wrangler2).

This is a temporary fix until we update all the documentations.


zebp (approved):
Looks great!

PR #188, opened 2022-05-18
Deps: remove `futures` crate in favor of minimal sub-crates

Some-Dood:
Hello there! This PR simply removes the usage of the full `futures` crate in favor of its more minimal sub-crates `futures-util` and `futures-channel`. This should slightly improve compilation times (at least when compiling the `workers` crate only) since we wouldn't have to pull down the `futures-executor` crate anymore, which comes with the larger `futures` crate (among other unnecessary compilations).

~~However, I must concede that this claim is technically not true _yet_. This crate depends on the `wasm-streams` crate, which depends on the full `futures` crate as well. This can be shown using the following command:~~

```
$ cargo tree -p worker -i futures-executor
futures-executor v0.3.21
└── futures v0.3.21
    └── wasm-streams v0.2.2
        └── worker v0.0.9 (/{REDACTED}/workers-rs/worker)
```

~~Thus, this PR doesn't actually improve much unless we work on `wasm-streams`. That is why I have worked on a separate PR (see MattiasBuelens/wasm-streams#11) which aims to resolve this issue. Until then, this PR simply minimizes our dependency on `futures`.~~

Update: the aforementioned PR has since been merged and published as part of `wasm-streams@0.2.3`.

On another note, while I was fixing up the dependencies, I also ran `cargo update` to make sure that everything is cleaned up in the dependency tree. This explains the large diff for `Cargo.lock`. Rest assured, that is all generated by Cargo. Thanks! 🎉


Some-Dood:
Quick update: `wasm-streams@0.2.3` has been released, which contains the changes I've made to finally remove the full `futures` crate from the dependency tree.

This PR is now fully ready for review. Thanks! 🥳

zebp (approved):
Thanks so much, this is great! Also amazed at the turn around time for that PR to getting it into a release.

Some-Dood:
> Also amazed at the turn around time for that PR to getting it into a release.

Special thanks to the quick response of the author @MattiasBuelens!

PR #189, opened 2022-05-18
Add DO Alarm API

fisherdarling:
This PR implements the DO Alarm API based off of the signatures found in the [storage docs](https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#methods).

I think there's a bug in the way I've implemented the binding for `getAlarm`. When an alarm does not exist, I expect `getAlarm` to return `null` and in rust `Ok(None)`, however I seem to be getting an error instead. When the alarm does exist (`set_alarm` immediately followed by a `get_alarm`) I get the expected output.

I haven't been able to test that the alarm does trigger, however everything compiles and it looks like the generated shim has the required methods. I'll be testing later tonight.

Also, I have `set_alarm` take in a `DateTime<Utc>` whereas `get_alarm` returns an `i64`. 
* Thoughts on having both a `set_alarm` and `set_alarm_datetime`? 
* Should there also be a `get_alarm_datetime` for convenience? 
* Should we have inputs and outputs just be `DateTime<Utc>`s?

### DIFF
--- a/worker-sandbox/src/counter.rs
+++ b/worker-sandbox/src/counter.rs
@@ -1,4 +1,3 @@
-use chrono::Utc;
 use worker::*;
 
 #[durable_object]
@@ -20,7 +19,7 @@ impl DurableObject for Counter {
         }
     }
 
-    async fn fetch(&mut self, req: Request) -> Result<Response> {
+    async fn fetch(&mut self, _req: Request) -> Result<Response> {
         if !self.initialized {
             self.initialized = true;
             self.count = self.state.storage().get("count").await.unwrap_or(0);
@@ -29,26 +28,10 @@ impl DurableObject for Counter {
         self.count += 10;
         self.state.storage().put("count", self.count).await?;
 
-        if req.path().contains("alarm") {
-            // Set an alarm to trigger in 500 ms:
-            let now = Utc::now();
-            self.state
-                .storage()
-                .set_alarm(now + chrono::Duration::milliseconds(500))
-                .await?;
-        }
-
         Response::ok(&format!(
             "[durable_object]: self.count: {}, secret value: {}",
             self.count,
             self.env.secret("SOME_SECRET")?.to_string()
         ))
-    }
-
-    async fn alarm(&mut self) -> Result<Response> {
-        self.count = 32;
-        self.state.storage().put("count", 32).await?;
-
-        Response::ok("ALARMED")
     }
 }
--- a/worker-sandbox/src/counter.rs
+++ b/worker-sandbox/src/counter.rs
@@ -1,3 +1,4 @@
+use chrono::Utc;
 use worker::*;
 
 #[durable_object]
@@ -19,7 +20,7 @@ impl DurableObject for Counter {
         }
     }
 
-    async fn fetch(&mut self, _req: Request) -> Result<Response> {
+    async fn fetch(&mut self, req: Request) -> Result<Response> {
         if !self.initialized {
             self.initialized = true;
             self.count = self.state.storage().get("count").await.unwrap_or(0);
@@ -27,10 +28,27 @@ impl DurableObject for Counter {
 
         self.count += 10;
         self.state.storage().put("count", self.count).await?;
+
+        if req.path().contains("alarm") {
+            // Set an alarm to trigger in 500 ms:
+            let now = Utc::now();
+            self.state
+                .storage()
+                .set_alarm(now + chrono::Duration::milliseconds(500))
+                .await?;
+        }
+
         Response::ok(&format!(
             "[durable_object]: self.count: {}, secret value: {}",
             self.count,
             self.env.secret("SOME_SECRET")?.to_string()
         ))
+    }
+
+    async fn alarm(&mut self) -> Result<Response> {
+        self.count = 32;
+        self.state.storage().put("count", 32).await?;
+
+        Response::ok("ALARMED")
     }
 }
--- a/worker-sandbox/wrangler.toml
+++ b/worker-sandbox/wrangler.toml
@@ -4,8 +4,8 @@ workers_dev = true
 compatibility_date = "2021-11-27" # required
 
 kv_namespaces = [
-    { binding = "SOME_NAMESPACE", id = "A", preview_id = "B" },
-    { binding = "FILE_SIZES", id = "C", preview_id = "D" },
+    { binding = "SOME_NAMESPACE", id = "", preview_id = "" },
+    { binding = "FILE_SIZES", id = "", preview_id = "" },
 ]
 
 vars = { SOME_VARIABLE = "some value" }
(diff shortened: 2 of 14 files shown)


worker/src/durable.rs
```diff
@@ -311,6 +312,41 @@ impl Storage {
...
+
+    /// Sets the current alarm time to the given datetime.
+    ///
+    /// If `set_alarm()` is called with a time equal to or before Date.now(), the alarm
+    /// will be scheduled for asynchronous execution in the immediate future. If the
+    /// alarm handler is currently executing in this case, it will not be canceled.
+    /// Alarms can be set to millisecond granularity and will usually execute within
+    /// a few milliseconds after the set time, but can be delayed by up to a minute
+    /// due to maintenance or failures while failover takes place.
+    pub async fn set_alarm(&self, scheduled_time: DateTime<Utc>) -> Result<()> {
```
zebp, line 333: Should have an optional options parameter that matches [setAlarm](https://github.com/cloudflare/workers-types/blob/master/index.d.ts#L434) from the workers-types repo.

As for the time parameter, maybe we could have an enum that's either a datetime or a delay and then update the method to be:
```rust
impl From<i64> for ScheduledTime { /* snip */ }
impl From<DateTime<Utc>> for ScheduledTime { /* snip */ }

pub async fn set_alarm(
    &self, 
    scheduled_time: impl Into<ScheduledTime>,
    options: Option<SetAlarmOptions>,
) -> Result<()> {
    // code
}
```

worker/src/durable.rs
```diff
@@ -311,6 +312,41 @@ impl Storage {
             .map_err(Error::from)
     }
 
+    /// Retrieves the current alarm time (if set) as integer milliseconds since epoch.
+    /// The alarm is considered to be set if it has not started, or if it has failed
+    /// and any retry has not begun. If no alarm is set, `get_alarm()` returns `None`.
+    pub async fn get_alarm(&self) -> Result<Option<i64>> {
```
zebp, line 318: Should have an optional options parameter that matches [getAlarm](https://github.com/cloudflare/workers-types/blob/master/index.d.ts#L433) from the workers-types repo.

zebp, line 318: Should have an optional options parameter that matches [getAlarm](https://github.com/cloudflare/workers-types/blob/master/index.d.ts#L433) from the workers-types repo.

worker/src/durable.rs
```diff
@@ -311,6 +312,41 @@ impl Storage {
...
+            .set_alarm_internal(js_sys::Date::new(&Number::from(
+                scheduled_time.timestamp_millis() as f64,
+            )))?
+            .into();
+        fut.await.map(|_| ()).map_err(Error::from)
+    }
+
+    /// Deletes the alarm if one exists. Does not cancel the alarm handler if it is
+    /// currently executing.
+    pub async fn delete_alarm(&self) -> Result<()> {
```
zebp, line 345: Should have an optional options parameter that matches [deleteAlarm](https://github.com/cloudflare/workers-types/blob/master/index.d.ts#L438) from the workers-types repo.

zebp, line 345: Should have an optional options parameter that matches [deleteAlarm](https://github.com/cloudflare/workers-types/blob/master/index.d.ts#L438) from the workers-types repo.

zebp (changes requested):
Some changes needed, but overall looks good. The new methods need to have option arguments and we need to add testing for this in `worker-sandbox`. Miniflare support for DO alarms seems a bit rough right now, but should be good enough for us to get basic testing.

Sorry for the delay on the review.

fisherdarling:
> Some changes needed, but overall looks good. The new methods need to have option arguments and we need to add testing for this in `worker-sandbox`. Miniflare support for DO alarms seems a bit rough right now, but should be good enough for us to get basic testing.
> 
> 
> 
> Sorry for the delay on the review.

@zebp Sounds good! I like the idea of using a trait for the argument. No worries on the delayed review too.

I'll also add some tests and those optional arguments as well.

I have some family visiting me right now so aim to get the additional work done sometime next Monday or Tuesday.

fisherdarling:
@zebp Currently stuck on testing. A few days ago I tried to add some and I didn't get it to really work locally.

worker/src/durable.rs
```diff
@@ -495,6 +564,46 @@ impl<'a> ListOptions<'a> {
     }
 }
 
+pub struct ScheduledTime {
```
zebp, line 567: Nitpick but this struct should probably impl `From<Duration>`. And a small explanation for why this exists in some documentation would probably be useful as it isn't exactly clear from the usage in `set_alarm`.

worker-sandbox/src/counter.rs
```diff
@@ -19,18 +20,35 @@ impl DurableObject for Counter {
...
+        }
+
         Response::ok(&format!(
             "[durable_object]: self.count: {}, secret value: {}",
             self.count,
             self.env.secret("SOME_SECRET")?.to_string()
         ))
     }
+
+    async fn alarm(&mut self) -> Result<Response> {
```
zebp, line 48: We should move the alarm logic to a different durable object and then run tests on that, but something like this is what I would do. We'd also need to update the test in `worker-sandbox/tests/requests.rs`.

worker-sandbox/wrangler.toml
```diff
@@ -4,8 +4,8 @@ workers_dev = true
 compatibility_date = "2021-11-27" # required
 
 kv_namespaces = [
-    { binding = "SOME_NAMESPACE", id = "", preview_id = "" },
-    { binding = "FILE_SIZES", id = "", preview_id = "" },
+    { binding = "SOME_NAMESPACE", id = "A", preview_id = "B" },
```
zebp, line 7: These should get changed back.

worker/src/durable.rs
```diff
@@ -495,6 +564,46 @@ impl<'a> ListOptions<'a> {
...
+impl From<DateTime<Utc>> for ScheduledTime {
+    fn from(date: DateTime<Utc>) -> Self {
+        ScheduledTime {
+            date: js_sys::Date::new(&Number::from(date.timestamp_millis() as f64)),
+        }
+    }
+}
+
+#[derive(Default, Serialize)]
+pub struct GetAlarmOptions {
```
zebp, line 594: These structs their fields marked as `pub` so they be set by anyone and the structs should also derive `Debug` and `Clone`.

zebp (changes requested):
Hey sorry about this review not happening sooner, but better late than never I guess! Some minor changes but this looks pretty good, we also probably should rebase this on `main`.

We'll probably finish this up ourselves as it's a blocker for us.

zebp (approved):

zebp:
Finished up and rebased this PR, thanks for the contribution!

Issue #181, opened 2022-05-12
R2 Support


ggregoriAIDEM:
Hello everyone,

I would like to know if you already have an ETA regarding R2 support.

Thank you

gsollazzo:
Hello,

Any ETA on native R2 support?

Thanks

nuvacore:
Same here. Also, in the meantime, is there an example of how to access S3 inside a Rust worker, whether AWS S3 or R2?

PR #212, opened 2022-08-15
wip: r2 binding

cometkim:
Trying #181 

I will get most of code from worker-kv. Let me know any feedback


cometkim:
Oh thanks soud great! Thanks @zebp

PR #216, opened 2022-08-31
`durable::ScheduledTime`: Properly compute offsets

cjpatton:
`From<Duration>` and `From<i64>` for `ScheduledTime` interpret the
inputs as timestamps (in particular, the number of milliseconds since
the UNIX epoch). Based on the test case in worker-sandbox
("durable_alarm"), the intention appears to interpret these as offsets
relative to the current time. This change aligns these methods with this
expectation.

### DIFF
--- a/worker/src/durable.rs
+++ b/worker/src/durable.rs
@@ -13,6 +13,7 @@
 use std::{ops::Deref, time::Duration};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


zebp (approved):
LGTM! CI failure is unrelated and has since been fixed in main.

PR #218, opened 2022-08-31
feat: add dynamic dispatch binding

zebp:
Adds support for dynamic dispatch bindings for Workers for Platforms customers.


ndisidore (approved):
I'm admittedly pretty green with wasm <-> rust bindings, but the rust code look great and we've chatted about how the JS iterop works, so lgtm!

PR #224, opened 2022-09-12
fix cargo-audit warnings

caass:


zebp (approved):
CI seems to be broken, but upon some digging it seems to be unrelated to these changes. LGTM

PR #226, opened 2022-09-19
fix: worker-build not working on windows

zebp:


devnought:
I gave your change a whirl on my local as I'm a Windows user, and have been struggling to build a worker.

`wasm-pack` wouldn't install due to `openssl`, but I worked around that via the provided installer at https://rustwasm.github.io/wasm-pack/installer/.

Next up was `worker-build` not finding that version of `wasm-pack`. I tried this change, but it still doesn't manage to find it.

Since I know my local is installed, I ended up just forcing the function to return `Ok(())` for now to get things rolling.

zebp:
> I gave your change a whirl on my local as I'm a Windows user, and have been struggling to build a worker.
> 
> `wasm-pack` wouldn't install due to `openssl`, but I worked around that via the provided installer at https://rustwasm.github.io/wasm-pack/installer/.
> 
> Next up was `worker-build` not finding that version of `wasm-pack`. I tried this change, but it still doesn't manage to find it.
> 
> Since I know my local is installed, I ended up just forcing the function to return `Ok(())` for now to get things rolling.

Sadly OpenSSL's weirdness is out of scope for this PR, you probably need to use [vc-pkg](https://github.com/microsoft/vcpkg) to get the OpenSSL headers so it can be build via Cargo. That's what I did and it was worked fine.

devnought:
I agree that's out of scope, just wanted to voice that bit of frustration :)

But I did want to let you know that even with this change, `worker-builder` still cannot find `wasm-pack` on my Windows machine.

johnomeara1:
@devnought I'm having issues with this now. How did you manage to get it to work?

PR #229, opened 2022-09-22
feat(esbuild): minify esbuild bundle by default

### LINKED ISSUE #210: [Feature] Pass `--minify` to `esbuild` in `worker-build`

### Description

Theoretically, it should be safe to pass `--minify` to `esbuild` here: https://github.com/cloudflare/workers-rs/blob/main/worker-build/src/main.rs#L129 (and adding a `--target`, see minify considerations below) which should result in a smaller glue code file (48.1kb to 22.8kb in the example I tested)

I've tested this locally for a Rust Cloudflare Worker and it worked correctly, although I'm not sure if there were specific edge cases due to which this was omitted. 

See considerations for minification with `esbuild` here: https://esbuild.github.io/api/#minify-considerations

oldcomputer2007:
esbuild bundle should be minified by default. If this behavior is not desired you may use the environment variable `NO_MINIFY`.

Closes #210

### DIFF
--- a/worker-build/src/main.rs
+++ b/worker-build/src/main.rs
@@ -118,19 +118,23 @@ fn use_glue_import() -> Result<()> {
 
 // Bundles the snippets and worker-related code into a single file.
 fn bundle(esbuild_path: &Path) -> Result<()> {
+    let no_minify = env::var("NO_MINIFY").is_ok();
     let path = PathBuf::from(OUT_DIR).join(WORKER_SUBDIR).canonicalize()?;
     let esbuild_path = esbuild_path.canonicalize()?;
-    let exit_status = Command::new(esbuild_path)
-        .args(&[
-            "--external:./index.wasm",
-            "--format=esm",
-            "--bundle",
-            "./shim.js",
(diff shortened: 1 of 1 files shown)


Jordin:
Thank you, I needed this! LGTM!

worker-build/src/main.rs
```diff
@@ -118,19 +118,23 @@ fn use_glue_import() -> Result<()> {
 
 // Bundles the snippets and worker-related code into a single file.
 fn bundle(esbuild_path: &Path) -> Result<()> {
+    let no_minify = env::var("NO_MINIFY").is_ok();
```
zebp, line 121: We should probalbly check for the [NotPresent](https://doc.rust-lang.org/stable/std/env/enum.VarError.html) error variant.

zebp (changes requested):

worker-build/src/main.rs
oldcomputer2007: Good catch. I made the changes.

zebp (approved):

zebp (approved):

PR #233, opened 2022-10-08
adding most of the missing fields in CfProperties

seeekr:
There's still a few holes, but it's good enough for my own purposes. Would be nice if someone could review this and leave me some feedback :)
Also there's a review comment in there that I'd like to discuss!

### DIFF
--- a/worker/src/request_init.rs
+++ b/worker/src/request_init.rs
@@ -114,6 +114,8 @@ pub struct CfProperties {
     /// zero and negative integers. A value of 0 indicates that the cache asset expires immediately.
     /// Any negative value instructs Cloudflare not to cache at all.
     pub cache_ttl_by_status: Option<HashMap<String, i32>>,
+    // TODO docs
+    pub image: Option<ResizeConfig>,
     /// Enables or disables AutoMinify for various file types.
     /// For example: `{ javascript: true, css: true, html: false }`.
     pub minify: Option<MinifyConfig>,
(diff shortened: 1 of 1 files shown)


worker/src/request_init.rs
```diff
@@ -305,6 +329,98 @@ impl From<PolishConfig> for &str {
...
+    #[wasm_bindgen(getter)]
+    pub fn blur(&self) -> Option<String> {
+        self.blur.clone()
+    }
+}
+
+#[wasm_bindgen]
+#[derive(Clone, Copy)]
+pub enum ResizeFit {
+    ScaleDown,
```
jakubadamw, line 375: Values of the enum types (`ResizeFit`, `ResizeFormat`, `ResizeMetadata`, `ResizeOnerror`) will need to be converted with their target naming scheme (kebab scheme) accounted for, essentially like it's done above with `PolishConfig`.

For example, `ScaleDown` should be serialised into the string `scale-down`.

I see they other `&str` conversions are spelled out explicitly, which is a bit of a shame, since that means a lot of repetition. I would use [strum::IntoStaticStr](https://docs.rs/strum_macros/latest/strum_macros/derive.IntoStaticStr.html) but I don't know if the maintainers are okay with pulling in that dependency.

jakubadamw:
@seeekr if you're keen on continuing with this PR, I have proposed some changes in https://github.com/jakubadamw/workers-rs/commits/add-cf-properties. 🙂

I tested the branch (well, some of the properties, at least) in a personal project. Seems to work fine! Thank you.

seeekr:
sorry never got around to finishing this!

seeekr:
great to hear it is useful nevertheless! good luck!

worker/src/request_init.rs
asamerican: i like to thank you for your help but i need more experience with you

PR #246, opened 2022-11-21
feat: added D1 integration

KernelFreeze:
Added a Cloudflare D1 integration using wasm_bindgen.

Pending work:
- [ ] Add tests to Worker Sandbox.
- [ ] Add `meta` field.
- [ ] Cleanup.

### DIFF
--- a/worker/src/d1.rs
+++ b/worker/src/d1.rs
@@ -1,0 +1,144 @@
+use js_sys::Array;
+use js_sys::ArrayBuffer;
+use js_sys::Uint8Array;
+use wasm_bindgen::{JsCast, JsValue};
+use wasm_bindgen_futures::JsFuture;
+use worker_sys::d1::D1Database as D1DatabaseSys;
+use worker_sys::d1::D1PreparedStatement as D1PreparedStatementSys;
+use worker_sys::d1::D1Result as D1ResultSys;
+
+use crate::env::EnvBinding;
+use crate::Result;
+
+pub struct D1Database(D1DatabaseSys);
+
+impl D1Database {
+    pub fn prepare(&self, query: &str) -> D1PreparedStatement {
+        self.0.prepare(query).into()
+    }
+
+    pub async fn dump(&self) -> Result<Vec<u8>> {
+        let array_buffer = JsFuture::from(self.0.dump()).await?;
+        let array_buffer = array_buffer.dyn_into::<ArrayBuffer>()?;
+        let array = Uint8Array::new(&array_buffer);
+        let mut vec = Vec::with_capacity(array.length() as usize);
+        array.copy_to(&mut vec);
+        Ok(vec)
+    }
+
+    pub async fn batch<T>(&self, statements: Vec<D1PreparedStatement>) -> Result<Vec<D1Result>>
+    where
+        T: JsCast,
+    {
+        let statements = statements.into_iter().map(|s| s.0).collect::<Array>();
+        let results = JsFuture::from(self.0.batch(statements)).await?;
+        let results = results.dyn_into::<Array>()?;
+        let mut vec = Vec::with_capacity(results.length() as usize);
+        for result in results.iter() {
+            let result = result.dyn_into::<D1ResultSys>()?;
+            let result = D1ResultSys::from(result);
+            vec.push(D1Result(result));
+        }
+        Ok(vec)
+    }
+
+    pub async fn exec<T>(&self, query: &str) -> Result<D1Result>
+    where
+        T: JsCast,
+    {
+        let result = JsFuture::from(self.0.exec(query)).await?;
+        let result = result.dyn_into::<D1ResultSys>()?;
+        Ok(D1Result(result))
+    }
+}
+
+impl EnvBinding for D1Database {
+    const TYPE_NAME: &'static str = "D1Database";
+}
+
+impl JsCast for D1Database {
+    fn instanceof(val: &wasm_bindgen::JsValue) -> bool {
+        val.is_instance_of::<D1Database>()
+    }
+
+    fn unchecked_from_js(val: wasm_bindgen::JsValue) -> Self {
+        Self(val.into())
+    }
+
+    fn unchecked_from_js_ref(val: &wasm_bindgen::JsValue) -> &Self {
+        unsafe { &*(val as *const JsValue as *const Self) }
+    }
+}
+
+impl From<D1Database> for JsValue {
+    fn from(database: D1Database) -> Self {
+        JsValue::from(database.0)
+    }
+}
+
+impl AsRef<wasm_bindgen::JsValue> for D1Database {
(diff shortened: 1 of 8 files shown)


worker/src/d1.rs
```diff
@@ -0,0 +1,140 @@
...
+impl D1Result {
+    pub fn success(&self) -> bool {
+        self.0.success()
+    }
+
+    pub fn error(&self) -> Option<String> {
+        self.0.error()
+    }
+
+    pub fn results(&self) -> Vec<JsValue> {
```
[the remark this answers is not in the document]
KernelFreeze: That's a good idea

KernelFreeze:
I applied your suggestion using Cloudflare's [serde_wasm_bindgen](https://github.com/cloudflare/serde-wasm-bindgen), is that ok?

zebp:
> I applied your suggestion using Cloudflare's [serde_wasm_bindgen](https://github.com/cloudflare/serde-wasm-bindgen), is that ok?

Yeah that seems fine! When you think this PR is ready, remove the draft status and give me a ping.

KernelFreeze:
I'm now working on improving the error handling. Some users reported to me that they have problems with some queries and they can't know what is the exact problem because right now it only returns "D1Error".

worker/src/d1.rs
```diff
@@ -0,0 +1,164 @@
...
+impl From<D1DatabaseSys> for D1Database {
+    fn from(inner: D1DatabaseSys) -> Self {
+        Self(inner)
+    }
+}
+
+pub struct D1PreparedStatement(D1PreparedStatementSys);
+
+impl D1PreparedStatement {
+    pub fn bind<T>(&self, value: &T) -> Result<()>
```
FlareLine, line 88: Should this be some variation of `pub fn bind<T>(&self, value: &T) -> Result<Result<D1PreparedStatement, JsValue>>`? Please correct me if I'm wrong but it looks like we're throwing away the returned statement / error?

FlareLine: Ideally replacing `JSValue` with some more appropriate type.

FlareLine:
Just playing around with this one (as I'm extremely interested in getting in across the line):

Looks like I'm having trouble using parameterized queries. A simple code example:

```rust
// ...
.get_async("/endpoint/:id", |_, ctx| async move {
let default = String::from("A");
let id = ctx.param("id").unwrap_or(&default).clone();
let d1 = ctx.env.d1("mydb")?;
let stmt = match d1
  .prepare("SELECT * FROM things WHERE thing_id = ?1")
  .bind(&id)
  {
    Ok(r) => r,
    Err(e) => return Response::error(format!("{:?}", e), 500),
  };

#[derive(Serialize, Deserialize)]
struct Thing {
  thing_id: String,
  desc: String,
  num: u16,
}
```

Using the above code (and some modifications to the workers create to allow for deeper debugging provides me with this error, irrespective of calls to `run()`/`all()`/`raw()`:

```
D1_ERROR
Error: D1_ERROR
at D1Database._send (d1-beta-facade.entry.js:682:13)
at async D1PreparedStatement.all (d1-beta-facade.entry.js:720:7))")
```

Using the same structure but with a non-parameterized query doesn't seem to error in the same way

```rust
.prepare("SELECT * FROM things WHERE thing_id = 'A'")
.bind(&id)
```

seems to return the expected result of

```
> 'Hello world!'
```

I've taken a look at serde_wasm_bindgen and it doesn't look like there's anything out of the ordinary here.

The errors provided by the binding with JS seem really obscure 🤔

Alex1607:
Having the same problem as @FlareLine when trying to bind a parameter.  
Funnily enough, the binding works fine when I added quotes around it. Like this: `[...] WHERE something = "?";`
However, it doesn't work, obviously. I just found that a bit strange.

worker/src/d1.rs
KernelFreeze: Yes, I need to rework the error handling

FlareLine:
It does appear that most of the unexpected errors occur when you attempt to use parameterized queries. Ruling out all known issues I still get errors when binding a simple string to a parameter in a query.

To get around this issue, I could use simple interpolation instead of parameter binding in my query string, as below:

```rust
.get_async("/db/:id", |_, ctx| async move {
  let default = String::from("A");
  let id = ctx.param("id").unwrap_or(&default);
  let d1 = ctx.env.d1("my-db")?;
  let query = format!("SELECT * FROM things WHERE thing_id = '{}'", id);
  let stmt = d1.prepare(&query);
  let result = stmt.all().await?;
  let things = result.results::<Thing>()?;
  Response::from_json::<Vec<Thing>>(&things)
})
.run(request, env)
.await
```

But this opens up the application to SQLi vulnerabilities and is a 'hacky' workaround.

Upon attempting to diagnose further, it looks like there's a lack of information coming back from the `wasm_bindgen` call - I'm way out of my depth to know how to diagnose past the Rust code, but does anyone have any doco on the process?

FlareLine:
### Documenting my findings for reference

I've forked `workers-rs` and `wrangler` locally to allow me to debug and make changes to the underlying JS.

I've changed a significant amount of the error handling within the local `wrangler` fork, in the D1 JS facade (`d1-beta-facade.js`) to facilitate better error visibility during debugging. This isn't something that we'd do in the JS layer to allow for this Rust D1 functionality, but we would have to work out the binding process to allow this information to be bound and discoverable at the Rust layer.

I've worked out that one of the underlying errors is actually a `SQL 9008 Invalid type error` as per this statement and response combination:
  - JS query payload: `"{\"sql\":\"SELECT * FROM things WHERE thing_id = ?1\",\"params\":[[\"A\"]]}"`
  - JS query response: `{\"error\":\"ERROR 9008: SQL query error: Invalid type\",\"success\":false,\"serve...`

From this, I've noticed that the `params` field actually contains a nested array? I'm not sure if this is intentional, but performing that same query without using binding results in a successful response.

I'll continue debugging here to see if I can get a consistent pattern working.

FlareLine (changes requested):

FlareLine:
@KernelFreeze are you still happy to work on this PR? I'd be keen to get this across the line myself - happy to attribute your contributions thus far? 😄

KernelFreeze:
> @KernelFreeze are you still happy to work on this PR? I'd be keen to get this across the line myself - happy to attribute your contributions thus far? 😄

Hey! I don't have too much time right now to finish this, so sure, go ahead ^^

FlareLine:
@zebp @Alex1607 I've created https://github.com/cloudflare/workers-rs/pull/270 to supersede this PR as per guidance from @KernelFreeze.

Alex1607:
> @zebp @Alex1607 I've created https://github.com/cloudflare/workers-rs/pull/270 to supersede this PR as per guidance from @KernelFreeze.

Thank you! I will try it out as soon as I got some time :D

PR #249, opened 2022-12-02
Add support for isEUCountry binding

slester:
Makes the `isEUCountry` binding available in Rust.

Docs: https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties

See also: https://github.com/cloudflare/workers-types/pull/245 for the TypeScript analogue.

### DIFF
--- a/worker/src/cf.rs
+++ b/worker/src/cf.rs
@@ -140,6 +140,11 @@ impl Cf {
         let tz = self.inner.timezone();
         tz.parse::<chrono_tz::Tz>().unwrap()
     }
+
+    /// Whether the country of the incoming request is in the EU
+    pub fn is_eu_country(&self) -> bool {
+        self.inner.is_eu_country() == Some("1".to_string())
+    }
 }
 
 /// Browser-requested prioritization information.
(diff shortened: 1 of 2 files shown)


worker/src/cf.rs
```diff
@@ -140,6 +140,11 @@ impl Cf {
         let tz = self.inner.timezone();
         tz.parse::<chrono_tz::Tz>().unwrap()
     }
+
+    /// Whether the country of the incoming request is in the EU
+    pub fn is_eu_country(&self) -> bool {
+        self.inner.is_eu_country() == Some("1")
```
zebp, line 146: Because the sys crate's definition returns an `Option<String>` we can't compare against a string reference in the form of a `Option<&str>`. Comparing to an own `String` would work here though.

```suggestion
        self.inner.is_eu_country() == Some("1".to_string())
```

zebp (changes requested):
Looks good for the most part, small compiler error with mismatched types.

worker/src/cf.rs
slester: Thanks for the review. Sorry about that! I refactored and forgot to change that back. Should be updated now.

zebp (approved):

PR #251, opened 2022-12-07
feat: Add queue support

jdon:
This PR adds support for [queues](https://developers.cloudflare.com/queues/).

I've only been able to test this manually as I'm not sure how I would set up the workers sandbox to test queues, if you have any ideas please let me know.


worker/src/queue.rs
```diff
@@ -0,0 +1,116 @@
...
+                .ok_or(Error::JsError(
+                    "Invalid message batch. Failed to get id from message.".to_string(),
+                ))?;
+
+            let message = Message {
+                id,
+                body,
+                timestamp: Date::from(js_date),
+            };
+            vec.push(message);
```
jdon, line 66: This whole for loop isn't ideal.

Originally I wanted:
```
pub struct Message<T> {
    pub body: T,
    pub timestamp: chrono::DateTime<UTC>,
    pub id: String,
}
```
Then we avoid all this, as we can use:

`vec.push(serde_wasm_bindgen::from_value(result)?)`

However in practice `serde_wasm_bindgen` can't deserialize JavaScript dates.

worker/src/lib.rs
```diff
@@ -36,6 +36,7 @@ pub use crate::formdata::*;
 pub use crate::global::Fetch;
 pub use crate::headers::Headers;
 pub use crate::http::Method;
+pub use crate::queue::*;
```
zebp, line 39: With queues being in beta I think we should at least gate it behind a feature flag.

jdon: Good idea, I've added a new 'queue' feature and updated the readme.

worker-macros/src/event.rs
```diff
@@ -124,6 +128,40 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            );
+            let wrapper_fn_ident = Ident::new("queue", input_fn.sig.ident.span());
+            // rename the original attributed fn
+            input_fn.sig.ident = input_fn_ident.clone();
+
+            let wrapper_fn = quote! {
+                pub async fn #wrapper_fn_ident(event: ::worker::worker_sys::MessageBatch, env: ::worker::Env, ctx: ::worker::worker_sys::Context) {
+                    // call the original fn
+                    let ctx = worker::Context::new(ctx);
+                    #input_fn_ident(::worker::MessageBatch::from(event), env, ctx).await
```
zebp, line 146: Like `fetch` handler I think we should have the user's `queue` handler return a `Result<()>`.

zebp (changes requested):
Looks really good, we should make the `queue` handler fallible and we need to add miniflare testing to `worker-sandbox`. I tried to write some tests around this but found out that miniflare's implementation of queues doesn't use the `WorkerQueue` name for the binding class. So until then I think that this PR is annoyingly blocked on https://github.com/cloudflare/miniflare/pull/455.

worker-macros/src/event.rs
jdon: Done, I've updated it to return `Result<()>`

jdon:
Thanks for pointing me in the right direction @zebp .
I ran miniflare with your changes and managed to write some tests for sending and receiving messages.

As you send we are now just blocked on  https://github.com/cloudflare/miniflare/pull/455

zebp:
It seems like it might be a while until we get a miniflare release with the fix in, if you want to update the CI workflow to use miniflare from Git in the meantime I think that's an acceptable way to get this unblocked.

I think that'll be as simple as changing [this line](https://github.com/cloudflare/workers-rs/blob/7a78ac9cd6be4d27d6fc435fa8dc60d59d55405d/.github/workflows/pullrequest.yml#LL62C38-L62C38) to `npm install -g 'cloudflare/miniflare#9d96aaa66aa48eb1107c86fe24cbfd444249533c'`.

jdon:
Good idea @zebp.

Unfortunately `npm install -g 'cloudflare/miniflare#9d96aaa66aa48eb1107c86fe24cbfd444249533c'` doesn't work, as the miniflare CLI package is inside a directory `packages/miniflare` and `npm install` doesn't support directories. 
Miniflare also uses typescript, so regardless when we install it from git we also need to compile it.

I've updated the CI to now pull the git repo and then build / link miniflare locally, it's a bit annoying but it works.

Also rust 1.66 released today and clippy included[ some extra checks](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-166). So I had to fix them to get CI to pass.

zebp (dismissed):

zebp (approved):

zebp:
LGTM!

PR #256, opened 2022-12-31
fix(worker-build): prevent script will never generate a response

### LINKED ISSUE #255: worker-build should emit all code inside a closure of the fetch function

0.0.0

### Describe the bug

Currently, the emitted code for `worker-build` has everything defined on the top-level.

I'm not sure why, but for some reason variables defined in the top level are re-used across requests.

I'm using the [`cached` crate](https://crates.io/crates/cached) in my project, and because the wasm memory is re-used across requests - it stops working and hangs with `The script will never generate a response.`

I'm not sure if this is the intended behaviour, re-using the wasm memory across requests - maybe it was done for perf reasons?

But anyways I've ran into this issue of randomly hanging workers loads of times, and it's really annoying. I think the default should be not to re-use memory across requests

### Steps To Reproduce

_No response_

samdenty:
This fixes https://github.com/cloudflare/workers-rs/issues/255

There's no clean way to convert the `index_bg.js` which contains ESM exports, into a function which is called on each `fetch()`.

So in this PR, I basically split the two parts of the bundle up and move the initialization inside the fetch function, by replacing the `INSERT_INIT()` token 

This is kinda hacky, but it works. It fixes the script will never generate a response errors that can occur when sharing the wasm state across requests.


worker-build/src/js/shim.js
```diff
@@ -1,7 +1,12 @@
-import * as imports from "./index_bg.js";
-export * from "./index_bg.js";
+export default {
+  fetch: async (...args) => {
+    INSERT_INIT();
```
samdenty, line 3: this is replaced in the script below

worker-build/src/main.rs
```diff
@@ -130,14 +131,50 @@ fn bundle(esbuild_path: &Path) -> Result<()> {
         "--outfile=shim.mjs",
     ]);
 
-    if !no_minify {
```
samdenty, line 133: we can't run minify, as we need the non-minified source to run transforms

worker-build/src/main.rs
```diff
@@ -130,14 +131,50 @@ fn bundle(esbuild_path: &Path) -> Result<()> {
...
+
+            let fetch_with_imports = format!("{}\n{}", imports, fetch);
+            let init_without_imports = imports_regex.replace_all(init, "");
+
+            let result = fetch_with_imports.replace("INSERT_INIT()", &init_without_imports);
+
+            fs::write(path.join("./shim.mjs"), result)?;
+
+            // re-bundle the shim with minification flag
+            if !no_minify {
```
samdenty, line 157: we re-run esbuild to minify it

samdenty:
ping @zebp i wonder if you know who to hit up to get this in, this is blocking me from using worker-rs

zebp:
> ping @zebp i wonder if you know who to hit up to get this in, this is blocking me from using worker-rs

left a reply in the issue

Smephite:
Is there any update on this?

worker-build/src/js/shim.js
```diff
@@ -1,7 +1,12 @@
...
-imports.start?.();
+    const imports = require("./index_bg.js");
 
-export default { fetch: imports.fetch, scheduled: imports.scheduled };
+    // Run the worker's initialization function.
+    imports.start?.();
+
+    return imports.fetch(...args);
+  },
+};
```
xtuc, line 12: this is missing the `scheduled` entrypoint

PR #264, opened 2023-01-13
Allow cloned Requests to be mutable

### LINKED ISSUE #263: [BUG] Cloned Requests are not mutable

0.0.12

### Describe the bug

Despite [the comment on `Request.headers_mut`](https://github.com/cloudflare/workers-rs/blob/main/worker/src/request.rs#L192-L201), a cloned `Request` cannot modify headers via `headers_mut`.

### Steps To Reproduce

_No response_

slester:
Fixes #263 by setting `immutable` to false when a `Request` is cloned, bringing it in line with the JS implementation.

### DIFF
--- a/worker/src/request.rs
+++ b/worker/src/request.rs
@@ -237,7 +237,11 @@ impl Request {
     pub fn clone(&self) -> Result<Self> {
         self.edge_request
             .clone()
-            .map(|req| req.into())
+            .map(|req| {
+                let mut req: Request = req.into();
+                req.immutable = false;
+                req
+            })
             .map_err(Error::from)
     }
 
--- a/worker/src/request.rs
+++ b/worker/src/request.rs
@@ -301,7 +301,11 @@ fn url_param_works() {
 
 #[test]
 fn clone_mut_works() {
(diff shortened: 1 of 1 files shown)


zebp (changes requested):
After doing some digging, I think this is actually a case of the docs on `headers_mut` being too vague. As noted in the MDN docs for [clone](https://developer.mozilla.org/en-US/docs/Web/API/Request/clone), If you intend to modify the request, you may prefer the [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) constructor. Cloned requests via the clone method are intended to be immutable, where-as when cloned via the constructor they are intended to be mutable.


I think the path forward here would be to make a `clone_mut` (not sure if that's a great name) method that uses the Request constructor rather than the clone method.

slester:
Interesting, thanks for looking into that for us @zebp! I couldn't come up with a better name than `clone_mut` :) Does it make more sense to use 

```
    pub fn clone_mut(&self) -> Result<Self> {
        let mut req = Request::new_with_init(self.url()?.as_str(), self.into());
        req.immutable = false;
        req
    }
```

I wasn't quite sure given the `EdgeRequest` structure underlying the `Request`.

zebp (approved):
Was about to suggest this solution, but it looks like you beat me to it! LGTM

PR #271, opened 2023-01-23
fix: Use range to calculate capacity of Vec in ObjectBody::bytes

### LINKED ISSUE #269: [BUG] Panic when calling .bytes() on R2 ObjectBody representing range of very large object

main branch

### Describe the bug

There is a `panicked at 'capacity overflow'` error, when calling `object_body.bytes()` on an object with a size of multiple GBs, even if there was only a very small range requested.

### Steps To Reproduce

```rs
let obj = bucket
             .get("my_file_with_multiple_gb")
             .range(Range::OffsetWithLength { offset: 0, length: 127 })
             .execute()
             .await
             .unwrap()
             .unwrap();

let obj_body =  obj.body().unwrap();

let bytes = obj_body.bytes().await.unwrap(); // this panics
```

### How to fix

The following line should probably take the length of the range (if there is any) and only use the size as a fallback:

https://github.com/cloudflare/workers-rs/blob/b55bb592ee6dbf11fa1971f2c93055ab121393b6/worker/src/r2/mod.rs#L253

DerZade:
This PR fixes #269

### DIFF
--- a/worker/src/r2/mod.rs
+++ b/worker/src/r2/mod.rs
@@ -257,7 +257,8 @@ impl<'body> ObjectBody<'body> {
             .length
             .or(range.suffix)
             .or(range.offset.map(|offset| size - offset))
-            .unwrap_or(size) as usize;
+            .unwrap_or(size)
+            .min(size) as usize;
 
         let mut bytes = Vec::with_capacity(capacity);
         let mut stream = self.stream()?;
--- a/worker/src/r2/mod.rs
+++ b/worker/src/r2/mod.rs
@@ -256,7 +256,7 @@ impl<'body> ObjectBody<'body> {
         let capacity = range
             .length
             .or(range.suffix)
-            .or(range.offset.and_then(|offset| Some(size - offset)))
+            .or(range.offset.map(|offset| size - offset))
             .unwrap_or(size) as usize;
 
         let mut bytes = Vec::with_capacity(capacity);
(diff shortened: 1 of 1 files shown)


DerZade:
Fixed the formatting.

The tests fail with `TimedOut`. I actually don't think that those errors are related to the changes in this PR.

DerZade:
Tbh I am a bit confused. The tests failed for the second time, but I still can't see how this is related to the changes in this PR.

At the end of the day, I'm just changing the capacity of a `Vec`. So in the worst case it is instantiated with the wrong capacity and therefore there is a reallocation of memory at some point. Can't see how that would lead to any failing tests.

I have rebased to main, maybe that will fix things, but otherwise I am at my wit's end

DerZade:
I played around with this a bit more and came to some pretty weird results:

Everything seems to work when hard coding the range like this:


So simply calling `self.inner.range()` seems to break the tests. 🤔 Still have no clue why this happends tho. My knowledge of the whole codebase and my time to investigate is just to limited to pinpoint the exact issue. I can just say with a fair degree of certainty that this is probably another bug unrelated to the proposed changes.

How should I continue with this? Should I just close the PR or keep it open (I allowed edits by maintainers)?

zebp:
Last week when investigating a performance issue related to R2 I made a branch with `.bytes()` being buffered in JS instead of streaming them (which has higher overhead) in [zeb/r2-buffered-bytes](https://github.com/cloudflare/workers-rs/tree/zeb/r2-buffered-bytes). I think it's probably best if we close this and instead make a PR from that branch. Thanks for the work @DerZade, sorry that it didn't end up getting merged :(

PR #260, opened 2023-01-07
Add support for multipart uploads to the R2 bindings

jakubadamw:
This extends the [recently implemented R2 bindings](https://github.com/cloudflare/workers-rs/pull/215) with support for [multipart upload](https://developers.cloudflare.com/r2/data-access/workers-api/workers-api-reference/#r2multipartupload-definition).

PR #273, opened 2023-01-30
Add a test for R2 multipart upload support to the test suite itself

jakubadamw:
I accidentally left it out in https://github.com/cloudflare/workers-rs/pull/260, so it doesn't actually run.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -63,7 +63,7 @@ jobs:
         run: |
           git clone https://github.com/cloudflare/miniflare.git
           cd miniflare
-          git checkout 9d96aaa66aa48eb1107c86fe24cbfd444249533c
+          git checkout v2.12.1
           cd packages/miniflare
           npm i
           npm link
--- a/worker-sandbox/tests/requests.rs
+++ b/worker-sandbox/tests/requests.rs
@@ -517,6 +517,12 @@ fn r2_put_with_properties() {
 }
 
 #[test]
+fn r2_put_multipart() {
+    let body = put("r2/put-multipart", |r| r).text().unwrap();
+    assert_eq!(body, "ok");
+}
+
+#[test]
 fn r2_delete() {
     let body = delete("r2/delete", |r| r).text().unwrap();
     assert_eq!(body, "ok");


jakubadamw:
Unfortunately, miniflare doesn't support the multipart APIs yet, so the test won't run… 🙁

zebp:
That's fine, I gave this a manual test before merging and everything went smoothly when writing my own test. I did discover two issues with the existing test though, part numbers must be non-zero and parts must be the same size except for the last part.

The test should be changed to:
```rs
let chunk_sizes = [
    R2_MULTIPART_CHUNK_MIN_SIZE,
    R2_MULTIPART_CHUNK_MIN_SIZE,
    500,
];
let mut uploaded_parts = vec![];
for (chunk_index, chunk_size) in chunk_sizes.iter().copied().enumerate() {
    let chunk = vec![chunk_index as u8; chunk_size];
    uploaded_parts.push(upload.upload_part((chunk_index + 1) as u16, chunk).await?);
}
upload.complete(uploaded_parts).await?;
```

jakubadamw:
> That's fine, I gave this a manual test before merging and everything went smoothly when writing my own test. I did discover two issues with the existing test though, part numbers must be non-zero and parts must be the same size except for the last part.
> 
> The test should be changed to:
> 
> ```rust
> let chunk_sizes = [
>     R2_MULTIPART_CHUNK_MIN_SIZE,
>     R2_MULTIPART_CHUNK_MIN_SIZE,
>     500,
> ];
> let mut uploaded_parts = vec![];
> for (chunk_index, chunk_size) in chunk_sizes.iter().copied().enumerate() {
>     let chunk = vec![chunk_index as u8; chunk_size];
>     uploaded_parts.push(upload.upload_part((chunk_index + 1) as u16, chunk).await?);
> }
> upload.complete(uploaded_parts).await?;
> ```

@zebp thank you for reviewing this and for doing your own tests.

Are you sure the chunks (well, all but the last one) need to be precisely equal to `R2_MULTIPART_CHUNK_MIN_SIZE`? I think I have seen this work in a side project of mine where I would only make sure the chunks are at least that size, but not necessarily of equal size. In other words, I think that's simply the minimum size, but a chunk can be larger than that?

Agreed on the chunk index. I will update the test when the miniflare PR has been merged in.

zebp:
> Are you sure the chunks (well, all but the last one) need to be precisely equal to `R2_MULTIPART_CHUNK_MIN_SIZE`?

They don't need to be the min length, but every chunk (except the last one) needs to be the same length. As long as the initial chunks are above 5M and the same size everything is fine. This has now been clarified in the docs for [uploadPart](https://developers.cloudflare.com/r2/data-access/workers-api/workers-api-reference/#r2multipartupload-definition).

Issue #276, opened 2023-02-06
[BUG] wrangler publish generates empty Worker


SebastiaanYN:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What version of `workers-rs` are you using?

0.0.11

### Describe the bug

When publishing a Worker the WASM binary is not included in the output, resulting in `Cannot read properties of undefined (reading 'fetch')`.

This occurs on both MacOS and WSL2 as well as with later versions/commits of workers-rs.

The WASM binary does get created:
```sh
$ ls -lah build/worker 
total 960
drwxr-xr-x  4 sebastiaan  staff   128B Feb  6 16:59 .
drwxr-xr-x  7 sebastiaan  staff   224B Feb  6 16:59 ..
-rw-r--r--  1 sebastiaan  staff   453K Feb  6 16:59 index.wasm
-rw-r--r--  1 sebastiaan  staff    21K Feb  6 16:59 shim.mjs
```

but not included in the output:
```sh
$ npx wrangler publish --dry-run --name foo | grep "Total Upload"
Total Upload: 21.23 KiB / gzip: 4.58 KiB
```

Versions:
```sh
$ cargo tree | grep worker
todo-worker v0.0.0
└── worker v0.0.11
    ├── worker-kv v0.5.1
    ├── worker-macros v0.0.6 (proc-macro)
    │   └── worker-sys v0.0.6
    └── worker-sys v0.0.6 (*)

$ rustup --version
rustup 1.25.1 (bb60b1e89 2022-07-12)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.66.0 (69f9c33d7 2022-12-12)`

$ node --version                              
v16.16.0
```

### Steps To Reproduce

1. `npm init cloudflare project_name worker-rust`
2. `npx wrangler publish --name foo` (`--dry-run` optionally)
3. Check upload size: `Total Upload: 21.23 KiB / gzip: 4.58 KiB`
4. Accessing the Worker results in `Cannot read properties of undefined (reading 'fetch')`

Alex1607:
I am having the same issue with my project on Windows 10.
I did some research and found out this issue doesn't exist when using the worker version `0.0.9`.
Since I already wrote something about this on the Discord, I just copied and pasted what I wrote there:
[...]

PR #278, opened 2023-02-08
Correct bindgen glue import in worker-build

CathalMullan:
`wasm-bindgen` seem to have altered their WASM import format (see https://github.com/rustwasm/wasm-bindgen/pull/3152), meaning the `use_glue_import` function in `worker-build` no longer correctly replaced the import.

Fixes https://github.com/cloudflare/workers-rs/issues/276

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1634,9 +1634,9 @@ checksum = "9c8d87e72b64a3b4db28d11ce29237c246188f4f51057d65a7eab63b7987e423"
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.83"
+version = "0.2.84"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


SebastiaanYN:
In the `install.rs` we should probably pin the versions used for `wasm-pack` to prevent issues like this in the future. I don't believe there's an easy way of running a specific version of a binary other than overwriting the existing one, but we could install a separate one locally by replacing https://github.com/cloudflare/workers-rs/blob/ff678bcf10939f14200c80526f200aca66353e04/worker-build/src/install.rs#L38 with
```rust
.args(["install", format!("wasm-pack@{WASM_PACK_VERSION}"), "--root", OUT_DIR]) 
```
and updating the `is_installed` check. And then replacing https://github.com/cloudflare/workers-rs/blob/ff678bcf10939f14200c80526f200aca66353e04/worker-build/src/main.rs#L49 with
```rust
Command::new(format!("{OUT_DIR}/bin/wasm-pack"))
```

zebp (approved):

PR #281, opened 2023-02-17
Restructure worker-sys

SebastiaanYN:
This PR consists of several things:
- Provide extensions to wasm-bindgen types instead of creating newtypes
- Restructure directory into extensions and newtypes
- Rename types to their JS equivalent
- A handful of correctness fixes

These changes are breaking for code relying on `worker_sys` rather than just `worker`.


zebp (approved):
Looks good, using extension traits is a *much* better way of handling `worker-sys` than re-doing half of `web-sys`.

PR #291, opened 2023-03-14
Fix response mangling in 0.0.14 by bringing back an accidentally rolled back fix from #115

jakubadamw:
See https://github.com/cloudflare/workers-rs/pull/115 for more details.
Fixes https://github.com/cloudflare/workers-rs/issues/64 which re-emerged in 0.0.14.

### DIFF
--- a/worker/src/response.rs
+++ b/worker/src/response.rs
@@ -354,16 +354,20 @@ impl From<ResponseInit> for web_sys::ResponseInit {
 impl From<Response> for web_sys::Response {
     fn from(res: Response) -> Self {
         match res.body {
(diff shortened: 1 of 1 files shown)


zebp (changes requested):
There's still UB in the `From<&Response>` for `web_sys::Response` which does something similar, instead it clones the bytes vec and then drops it. It needs to be fixed there but otherwise this is good.

zebp:
Changed the base of the PR to a new branch that I have write access to so I can commit the second fix.

jakubadamw:
@zebp right, sorry, a busy day. Thanks! 🙂

PR #306, opened 2023-04-11
expose Context underlying JsObject

xtuc:


worker/src/context.rs
```diff
@@ -42,4 +42,8 @@ impl Context {
     pub fn pass_through_on_exception(&self) {
         self.inner.pass_through_on_exception()
     }
+
+    pub fn inner(&self) -> &JsContext {
```
zebp, line 46: I think this would be better if we exposed it via an  `AsRef<JsContext>` impl.

zebp (changes requested):

zebp (approved):

Issue #54, opened 2021-09-18
Workers Sites support


SeokminHong:
Hi 👋 

I had tried to re-implement the Workers Sites using Rust, but I was stuck by KV access.

As far as I know, the [kv-asset-handler](https://github.com/cloudflare/kv-asset-handler) package uses the `STATIC_CONTENT_MANIFEST` variable from global context generated by wrangler. Am I right?

If it is, can you provide some ways to accessing the manifest or Rust version of `getAssetFromKV` function?

nilslice:
I don't have a solution in mind for this, but it is a good question.  
The team is generally recommending the use of Cloudflare Pages instead of Worker Sites, but I understand the desire to use it.

I'm not super familiar with the asset manifest, or with Workers Sites, but will start tracking this request and see if there is something we will officially support. 

For now, the code in the Wrangler codebase (https://github.com/cloudflare/wrangler) may point you in the right direction as to reading/deconstructing the asset manifest for use in a Rust KvAssetHandler.

Dav1dde:
You can access static files through the`__STATIC_CONTENT` KV:

```rust
#[event(fetch)]
pub async fn main(req: worker::Request, env: worker::Env) -> worker::Result<worker::Response> {
    let kv = worker::kv::KvStore::from_this(&env, "__STATIC_CONTENT")?;

    let index = kv.get("index.html").text().await?.expect("index html");
    worker::Response::from_html(index)
}
```

There is still a lot missing, like caching, mime types etc, butyou can just follow the JS implementation of `getAssetFromKV` to do that yourself, e.g. I have a small helper like this:

```rust
pub async fn serve_asset(req: Request, store: KvStore) -> worker::Result<Response> {
    let path = req.path();
    let path = path.trim_start_matches('/');
    let value = match store.get(path).bytes().await? {
        Some(value) => value,
        None => return Response::error("Not Found", 404),
    };
    let mut response = Response::from_bytes(value)?;
    response
        .headers_mut()
        .set("Content-Type", get_mime(path).unwrap_or("text/plain"))?;
    Ok(response)
}

fn get_mime(path: &str) -> Option<&'static str> {
    let ext = if let Some((_, ext)) = path.rsplit_once(".") {
        ext
    } else {
        return None;
    };

    let ct = match ext {
        "html" => "text/html",
        "css" => "text/css",
        "js" => "text/javascript",
        "json" => "application/json",
        "png" => "image/png",
        "jpg" => "image/jpeg",
[...]

PR #308, opened 2023-04-13
[feature] Add support for kv assets.

armfazh:
It exposes `env.asset_key` function so filenames can be queried from the "__STATIC_CONTENT" KV namespace.

Previously, names were mangled due to an additional hash appended to the names.
Now, inside the worker handler one can access to `favicon.ico` as (simplified):

```rust
let key = ctx.env.asset_key("favicon.ico");
let icon = ctx.kv("__STATIC_CONTENT")?.get(&key);
```

This PR depends on:

* **workers-sdk**: https://github.com/cloudflare/workers-sdk/pull/3084 
  A small change to always load a (possibly empty) manifest module.
* **miniflare**: https://github.com/cloudflare/miniflare/pull/622
  For running tests.
* **workerd**: https://github.com/cloudflare/workerd/pull/828
  For running tests.

Once these PRs get merged, CI will pass.

---
Credits: This code was based on @SeokminHong comment in #54


worker-sys/src/asset_handler.rs
```diff
@@ -0,0 +1,13 @@
...
+use once_cell::sync::Lazy;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(module = "__STATIC_CONTENT_MANIFEST")]
+extern "C" {
+    #[wasm_bindgen(js_name = "default")]
+    static MANIFEST_STR: String;
+}
+
+pub static ASSET_MANIFEST: Lazy<HashMap<&str, &str>> =
```
zebp, line 12: Not sure why this file is included twice.

worker-sys/src/ext/asset_handler.rs
```diff
@@ -0,0 +1,13 @@
+use std::collections::HashMap;
+
+use once_cell::sync::Lazy;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(module = "__STATIC_CONTENT_MANIFEST")]
+extern "C" {
```
zebp, line 7: We should try to avoid having anything but wasm-bindgen glue in the `worker-sys` crate if possible, this global should be moved to the `worker` crate.

worker/src/asset_handler.rs
```diff
@@ -0,0 +1,22 @@
+use crate::request::Request;
+use crate::response::Response;
+use crate::{Headers, Result, RouteContext};
+use worker_sys::ext::ASSET_MANIFEST;
+
+const ASSET_NAMESPACE: &str = "__STATIC_CONTENT";
+
+pub async fn get_asset_from_kv<D>(req: Request, ctx: RouteContext<D>) -> Result<Response> {
```
zebp, line 8: Needs documentation.

zebp (changes requested):
I'm unsure if this deserves to be merged into the workers-rs framework itself. Like the existing TypeScript [kv-asset-handler](https://github.com/cloudflare/kv-asset-handler/tree/main) this would be better served as a standalone library IMO.

That being said, I'm not totally against adding first party support if there's a good reason as to why it should be in the official crate.

worker-sys/src/ext/asset_handler.rs
```diff
@@ -0,0 +1,13 @@
...
+    #[wasm_bindgen(js_name = "default")]
+    static MANIFEST_STR: String;
+}
+
+pub static ASSET_MANIFEST: Lazy<HashMap<&str, &str>> =
```
zebp, line 12: We should avoid things other than wasm-bindgen glue in the `worker-sys` crate, this global should be moved to the `worker` crate.

zebp (changes requested):

worker-sys/src/ext/asset_handler.rs
armfazh: I took a different approach.
It's true that all the boilerplate code can be in a different crate. 
So the essential function that is missing to anyone do that is a function that converts the filenames under `/public`:
`favicon.ico` -> `favicon.<hash>.ico` according to the value stored in the manifest file.

In the new changes, we expose this function through the `env` variable.

penalosa:
@zebp From the wrangler/miniflare side I tend to agree. Use of Workers Sites is discouraged in favour of Pages, and the assets story in Wrangler will be revisited as part of Pages/Workers convergence in the near future. As such, adding additional functionality to Workers Sites is generally not something we want to do, and introducing a new API directly to the `workers-rs` framework for this seems like a maintenance footgun in the not too distant future.

Issue #151, opened 2022-03-09
I tried omitting (await) in durable objects, but it doesn't seem to work.


piloer:
I saw an explanation of durable objects lifespan here
https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#durable-object-lifespan

When I try to omit ctx.wait_until and remove .await from the Fetch, the durable objects does not wait for the request to complete

SeokminHong:
`async` bodies and other futures are lazy: they do nothing until they are run. (https://rust-lang.github.io/async-book/03_async_await/01_chapter.html)

You can use `worker::Context::wait_until` function that makes the execution context can live after the response is returned.

PR #309, opened 2023-04-14
Exposes wait_until method on Durable Object state

will118:
This is needed to get the usual `waitUntil` behaviour for DO.

In JS you can just "fire and forget" the promise and DO will take care of ensuring it runs: https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#durable-object-class-definition

But as futures are lazy in Rust, this is the only way.

Related: https://github.com/cloudflare/workers-rs/issues/151

### DIFF
--- a/worker/src/durable.rs
+++ b/worker/src/durable.rs
@@ -23,6 +23,7 @@ use crate::{
 
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
+use futures_util::Future;
 use js_sys::{Map, Number, Object};
 use serde::{de::DeserializeOwned, Serialize};
 use wasm_bindgen::{prelude::*, JsCast};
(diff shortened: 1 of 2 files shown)


worker/src/durable.rs
```diff
@@ -200,7 +211,9 @@ pub struct Storage {
 
 impl Storage {
     /// Retrieves the value associated with the given key. The type of the returned value will be
-    /// whatever was previously written for the key, or undefined if the key does not exist.
+    /// whatever was previously written for the key.
+    ///
+    /// Returns [Err] if the key does not exist.
```
will118, line 216: I got tripped up by this while testing

zebp (approved):

worker/src/durable.rs
zebp: I guess we accidentally left this in when copy pasting the JS documentation, I'll deal with that in a follow up PR when I clean  up some documentation.

PR #270, opened 2023-01-20
feat: Added D1 Support

FlareLine:
# Summary

These changes add D1 support to the `worker` crate, resolving #221.

Due to the nature of local development with Rust workers, developing an easy process for locally debugging this integration may be somewhat difficult.

Attributions:

- @KernelFreeze for initial contributions to the feature.
- @SebastiaanYN for assistance with the `query!` macro.
- @logankeenan for working on the integration tests.
- @zebp for all the work they do on the crate, despite their massive existing workload.

Issue #221, opened 2022-09-05
[Feature] D1 Support


vidiabtc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

It would be great if workers-rs had the functionality of wrangler@d1. Although I understand that it may not be the best time to do this as D1 is still in private beta, but it would be cool if the D1 integration was ready to go by the time D1 is public!

Yrlish:
I would also be interested in this. I'm also aware that the current state of D1 is in private beta so it may not be doable at the moment.

Yrlish:
> In the future I think we'll have to figure out how we're going to go about private-beta bindings, possible through an `unstable` feature flag?

Yeah this sound like a good idea, some sort of unstable branch for non-GA features.

I was also thinking of an alternative solution to what I was thinking about above. Having two separate workers, one in Rust and one "normal" in Typescript and connecting them via a service binding.

yzn-h:
it is now open alpha ╰(*°▽°*)╯

KianNH:
https://github.com/cloudflare/workers-rs/pull/246

Keavon:
#270

frontmesh:
Will the d1 support be available in 0.0.18 release ? I see that it has been merged, but not yet released.

spicyzboss:
I can't enable d1 feature but I have seen [this commit](https://github.com/cloudflare/workers-rs/commit/e7b41bf5e37d38aa82c7b8fde7fbbaf316f1192c) that not released in 0.0.17. Any updates?

PR #314, opened 2023-05-05
feat: Added D1 Support

logankeenan:
This is a continuation of #270.  

@FlareLine made the following [statement](https://github.com/cloudflare/workers-rs/pull/270#issuecomment-1495492188) the existing PR:

> I'd be more than happy if you wanted to contribute some more tests to the PR - I've unfortunately come down with something so haven't had time to come back to this.

I made a [PR](https://github.com/FlareLine/workers-rs/pull/1) on his fork, but it haven't had a response.  I'll see if I can moe this PR along and would be happy to close this PR if they can come back. 

# Summary

These changes add D1 support to the worker create, resolving #221. 

I've added some integration tests to worker-sandbox and left a few comments to see how to proceed. @zebp let me know your thoughts and thank you for your time.


worker-sandbox/tests/d1.rs
```diff
@@ -0,0 +1,115 @@
...
+    expect_wrangler();
+    setup();
+    let person = create_person();
+
+    assert_eq!("Ada LoveLace", person.name);
+    assert_eq!(32, person.age);
+}
+
+#[test]
+fn d1_select_prepare_all() {
```
logankeenan, line 44: I tried to highlight the API that is being used in the tests, but I think they might need to be a bit more verbose

worker/src/d1/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /// Executes a query against the database and returns all rows and metadata.
+    pub async fn all(&self) -> Result<D1Result> {
+        let result = JsFuture::from(self.0.all()).await?;
+        Ok(D1Result(result.into()))
+    }
+
+    /// Executes a query against the database and returns a `Vec` of rows instead of objects.
+    pub async fn raw<T>(&self) -> Result<Vec<Vec<T>>>
```
logankeenan, line 175: I don't think it should take a generic type T because consumers will unlikely write a select statement where all the types are the same. In the people case, select * from my table will cause a type error between the String, Integer or other types defined in the table's column.  Perhaps, it should be a String rather than T?

worker/src/d1/macros.rs
```diff
@@ -0,0 +1,38 @@
...
+/// ```
+/// let query = worker::query!(
+///   &d1,
+///   "SELECT * FROM things WHERE num > ?1 AND num < ?2",
+///   &min,
+///   &max,
+/// )?;
+/// ```
+#[macro_export]
+macro_rules! query {
```
logankeenan, line 18: I haven't written any tests for this, but could

worker-sandbox/tests/d1.rs
logankeenan, line 1: This doesn't test all functionality yet, but I wanted to see what thoughts are on my approach.

FlareLine:
Hey @logankeenan! I've just commented on the original PR - I would be happy to merge these changes into the main PR if you're happy with that? I believe there's still some conversations going on with the D1 team, which is why the PR hasn't been merged / worked on recently.

logankeenan:
> Hey @logankeenan! I've just commented on the original PR - I would be happy to merge these changes into the main PR if you're happy with that? I believe there's still some conversations going on with the D1 team, which is why the PR hasn't been merged / worked on recently.

That sounds good to me, thank you!

worker/src/d1/mod.rs
FlareLine: This is true - I'll take a look. I think this was supposed to be a `Result<Vec<T>>`, but not sure.

PR #318, opened 2023-05-10
chore: expose get_binding for env

zebp:
Exposes `get_binding` for accessing bindings that do not have a type in workers-rs.


worker-sandbox/src/r2.rs
```diff
@@ -187,7 +187,7 @@ pub async fn put_properties(_req: Request, ctx: RouteContext<SomeSharedData>) ->
 
 pub async fn put_multipart(_req: Request, ctx: RouteContext<SomeSharedData>) -> Result<Response> {
     const R2_MULTIPART_CHUNK_MIN_SIZE: usize = 5 * 1_024 * 1_024; // 5MiB.
-    const TEST_CHUNK_COUNT: usize = 3;
+    // const TEST_CHUNK_COUNT: usize = 3;
```
aseure, line 190: If unneeded, could we just remove that line? Because the formatter wants it to be aligned with the previous `// 5 MiB` comment.

aseure:
Thank you very much for this @zebp

PR #324, opened 2023-05-23
Outbound TCP Sockets

kflansburg:
- [x] Box JS types so that Socket is Sync / Send.
- [x] Test Secure Transport On
- [x] Test Allow Half Open
- [x] Remove unwraps
- [x] Test StartTls
- [x] Test / clean up methods / attributes other than `read` and `write`

### DIFF
--- a/worker-build/src/js/shim.js
+++ b/worker-build/src/js/shim.js
@@ -4,4 +4,4 @@ export * from "./index_bg.js";
 // Run the worker's initialization function.
 imports.start?.();
 
-export default { fetch: imports.fetch, scheduled: imports.scheduled, queue: imports.queue };
+export default { connect: imports.connect, fetch: imports.fetch, scheduled: imports.scheduled, queue: imports.queue };
--- a/worker-build/src/js/shim.js
+++ b/worker-build/src/js/shim.js
@@ -4,4 +4,4 @@ export * from "./index_bg.js";
 // Run the worker's initialization function.
 imports.start?.();
 
-export default { connect: imports.connect, fetch: imports.fetch, scheduled: imports.scheduled, queue: imports.queue };
+export default { fetch: imports.fetch, scheduled: imports.scheduled, queue: imports.queue };
--- a/worker-sys/src/types/socket.rs
+++ b/worker-sys/src/types/socket.rs
@@ -21,3 +21,101 @@ extern "C" {
     #[wasm_bindgen(method, js_name=startTls)]
     pub fn start_tls(this: &Socket) -> Socket;
 }
+
+impl Socket {
+    pub fn readable(&self) -> ReadableStream {
+        js_sys::Reflect::get(self, &JsValue::from("readable"))
+            .unwrap()
+            .into()
+    }
+
+    pub fn writable(&self) -> WritableStream {
+        js_sys::Reflect::get(self, &JsValue::from("writable"))
+            .unwrap()
+            .into()
+    }
+}
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends=js_sys::Object)]
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub type ReadableStream;
+
+    #[wasm_bindgen(method, js_name=getReader)]
+    pub fn get_reader(this: &ReadableStream) -> ReadableStreamDefaultReader;
+}
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends=js_sys::Object)]
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub type ReadableStreamDefaultReader;
+
+    #[wasm_bindgen(method)]
+    pub fn read(this: &ReadableStreamDefaultReader) -> js_sys::Promise;
+
+    #[wasm_bindgen(method, js_name=releaseLock)]
+    pub fn release_lock(this: &ReadableStreamDefaultReader);
+}
+
+impl ReadableStreamDefaultReader {
+    pub fn closed(&self) -> js_sys::Promise {
+        js_sys::Reflect::get(self, &JsValue::from("closed"))
+            .unwrap()
+            .into()
+    }
+}
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends=js_sys::Object)]
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub type WritableStream;
+
+    #[wasm_bindgen(method, js_name=getWriter)]
+    pub fn get_writer(this: &WritableStream) -> WritableStreamDefaultWriter;
+}
+
+impl WritableStream {
+    pub fn locked(&self) -> bool {
+        js_sys::Reflect::get(self, &JsValue::from("locked"))
+            .unwrap()
+            .as_bool()
+            .unwrap()
+    }
+}
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends=js_sys::Object)]
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub type WritableStreamDefaultWriter;
+
+    #[wasm_bindgen(method)]
+    pub fn write(this: &WritableStreamDefaultWriter, chunk: JsValue) -> js_sys::Promise;
+
+    #[wasm_bindgen(method)]
+    pub fn close(this: &WritableStreamDefaultWriter) -> js_sys::Promise;
+
+    #[wasm_bindgen(method, js_name=releaseLock)]
+    pub fn release_lock(this: &WritableStreamDefaultWriter);
+
+    #[wasm_bindgen(method)]
+    pub fn abort(this: &WritableStreamDefaultWriter, reason: String) -> js_sys::Promise;
+}
+
+impl WritableStreamDefaultWriter {
+    pub fn closed(&self) -> js_sys::Promise {
+        js_sys::Reflect::get(self, &JsValue::from("closed"))
+            .unwrap()
+            .into()
+    }
+
+    pub fn desired_size(&self) -> usize {
+        js_sys::Reflect::get(self, &JsValue::from("desiredSize"))
+            .unwrap()
+            .as_f64()
+            .unwrap() as usize
+    }
+}
--- a/worker-sys/src/types/socket.rs
+++ b/worker-sys/src/types/socket.rs
@@ -20,102 +20,10 @@ extern "C" {
 
     #[wasm_bindgen(method, js_name=startTls)]
     pub fn start_tls(this: &Socket) -> Socket;
-}
 
-impl Socket {
-    pub fn readable(&self) -> ReadableStream {
-        js_sys::Reflect::get(self, &JsValue::from("readable"))
-            .unwrap()
-            .into()
-    }
+    #[wasm_bindgen(method, getter = readable)]
+    pub fn readable(this: &Socket) -> web_sys::ReadableStream;
 
-    pub fn writable(&self) -> WritableStream {
-        js_sys::Reflect::get(self, &JsValue::from("writable"))
-            .unwrap()
-            .into()
-    }
-}
-
-#[wasm_bindgen]
-extern "C" {
-    #[wasm_bindgen(extends=js_sys::Object)]
-    #[derive(Debug, Clone, PartialEq, Eq)]
-    pub type ReadableStream;
-
-    #[wasm_bindgen(method, js_name=getReader)]
-    pub fn get_reader(this: &ReadableStream) -> ReadableStreamDefaultReader;
-}
-
-#[wasm_bindgen]
-extern "C" {
-    #[wasm_bindgen(extends=js_sys::Object)]
-    #[derive(Debug, Clone, PartialEq, Eq)]
-    pub type ReadableStreamDefaultReader;
-
-    #[wasm_bindgen(method)]
-    pub fn read(this: &ReadableStreamDefaultReader) -> js_sys::Promise;
-
-    #[wasm_bindgen(method, js_name=releaseLock)]
-    pub fn release_lock(this: &ReadableStreamDefaultReader);
-}
-
-impl ReadableStreamDefaultReader {
-    pub fn closed(&self) -> js_sys::Promise {
-        js_sys::Reflect::get(self, &JsValue::from("closed"))
-            .unwrap()
-            .into()
-    }
-}
-
-#[wasm_bindgen]
-extern "C" {
-    #[wasm_bindgen(extends=js_sys::Object)]
-    #[derive(Debug, Clone, PartialEq, Eq)]
-    pub type WritableStream;
-
-    #[wasm_bindgen(method, js_name=getWriter)]
-    pub fn get_writer(this: &WritableStream) -> WritableStreamDefaultWriter;
-}
-
-impl WritableStream {
-    pub fn locked(&self) -> bool {
-        js_sys::Reflect::get(self, &JsValue::from("locked"))
-            .unwrap()
-            .as_bool()
-            .unwrap()
-    }
-}
-
-#[wasm_bindgen]
-extern "C" {
-    #[wasm_bindgen(extends=js_sys::Object)]
-    #[derive(Debug, Clone, PartialEq, Eq)]
-    pub type WritableStreamDefaultWriter;
-
-    #[wasm_bindgen(method)]
-    pub fn write(this: &WritableStreamDefaultWriter, chunk: JsValue) -> js_sys::Promise;
-
-    #[wasm_bindgen(method)]
-    pub fn close(this: &WritableStreamDefaultWriter) -> js_sys::Promise;
-
-    #[wasm_bindgen(method, js_name=releaseLock)]
-    pub fn release_lock(this: &WritableStreamDefaultWriter);
-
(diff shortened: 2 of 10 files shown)


kflansburg:
@zebp do you know what is going on with the build here? `cargo test` compiles fine locally.

KianNH:
I couldn't really find any reasons *why* Mio would be compiled with all of the features disabled, and it turns out that since this project is using `edition = "2018"` that it's just a bug with the old resolver (https://doc.rust-lang.org/nightly/cargo/reference/resolver.html#resolver-versions).

Adding `resolver = "2"` to the `Cargo.toml`'s `[workspace]` seems to have fixed it?

```diff
[workspace]
members = [
    "worker",
    "worker-build",
    "worker-macros",
    "worker-sandbox",
    "worker-sys",
]
+ resolver = "2"
```

worker/src/socket.rs
```diff
@@ -0,0 +1,340 @@
...
+        JsFuture::from(self.inner.closed()).await?;
+        Ok(())
+    }
+
+    /// Upgrades an insecure socket to a secure one that uses TLS,
+    /// returning a new Socket. Note that in order to call this method,
+    /// you must set [`secure_transport`](SocketOptions::secure_transport)
+    /// to [`StartTls`](SecureTransport::StartTls) when initially
+    /// calling [`connect`](connect) to create the socket.
+    pub async fn start_tls(self) -> Socket {
```
dom96, line 94: Should this be an `async` function? the JS API doesn't return a Promise here

worker/src/socket.rs
```diff
@@ -0,0 +1,340 @@
...
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> std::task::Poll<std::io::Result<()>> {
+        // Writes as much as possible to buf, and stores the rest in internal buffer
+        fn handle_data(
+            buf: &mut tokio::io::ReadBuf<'_>,
+            data: Vec<u8>,
+        ) -> (Reading, Poll<std::io::Result<()>>) {
+            let idx = buf.remaining().min(data.len());
+            console_log!("Ready {}/{} bytes", data.len(), idx);
```
dom96, line 126: Remove these (and below)

worker/src/socket.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn poll_read(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> std::task::Poll<std::io::Result<()>> {
+        // Writes as much as possible to buf, and stores the rest in internal buffer
+        fn handle_data(
+            buf: &mut tokio::io::ReadBuf<'_>,
+            data: Vec<u8>,
+        ) -> (Reading, Poll<std::io::Result<()>>) {
```
dom96, line 124: We ideally need test coverage for this code: buffer management is tricky.

kflansburg:
Thanks, I'll try to dig into what is happening here, I've built a worker with this change, so somehow targeting Wasm in that context is not an issue.

worker/src/socket.rs
```diff
@@ -0,0 +1,340 @@
...
+            let idx = buf.remaining().min(data.len());
+            console_log!("Ready {}/{} bytes", data.len(), idx);
+            buf.put_slice(&data[..idx]);
+            if idx == data.len() {
+                console_log!("Read to end");
+                (Reading::None, Poll::Ready(Ok(())))
+            } else {
+                console_log!("Storing {} in internal buffer", data.len() - idx);
+                let text = std::str::from_utf8(&data[idx..]).unwrap();
```
dom96, line 133: This assumes UTF8. Do we want to leave that assumption to the client code and let it handle the decoding? (Similar to how it's done in JS with TextDecoder)

worker/src/socket.rs
```diff
@@ -0,0 +1,340 @@
...
+        self
+    }
+
+    pub fn secure_transport(mut self, secure_transport: SecureTransport) -> Self {
+        self.options.secure_transport = secure_transport;
+        self
+    }
+
+    /// Open the connection, returning a [`Socket`](Socket).
+    /// You must set both `hostname` and `port` before invoking this method.
```
dom96, line 313: how about we make these required arguments to `new` instead?

dom96 (commented):
Really cool!

worker/src/socket.rs
```diff
@@ -0,0 +1,340 @@
...
+        Self::None
+    }
+}
+
+/// Represents an outbound TCP connection from your Worker.
+pub struct Socket {
+    inner: worker_sys::Socket,
+    // TODO: Box this stuff so it can be sync / send?
+    _writable: worker_sys::WritableStream,
+    writer: worker_sys::WritableStreamDefaultWriter,
```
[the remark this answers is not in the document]
kflansburg: I originally had it create a writer and reader on ever call to `poll_read`, which also seems wrong. With my current implementation, I think I could take the lock only when a read future is pending, and release after, but its not super clear to me what this buys you since everything is privately wrapped up in the `Socket` struct and I think most people will interact with it via `AsyncRead + AsyncWrite`.

worker/src/socket.rs
kflansburg: Does it? Our docs do not indicate that. https://developers.cloudflare.com/workers/runtime-apis/tcp-sockets/#socket

I have not tested this.

worker/src/socket.rs
kflansburg: Yep, will do on cleanup

worker/src/socket.rs
kflansburg: This was only used for debugging, I will not be doing this in practice.

worker/src/socket.rs
kflansburg: Possibly, I need to review the Rust builder pattern to see what is idiomatic here.

KianNH:
IMG_8643

🎉

If we can keep the resolver change in this PR then I can look at moving to the 2021 edition in a follow-up PR separately

worker/src/socket.rs
kflansburg: Ok, there seem to be two options, one involves creating distinct types so that the compiler can detect a builder missing the required fields (https://www.greyblake.com/blog/builder-with-typestate-in-rust/). I have to say I don't _love_ this but it at least eliminates runtime errors. 

Another thought I have is to just require the `host` and `port` as arguments to the builder finalizer (i.e. `Builder::connect`). This feels more ergonomic to me vs. them having to compile and decipher the type error. 

Thoughts?

worker/src/socket.rs
dom96: Hm, am I missing something? The docs don't show it returning a Promise:
Screenshot 2023-05-24 at 19 24 48

kflansburg: Oh, I see what you are saying, I thought you meant that it should be async, yep, will fix.

worker-sys/src/types/socket.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    #[wasm_bindgen(method)]
+    pub fn closed(this: &Socket) -> js_sys::Promise;
+
+    #[wasm_bindgen(method, js_name=startTls)]
+    pub fn start_tls(this: &Socket) -> Socket;
+}
+
+impl Socket {
+    pub fn readable(&self) -> ReadableStream {
```
zebp, line 26: We probably want to just use a wasm-bindgen macro with the [getter](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/getter-and-setter.html) flag.

worker-sys/src/types/socket.rs
```diff
@@ -0,0 +1,121 @@
...
+            .unwrap()
+            .into()
+    }
+}
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends=js_sys::Object)]
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub type ReadableStream;
```
zebp, line 43: We shouldn't redefine this and instead use [ReadableStream](https://docs.rs/web-sys/latest/web_sys/struct.ReadableStream.html) from the existing web-sys crate.

worker/src/socket.rs
```diff
@@ -0,0 +1,368 @@
...
+    pub fn new() -> Self {
+        ConnectionBuilder {
+            hostname: None,
+            port: None,
+            options: SocketOptions::default(),
+        }
+    }
+
+    /// Set the hostname to connect to. Example: `cloudflare.com`.
+    pub fn host(mut self, host: &str) -> Self {
```
zebp, line 269: nit: it might be nice to use `Into<String>` for the host so we accept `String`s as well.

```rust
pub fn host(mut self, host: impl Into<String>) -> Self {
    self.hostname = Some(host.to_string());
    self
}

File: worker/src/socket.rs [hunk 10]
```diff
@@ -0,0 +1,368 @@
...
+        self
+    }
+
+    pub fn secure_transport(mut self, secure_transport: SecureTransport) -> Self {
+        self.options.secure_transport = secure_transport;
+        self
+    }
+
+    /// Open the connection to `hostname` on port `port`, returning a [`Socket`](Socket).
+    pub fn connect(self, hostname: &str, port: u16) -> Result<Socket> {
```
zebp, line 293: Shouldn't this either use the `host` and `port` builder methods and not accept any args here or those methods not exist?

kflansburg:
Some notes from testing:
* `sqlx` requires some TLS implementation (I guess it manages it itself), so that may need some further investigation.
* `tokio-postgres` requires `rand`, not sure why, but that will need to be bypassed.
* `Socket` is not `Send + Sync` (even when using `Box`). This seems to be a weird quirk of `wasm-bindgen-futures` which prevents `Socket` from being used with `hyper`. 

Thoughts on the third one @zebp? Have you run into this before?

TonyGiorgio:
Just wanted to `tACK` a2b09e961386948d8664d556b6135281386c64f6 - I'm testing this out in a project moving from using axum to cloudflare workers doing a websocket to tcp socket proxy and this is working well locally. The APIs are incredibly similar so it made copy-pasting work pretty well by just swapping out a few libs.

zebp:
> * `Socket` is not `Send + Sync` (even when using `Box`). This seems to be a weird quirk of `wasm-bindgen-futures` / `JsValue` which prevents `Socket` from being used with `hyper`.

We run into this quite a bit. Because the runtime only ever has one thread of execution for the Worker and we use `SharedArrayBuffer` we just do an `unsafe impl Send/Sync`.

kflansburg:
I do have this working with `hyper`:

**Cargo.toml**
```toml
hyper = { version="0.14", features=['http1', 'client'], default-fetaures=false }
tokio = {version = "1.0", default-features=false, features=['io-util', 'macros']}
```

**main.rs**
```rust
async fn make_request(
    mut sender: hyper::client::conn::SendRequest<hyper::Body>,
    request: hyper::Request<hyper::Body>,
) -> Response {
    let hyper_response = sender.send_request(request).await.expect("send_request");
    let buf = hyper::body::to_bytes(hyper_response)
        .await
        .expect("collect body");
    let text = std::str::from_utf8(&buf).expect("utf-8");
    let mut response = Response::ok(text).unwrap();
    response
        .headers_mut()
        .append("Content-Type", "text/html")
        .ok();
    response
}

#[event(fetch)]
async fn main(_req: Request, _env: Env, _ctx: Context) -> Result<Response> {
    set_panic_hook();
    console_log!("Connecting...");
    let socket = Socket::builder().connect("example.com", 80)?;
    console_log!("Got socket!");

    let (request_sender, connection) = hyper::client::conn::handshake(socket)
        .await
        .expect("handshake");

    let request = hyper::Request::builder()
        // We need to manually add the host header because SendRequest does not
        .header("Host", "example.com")
        .method("GET")
        .body(hyper::Body::from(""))
        .expect("build request");

    tokio::select!(
        res = connection => {
            console_error!("Connection exited: {:?}", res);
            Err(worker::Error::RustError("Connection exited".to_string()))
        },
        response = make_request(request_sender, request) => {
            Ok(response)
        }
    )
}
```

worker-build/src/js/shim.js
```diff
@@ -4,4 +4,4 @@ export * from "./index_bg.js";
 // Run the worker's initialization function.
 imports.start?.();
 
-export default { fetch: imports.fetch, scheduled: imports.scheduled, queue: imports.queue };
+export default { connect: imports.connect, fetch: imports.fetch, scheduled: imports.scheduled, queue: imports.queue };
```
KianNH, line 7: These are the exported event handlers, I'm not *aware* of any inbound TCP handlers yet so unless this is a sneak peek into inbound socket workers (👀), it should be fine to remove the addition haha

kflansburg: No, I wasn't sure how the global scope was constructed, I will remove it.

worker-sys/src/types/socket.rs
```diff
@@ -0,0 +1,29 @@
...
+    #[wasm_bindgen(method)]
+    pub fn close(this: &Socket) -> js_sys::Promise;
+
+    #[wasm_bindgen(method)]
+    pub fn closed(this: &Socket) -> js_sys::Promise;
+
+    #[wasm_bindgen(method, js_name=startTls)]
+    pub fn start_tls(this: &Socket) -> Socket;
+
+    #[wasm_bindgen(method, getter = readable)]
```
KianNH, line 24: It should be okay to just specify `getter`, rather than `getter = foo`, as the function name (`readable`) is the same as JavaScript property.

https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/getter-and-setter.html

kflansburg:
Full `hyper` example: https://github.com/kflansburg/hyper-on-workers

kflansburg:
I think I'm pretty close to having `hyper::Client` working, but unfortunately it appears to rely on `Instant::now`, which I don't think we can patch. `v1.0` might have some plumbing for configuring a custom timer, so this might have to wait until that stabilizes. 

https://github.com/kflansburg/hyper-on-workers/blob/client/src/lib.rs

worker/src/socket.rs
```diff
@@ -0,0 +1,432 @@
...
+        let inner = self.inner.start_tls();
+        Socket::new(inner)
+    }
+
+    pub fn builder() -> ConnectionBuilder {
+        ConnectionBuilder::default()
+    }
+
+    fn handle_write_future(
+        cx: &mut std::task::Context<'_>,
```
zebp, line 100: nit: I think this code would be easier to read if we used `use` to import instead of a bunch of qualified paths.

worker/src/socket.rs
```diff
@@ -0,0 +1,432 @@
...
+    let s = if value.is_string() {
+        value.as_string().unwrap()
+    } else if let Some(value) = value.dyn_ref::<js_sys::Error>() {
+        value.to_string().into()
+    } else {
+        format!("Error interpreting JsError: {:?}", value)
+    };
+    std::io::Error::new(std::io::ErrorKind::Other, s)
+}
+impl tokio::io::AsyncRead for Socket {
```
zebp, line 128: same nit here, I think a lot of these should use imports instead of qualified paths. eg:

```rust
use std::io::Result as IOResult;
use std::pin::Pin;
use std::task::{Context, Poll};

use tokio::io::{ReadBuf, AsyncRead};

// ....

impl AsyncRead for Socket {
    fn poll_read(
        mut self: Pin<&mut Self>,
        cx: &mut Context<'_>,
        buf: &mut ReadBuf<'_>,
    ) -> Poll<IOResult<()>> {
        // ...
    }
```

zebp (changes requested):
Mostly looks good, some nits in terms of style but we also need to add some testing via `worker-sandbox`. This will probably include a test that just spawns a tcp listener that'll expect a connection from the worker, which should be simple but it might involve us upgrading to miniflare 3 for the janky test suite.

kflansburg:
Manual test for now located here: https://github.com/kflansburg/workers-rs-sockets-test

@dom96 @zebp

zebp:
Still need to do another review pass but I played around with this to see if I could get [tokio-postgres](https://github.com/sfackler/rust-postgres/tree/master/tokio-postgres) working and I was able to! The changes necessary to tokio-postgres are [here](https://github.com/zebp/rust-postgres) if interested. This is really cool

```rs
use futures_util::FutureExt;
use tokio_postgres::{Config, NoTls};
use worker::*;

#[event(start)]
fn init() {
}

#[event(fetch)]
async fn main(req: Request, env: Env, _ctx: Context) -> Result<Response> {
    let router = Router::new().get_async("/", |_, ctx| async move {
        let socket = Socket::builder()
            .secure_transport(SecureTransport::Off)
            .connect("0.0.0.0", 5432)?;

        let (client, connection) = Config::new()
            .user("postgres")
            .password("postgres")
            .dbname("postgres")
            .connect_raw(socket, NoTls)
            .await
            .unwrap();

        let conn = connection.map(|r| r.unwrap());
        wasm_bindgen_futures::spawn_local(conn);

        let rows = client.query("select 10;", &[]).await.unwrap();
        let value: i32 = rows[0].get(0);

        Response::ok(format!("Hello from Rust! {}", value))
    });

    router.run(req, env).await
}
```

kflansburg:
Awesome, I’m really excited about this!

zebp (approved):

PR #331, opened 2023-06-05
Add documentation for testing using Miniflare

aseure:
**🔖 Summary**

This commit adds some documentation to the README on how to configure the latest version of Miniflare (`3.0.1` at the time of writing) to test Rust-based, WASM-generated workers programatically in JavaScript / TypeScript tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,59 @@ pub async fn main(message_batch: MessageBatch<MyType>, env: Env, _ctx: Context) 
 }
 ```
 
+## Testing with Miniflare
+
+In order to test your Rust worker locally, the best approach is to use
+[Miniflare](https://github.com/cloudflare/miniflare). However, because Miniflare
+is a Node package, you will need to write your end-to-end tests in JavaScript or
+TypeScript in your project. The official documentation for writing tests using
+Miniflare is [available here](https://miniflare.dev/testing). This documentation
+being focused on JavaScript / TypeScript codebase, you will need to configure
+as follows to make it work with your Rust-based, WASM-generated worker:
+
+### Step 1: Add Wrangler and Miniflare to your `devDependencies`
+
+```sh
+npm install --save-dev wrangler miniflare
+```
+
+### Step 2: Build your worker before running the tests
+
+Make sure that your worker is built before running your tests by calling the
+following in your build chain:
+
+```sh
+wrangler deploy --dry-run
+```
+
+By default, this should build your worker under the `./build/` directory at the
+root of your project.
+
+### Step 3: Configure your Miniflare instance in your JavaScript / TypeScript tests
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -289,6 +289,51 @@ pub async fn main(message_batch: MessageBatch<MyType>, env: Env, _ctx: Context)
...
+is a Node package, you will need to write your end-to-end tests in JavaScript or
+TypeScript in your project. The official documentation for writing tests using
+Miniflare is [available here](https://miniflare.dev/testing). This documentation
+being focused on JavaScript / TypeScript codebase, you will need to configure
+as follows to make it work with your Rust-based, WASM-generated worker:
+
+### Step 1: Add Wrangler and Miniflare to your `devDependencies`
+
+```sh
+npm add -D wrangler@3.0.1 miniflare@3.0.1
```
mrbbot, line 305: I'd remove the version constraints here, and use `npm install` instead of `npm add`, which tends to be used more.
```suggestion
npm install --save-dev wrangler miniflare
```

README.md
```diff
@@ -289,6 +289,51 @@ pub async fn main(message_batch: MessageBatch<MyType>, env: Env, _ctx: Context)
...
+resolve the `*.wasm` file imported from that JavaScript worker:
+
+```js
+const miniflare = new Miniflare({
+  scriptPath: "./build/worker/shim.mjs",
+  modules: true,
+  modulesRules: [
+    { type: "CompiledWasm", include: ["**/*.wasm"], fallthrough: true }
+  ]
+});
```
mrbbot, line 334: Maybe add a basic test here too, and show importing `miniflare`? I'd also include the filename, as `*.mjs` will be required to use `import` syntax, assuming users don't have `"type": "module"` in their `package.json`s. Something like...

```suggestion
// test.mjs
import assert from "node:assert";
import { Miniflare } from "miniflare";

const mf = new Miniflare({
  scriptPath: "./build/worker/shim.mjs",
  modules: true,
  modulesRules: [
    { type: "CompiledWasm", include: ["**/*.wasm"], fallthrough: true }
  ]
});

const res = await mf.dispatchFetch("http://localhost");
assert(res.ok);
assert.strictEqual(await res.text(), "Hello, World!");
```

mrbbot (commented):
Nice! 👍 Thanks for doing this!

README.md
zebp: Also a reference to https://miniflare.dev/testing for guides for other test frameworks is a good idea.

aseure: This reference is already shown in the description above.

zebp: Right you are... Can't believe I missed that!

mrbbot (approved):

PR #335, opened 2023-06-13
Update queue bindings

jdon:
This PR updates the queue bindings to use the latest features and makes a breaking change to the handling of messages. 

Breaking change:
* `message.id` , `message.timestamp` and `message.body` have been changed to function calls, 
  * i.e `message.id()` and `message.timestamp()`.
  * Internally we now store the `MessageSys` inside the `message`. So rather than eagerly converting the ids and timestamp, we now only do it when the function is called.
* `iter()` fn on `MessageBatch` no longer has a lifetime.

New queue features added:
* `ack()` and `retry()` are now available on the individual message
  *  `retry_with_options()` was added to allow specifying the `QueueRetryOptions`
* `ack_all()` is now available on the message batch
* `retry_all_with_options()` was added to the message batch to allow specifying the `QueueRetryOptions`
* `send_batch()` is now available on the queue
* It's now possible to set `QueueSendOptions` when sending a message
  * New builder structs `MessageBuilder` & `RawMessageBuilder` added to allow for the creation of messages with options.

Other changes:
* Added queue example to the examples folder
* The signature of `Queue::send` was changed to take `Into<SendMessage<T>` where `T: Serialize`, this change was made to allow users to optionally specify `QueueSendOptions` using the new builders.
  * This isn't a breaking a change as I've added `impl<T: Serialize> From<T> for SendMessage<T>`, so users can send the same values as before.
* Allow for sending and receiving of the raw `JsValue`:
  * `Message` now has a `raw_body` function that returns a `JsValue`.
  * `raw_iter()` fn has been added to `MessageBatch`.
  * `Queue` now has a `send_raw` and `send_raw_batch` function that allows sending of `JsValue`.
  * I was unable to get the `bytes` content type working with batch messages, so I've not added that to the `QueueContentType` enum. I _think_ the issue is bytes content types needs a [js_sys](https://rustwasm.github.io/wasm-bindgen/api/js_sys/index.html)::[ArrayBuffer](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.ArrayBuffer.html#) but [js_sys](https://rustwasm.github.io/wasm-bindgen/api/js_sys/index.html)::[Reflect](https://rustwasm.github.io/wasm-bindgen/api/js_sys/Reflect/index.html)::[set](https://rustwasm.github.io/wasm-bindgen/api/js_sys/Reflect/fn.set.html#) sets an object.


worker/src/queue.rs
```diff
@@ -1,124 +1,109 @@
...
-use worker_sys::{MessageBatch as MessageBatchSys, Queue as EdgeQueue};
-
-static BODY_KEY_STR: &str = "body";
-static ID_KEY_STR: &str = "id";
-static TIMESTAMP_KEY_STR: &str = "timestamp";
+use worker_sys::{Message as MessageSys, MessageBatch as MessageBatchSys, Queue as EdgeQueue};
 
-pub struct MessageBatch<T> {
+/// A batch of messages that are sent to a consumer Worker.
+pub struct MessageBatch {
```
zebp, line 9: I'm not too fond of the removal of the body type generic. I get that we want to support multiple types for the body but that's currently possible using [serde's enum representations](https://serde.rs/enum-representations.html#untagged) unless theres some edge case I'm not aware of. Besides that I think it's good that we stay in-line with the TypeScript type definition that does use a generic.

worker/src/queue.rs
```diff
@@ -1,124 +1,109 @@
...
+
+/// A message that is sent to a consumer Worker.
+pub struct Message {
+    inner: MessageSys,
+}
 
-    /// An array of messages in the batch. Ordering of messages is not guaranteed.
-    pub fn messages(&self) -> Result<Vec<Message<T>>>
+impl Message {
+    pub fn new(message_sys: MessageSys) -> Self {
```
zebp, line 54: This should probably be an `impl From<MessageSys>`

zebp (changes requested):

worker/src/queue.rs
jdon: Done

worker/src/queue.rs
jdon: Thanks for the reviewing.

I've changed it back to `MessageBatch<T>` and also re-added the `messages()` fn, so it'll match the typescript types.

The main change is now the `MessageBatch<T>` does require `T: DeserializeOwned`, this means that the `iter()` fn now gives `Item=RawMessage`. If you don't like this, happy to I can re-add the `DeserializeOwned` bound and remove the whole `RawMessage` part.

zebp: I think the best approach to keep both the flexibility and the ease of use is to rename the current `iter` method to `raw_iter()` (and rename `MessageIter` to `RawMessageIter`) and then add:

```rs
impl<T: DeserializeOwned> MessageBatch<T> {
    /// Iterator for messages in the message batch. Ordering of messages is not guaranteed.
    pub fn iter(&self) -> MessageIter<T> {
        let messages = self.inner.messages();
        MessageIter {
            range: 0..messages.length(),
            array: messages,
        }
    }
}
```

jdon: Good idea, I've updated it to follow your suggestion.

jakubadamw:
@zebp any chance this could be reviewed and merged? It seems like a useful change, and the review feedback has been addressed?

jdon:
@zebp & @kflansburg, I've updated this PR with the latest changes from main, it would be great if we could get this reviewed and merged at some point🙏

kflansburg (commented):
I think this looks good. Will probably have it go out in its own release.

worker-sys/src/types/queue.rs
```diff
@@ -23,5 +48,8 @@ extern "C" {
     pub type Queue;
 
     #[wasm_bindgen(method)]
-    pub fn send(this: &Queue, message: JsValue) -> js_sys::Promise;
+    pub fn send(this: &Queue, mesage: JsValue) -> js_sys::Promise;
+
+    #[wasm_bindgen(method)]
+    pub fn send_batch(this: &Queue, mesage: JsValue) -> js_sys::Promise;
```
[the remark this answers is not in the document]
jdon: > Is mesage a type (here and above)

That's a typo 😅 , it should be `message`, I've fixed that in https://github.com/cloudflare/workers-rs/pull/335/commits/00a3ab17c8f8e76315373bb254d64441e7d67608

> should send_batch take a list of messages (or is JsValue intended to be an array of messages?

Good point, I had a look at the latest types and it should be an array, so I've updated it in https://github.com/cloudflare/workers-rs/pull/335/commits/00a3ab17c8f8e76315373bb254d64441e7d67608

jdon:
I had a fresh look at the latest types and realized there had been more changes to queues since this PR was created.

I've made some more changes,  so PR should now be up to date with the latest [2024-03-26](https://developers.cloudflare.com/queues/platform/changelog/#2024-03-26) changes

worker/src/queue.rs
```diff
@@ -1,124 +1,255 @@
...
+    inner: MessageSys,
+}
+
+impl RawMessage {
+    /// The body of the message.
+    pub fn body<T>(&self) -> Result<T>
+    where
+        T: DeserializeOwned,
+    {
+        Ok(serde_wasm_bindgen::from_value(self.raw_body())?)
```
kflansburg, line 122: I think I would have expected this to return a `JsValue` and then not have the `Deserialize` trait bound. Then, when its converted to a `Message` above, it is deserialized.

worker/src/queue.rs
```diff
@@ -179,15 +342,343 @@ impl AsRef<JsValue> for Queue {
...
+
+    #[must_use]
+    /// The number of seconds to delay a message for within the queue, before it can be delivered to a consumer
+    pub fn with_delay_seconds(mut self, delay_seconds: u32) -> Self {
+        self.delay_seconds = Some(delay_seconds);
+        self
+    }
+
+    /// Build the message with a content type.
+    pub fn build_with_content_type(self, content_type: QueueContentType) -> SendMessage<JsValue> {
```
kflansburg, line 440: This deviates from `MessageBuilder` because they must specify a `content_type`?

worker/src/queue.rs
```diff
@@ -179,15 +342,343 @@ impl AsRef<JsValue> for Queue {
...
+            message: self.message,
+            options: Some(QueueSendOptions {
+                content_type: Some(content_type),
+                delay_seconds: self.delay_seconds,
+            }),
+        }
+    }
+}
+
+pub struct SendMessage<T> {
```
kflansburg, line 451: The existence of this type might be confusing, could we add documentation? I'm also a little bit worried by the fact that this could either be a `JsValue` or `T`.

worker/src/queue.rs
```diff
@@ -179,15 +342,343 @@ impl AsRef<JsValue> for Queue {
...
+    message: T,
+    options: Option<QueueSendOptions>,
+}
+
+impl<T: Serialize> SendMessage<T> {
+    pub fn into_raw_message(self) -> Result<SendMessage<JsValue>> {
```
kflansburg, line 457: This is confusing, I would expect it to return `Result<RawMessage>`. Could this be a private method? (Same for BatchSendMessage)

worker/src/queue.rs
```diff
@@ -179,15 +342,343 @@ impl AsRef<JsValue> for Queue {
...
+}
+
+pub struct BatchMessageBuilder<T> {
+    messages: Vec<SendMessage<T>>,
+    delay_seconds: Option<u32>,
+}
+
+impl<T> BatchMessageBuilder<T> {
+    /// Creates a new batch message builder.
+    pub fn message<U: Into<SendMessage<T>>>(message: U) -> Self {
```
kflansburg, line 495: I think its a little weird to have both the option to add messages in the constructor and using the builder pattern.

worker/src/queue.rs
```diff
@@ -179,15 +342,343 @@ impl AsRef<JsValue> for Queue {
...
+impl<T: Serialize> SendMessage<T> {
+    pub fn into_raw_message(self) -> Result<SendMessage<JsValue>> {
+        Ok(SendMessage {
+            message: serde_wasm_bindgen::to_value(&self.message)?,
+            options: self.options,
+        })
+    }
+}
+
+impl<T: Serialize> From<T> for SendMessage<T> {
```
kflansburg, line 465: Could / Should you also implement `From<JsValue>` for easily sending raw messages? Or would it be impossible to discern V8 types?

kflansburg (commented):
Thanks for putting this together, I think this is going to be a really awesome improvement. Just have a few questions.

worker/src/queue.rs
jdon: Yes, with the MessageBuilder I think it makes sense for it to default to `json` content type. For raw messages I wasn't sure what the default should be, so I made it so the `content_type` must be specified here.

Happy to change it, if you think we should have a default content type for raw messages.

worker/src/queue.rs
jdon: > Could you also implement From `<JsValue>`

Rust won't allow it as the upstream crate may add a new impl of trait `serde::Serialize` for type `wasm_bindgen::JsValue` in future versions

> Should you also implement From<JsValue> for easily sending raw messages? Or would it be impossible to discern V8 types?

I think it's the same issue as in https://github.com/cloudflare/workers-rs/pull/335#discussion_r1543841021 , AFAIK we can't discern the type (I only had a quick look at the js_sys, so I could be missing something), so we would have to decide on a default content type for raw messages.

worker/src/queue.rs
jdon: 👍, The `BatchMessageBuilder` now takes nothing in the `new` constructor and users can add messages using `message` or `messages` functions.

I've also updated all the other builders use `new` as a constructor so we are more consistent across them.

worker/src/queue.rs
jdon: > This is confusing, I would expect it to return Result<RawMessage>

I've updated the names to make it clearer that this is for send messages.

> Could this be a private method? (Same for BatchSendMessage)

Yes, these should have been private 🤦 , I only added these for usage in the send messages.

worker/src/queue.rs
jdon: The intention here is for this to be a mostly internal type, it needs to be `pub` to be in the function signatures but users can't construct it directly as the fields are private.

I've added doc to clarify this.

worker/src/queue.rs
jdon: 👍, done in https://github.com/cloudflare/workers-rs/pull/335/commits/09795506f4cefb4c9d69f95856f09ba2be4efe8d

kflansburg (approved):

PR #339, opened 2023-06-15
chore: various typo fixes

KianNH:
Just various typo fixes I found around the codebase whilst working on another PR.

### DIFF
--- a/worker/src/r2/builder.rs
+++ b/worker/src/r2/builder.rs
@@ -5,7 +5,7 @@ use wasm_bindgen::{JsCast, JsValue};
 use wasm_bindgen_futures::JsFuture;
 use worker_sys::{
     R2Bucket as EdgeR2Bucket, R2HttpMetadata as R2HttpMetadataSys,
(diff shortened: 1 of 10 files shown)


worker/src/r2/builder.rs
```diff
@@ -164,12 +164,12 @@ impl<'bucket> PutOptionsBuilder<'bucket> {
     }
 
     /// A map of custom, user-defined metadata that will be stored with the object.
-    pub fn custom_metdata(mut self, metadata: impl Into<HashMap<String, String>>) -> Self {
+    pub fn custom_metadata(mut self, metadata: impl Into<HashMap<String, String>>) -> Self {
```
KianNH, line 167: This change (and the same on Line 233 of the same file) would be a breaking change as the function name exposed to the user has changed.

zebp: That's unfortunate..... I'm fine with the break change since 0.1.0 is going to have quite a few of them.

zebp (approved):

PR #343, opened 2023-06-20
Updated matchit dependency to 0.6.0

bobthesponge10:
[matchit 0.6.0](https://github.com/ibraheemdev/matchit/releases/tag/v0.6.0) allows catch all routes to overlap with normal routes. So routes like `/*foo` and `/bar` can both exist with `/bar` getting priority.

The closest behavior that currently exists allows only for routes like `/:foo` and `/bar` to overlap. This is less usable due to `/:foo` only matching till the next slash and not matching a full path.

The new behavior should be preferred to allow for path like the following:
`/`
`/api/:function`
`/*filepath`


kflansburg:
Fixed in #450

PR #346, opened 2023-06-30
upgrade libs, add debug to D1PreparedStatement

gdoteof:
- Node->Router in `matchit` upgrade.


kflansburg:
Deps were updated in #450. Feel free to re-open with just the `Debug` change.

Issue #353, opened 2023-07-09
[BUG] unsafe unwrapping of cloudflare properties in `cf()` method on request struct


asimpletune:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What version of `workers-rs` are you using?

0.0.17

### What version of `wrangler` are you using?

3.1.2

### Describe the bug

https://github.com/cloudflare/workerd/issues/825 does a pretty good job explaining the issue, but to summarize: it's possible for a bound rust worker's request to have `NONE` as a value for `cf`, however the `cf()` method signature guarantees a `Cf` type and it does that by making an unsafe `.unwrap()`. 

### Steps To Reproduce

See repro steps in https://github.com/cloudflare/workerd/issues/825

PR #354, opened 2023-07-09
Make `cf()` return an option

asimpletune:
This is in response to issue #353 

I understand that this change is technically a breaking change. Another option is to silently add the `cf` field to any requests constructed. There are situations, like for the mail events, where there is no request to copy from. 

Bearing this all in mind, the `cf` field _is_ wrapped in `Option` on the request struct, and therefore having the `cf()` method also return an `Option` would make sense. 

## small example

There're more details in the linked issue (in the workerd repo), but the issue can be seen for example in an upstream TS service calls a bound rust worker. Something like:

```typescript
let new_request = new Request("https://foo.com", { method: "POST", body: request.body })
```

Here a new request is being made, but the `cf` property is not being copied over.  On the other hand something like

```typescript
env.B.fetch(new Request("https://foo.com", request))
```

copies over the `cf` properties from the initial request from the client.

### DIFF
--- a/worker/src/request.rs
+++ b/worker/src/request.rs
@@ -204,8 +204,9 @@ impl Request {
     }
 
     /// Access this request's Cloudflare-specific properties.
-    pub fn cf(&self) -> &Cf {
-        self.cf.as_ref().unwrap()
+    /// **Note:** returns an Option because the assumption that the struct `Cf` will always exists can not be relied upon
+    pub fn cf(&self) -> Option<&Cf> {
+        self.cf.as_ref()
     }
 
     /// The HTTP Method associated with this `Request`.
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ pub async fn main(req: Request, env: Env, _ctx: worker::Context) -> Result<Respo
         "{} {}, located at: {:?}, within: {}",
         req.method().to_string(),
         req.path(),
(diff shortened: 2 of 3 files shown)


worker/src/request.rs
```diff
@@ -204,8 +204,8 @@ impl Request {
     }
 
     /// Access this request's Cloudflare-specific properties.
-    pub fn cf(&self) -> &Cf {
-        self.cf.as_ref().unwrap()
+    pub fn cf(&self) -> Option<&Cf> {
```
zebp, line 207: If we're going to make a breaking change can you update the documentation on why it now returns an `Option`, I think this is a necessary change but I don't want people to be surprised.

zebp (changes requested):

worker/src/request.rs
asimpletune: Sure thing, I updated the docs as well as the examples in the README.

asimpletune: @zebp are the documentation updates I did sufficient?

zebp:
Sorry this took so long, I was on vacation and then had a nasty deadline I had to hit. I want to make a minor doc change so I'll merge this into it's own branch where I can update later. Thanks for the PR!

asimpletune:
No worries, hope you enjoyed your vacation, and thanks for eventual merge.

PR #367, opened 2023-08-07
Add Storage::put_multiple_raw method to allow for runtime defined kv pairs to be stored

mendess:
The current `put_multiple` method forces the caller to use a predefined struct
which means a user of the lib can't store multiple runtime defined keys at once.
Having a `_raw` methods let's the caller construct their own object at runtime
to pass in.

### DIFF
--- a/worker-sandbox/src/test/export_durable_object.rs
+++ b/worker-sandbox/src/test/export_durable_object.rs
@@ -68,6 +68,18 @@ impl DurableObject for MyClass {
                         "Didn't get the right HashMap<String, i32> using get_multiple"
                     );
 
+                    {
+                        let bytes = Uint8Array::new_with_length(3);
+                        bytes.copy_from(b"123");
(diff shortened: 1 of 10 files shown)


worker-sandbox/src/test/export_durable_object.rs
```diff
@@ -68,6 +68,18 @@ impl DurableObject for MyClass {
                         "Didn't get the right HashMap<String, i32> using get_multiple"
                     );
 
+                    {
```
zebp, line 71: This code isn't actually hit via the testing suite sadly. I went ahead and added a commit adding testing for it using the new vitest + miniflare pattern.

zebp (approved):

worker-sandbox/src/test/export_durable_object.rs
zebp: We should rewrite all the durable object tests in the future, but this'll do for now.

PR #382, opened 2023-08-30
chore:fix d1 Example Usage

nakamura-shuta:
When I tried to use the sample code as is error, 
so I modified it to a format that can be used as it is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,9 @@ pub async fn main(request: Request, env: Env, _ctx: Context) -> Result<Response>
 		.get_async("/:id", |_, ctx| async move {
 			let id = ctx.param("id").unwrap();
 			let d1 = ctx.env.d1("things-db")?;
-			let statement = d1.prepare("SELECT * FROM things WHERE thing_id = ?1");
+			let statement = d1.prepare("SELECT * FROM things WHERE thing_id = ? LIMIT 1");
 			let query = statement.bind(&[id.into()])?;
-			let result = query.first::<Thing>(None).await?;
+			let result: Option<Thing> = query.first(None).await?;
 			match result {
 				Some(thing) => Response::from_json(&thing),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,31 +10,32 @@ Read the [Notes and FAQ](#notes-and-faq)
...
+}
 
-    Response::error("Bad Request", 400)
+#[event(fetch, respond_with_errors)]
+pub async fn main(request: Request, env: Env, _ctx: Context) -> Result<Response> {
+	Router::new()
+		.get_async("/:id", |_, ctx| async move {
+			let id = ctx.param("id").unwrap();
+			let d1 = ctx.env.d1("things-db")?;
+			let statement = d1.prepare("SELECT * FROM things WHERE thing_id = ?1");
```
film42, line 29: I think you have a typo here: `thing_id = ?1`

README.md
```diff
@@ -10,31 +10,32 @@ Read the [Notes and FAQ](#notes-and-faq)
...
+			let query = statement.bind(&[id.into()])?;
+			let result = query.first::<Thing>(None).await?;
```
film42, line 31: nit: I find these easier to read as `let result: Option<Thing> = query.first(Nonte).await?;` Since the turbo-fish with arg as parameter. Also, I think `.first()` is just a convenience wrapper around `.all()` so I think you might also need to add a `LIMIT 1` to your sql query. I'm not sure where the d1 source lives but I did find the deprecated dep on npm: https://www.npmjs.com/package/@cloudflare/d1?activeTab=code

README.md
nakamura-shuta: thanks.fix it.

README.md
nakamura-shuta: thanks.
Your suggestion looks better, fix it.

PR #388, opened 2023-09-22
Update wasm-bindgen from 0.2.86 to 0.2.87

fornwall:


rakshith-ravi:
I feel like just keeping the version at `0.2` instead of a specific version would be better.

Also, not sure why versions are pinned to specific ones

rakshith-ravi:
@cloudflare is anybody even reviewing these PRs?

jakubadamw:
@zebp any chance this could be reviewed? I'd be obliged.

rakshith-ravi:
Cloudflare's opensource projects seems to just be namesake. None of the PRs (across so many repos) have been reviewed at all and have been stale for over so long. Would it be better to perhaps maintain a fork that actually keeps itself up to date to the ecosystem?

zebp (approved):

PR #395, opened 2023-09-27
Add derive debug, clone for FormData and File

dev-ardi:
This adds derive macros for a couple of types. There's no reason not to have them.

### DIFF
--- a/worker/src/headers.rs
+++ b/worker/src/headers.rs
@@ -13,6 +13,7 @@ use worker_sys::ext::HeadersExt;
 
 /// A [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) representation used in
 /// Request and Response objects.
+#[derive(Clone, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


dev-ardi:
I've added a few more

worker/src/headers.rs
```diff
@@ -182,10 +183,3 @@ impl From<Headers> for HeaderMap {
         (&headers).into()
     }
 }
-
-impl Clone for Headers {
```
zebp, line 186: We can't use the derive-based clone implementation here as we do an explicit call to `web_sys::Headers::new_with_headers` because the inner JS headers are sometimes immutable. The derive clone will just clone a reference to those potentially immutable headers where-as the explicit clone will perform a deep copy.

zebp (changes requested):

worker/src/headers.rs
dev-ardi: I won't use workers anymore, feel free to close this PR

PR #412, opened 2023-11-10
Add Headers len and is_empty method

sengho66:
## Features
- Headers::len method
- Headers::is_empty method
- New addition of macro that's used to set or append to headers to avoid duplicated code

This PR introduces the above to let caller get the number of elements that are in the headers.

Currently the only way to get the number of elements is by iterating the headers and call the iterator's method `.count()`, the above allows caller to do so without iterating the elements.

### DIFF
--- a/worker/src/request.rs
+++ b/worker/src/request.rs
@@ -37,7 +37,7 @@ impl From<web_sys::Request> for Request {
                     }
                     u
                 }),
-            headers: Headers(req.headers()),
+            headers: req.headers().into(),
             cf: req.cf().map(Into::into),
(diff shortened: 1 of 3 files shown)


worker/src/request.rs
```diff
@@ -37,7 +37,7 @@ impl From<web_sys::Request> for Request {
                     }
                     u
                 }),
-            headers: Headers(req.headers()),
+            headers: req.headers().into(),
```
kflansburg, line 40: It's unfortunate that this is a breaking API change, but I suppose the original API was very brittle.

PR #420, opened 2023-12-01
Updated dependencies, Improved R2 Get and Put support for Checksums

spigaz:
I started using workers recently, already knowing that I would have to contribute...

So I started by updating some dependencies, to see if things improved, but a test that was flaky to me (timeout) is still flaky.

Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,12 +81,6 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.13.1"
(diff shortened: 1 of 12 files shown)


spigaz:
This fork went a wrong path from the start and not longer makes any sense, I'll have to extract the relevant changes into a fresh one.

kflansburg:
@spigaz if you rebase, the timeout test should be fixed and deps are updated. Can you re-open with just the R2 changes?

spigaz:
Sure, thanks!

Issue #426, opened 2023-12-20
[Feature] use `http::Method` instead of re-implementing `Method`.


mendess:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

This crate implements it's own [method type](https://docs.rs/worker/latest/worker/enum.Method.html) instead of using the more ecosystem standard `http::Method`. This makes it harder to integrate with `reqwest_wasm` and other crates that make use of the `http::Method` type.

Changing this now is a breaking change, which is annoying but I think it's worth it in the long run. Specially since `http` has just reached `1.0`! A non breaking alternative would be to add two from implementations in order to make the conversion easier.
```rs
impl From<http::Method> for Method {}
impl From<Method> for http::Method {}
```

What do you think?

fwqaaq:
Yeah, and it can not create custom `Method`, which makes it completely impractical when I want to build webdav.

Maybe someone knows?

kflansburg:
Fixed in #477

PR #428, opened 2023-12-24
fix: Use method of http crate instead of the original method

fwqaaq:
close: #426

In the original `worker-rs`, the `Method` was limited to only 8 standard methods. However, in reality, there are more than 8 methods defined on top of HTTP, and custom methods cannot be created. But, using JavaScript, the custom methods are possible.

I believe this is a bug, using `http` crate from Rust is a better choice, because `reqwest` and `hyper` are based on it. So I hope the offical team can address and fix this bug.


avsaase:
I think this is superseded by https://github.com/cloudflare/workers-rs/pull/286

PR #436, opened 2024-01-17
Implement Hibernatable Web Sockets API

DylanRJohnston:
Hey all, this contains my attempt at writing the bindings for the Hibernating Web Sockets API. I haven't implemented the `setWebSocketAutoResponse` suite of functions yet as I don't have immediate use for them and I'd like feedback on the implementation so far.

Supersedes #434 as I was unable to change the branch for the pull request and it contained more changes than just the Hibernatable web sockets API.

### DIFF
--- a/worker-macros/src/durable_object.rs
+++ b/worker-macros/src/durable_object.rs
@@ -24,9 +24,9 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
             #[derive(Default)]
             struct OptionalMethods {
                 has_alarm: bool,
-                has_on_message: bool,
-                has_on_close: bool,
-                has_on_error: bool,
+                has_websocket_message: bool,
+                has_websocket_close: bool,
+                has_websocket_error: bool,
             }
 
             let mut optional_methods = OptionalMethods::default();
@@ -125,16 +125,23 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
                             #method
                         }
                     },
-                    "on_message" => {
-                        optional_methods.has_on_message = true;
+                    "websocket_message" => {
+                        optional_methods.has_websocket_message = true;
 
                         let mut method = impl_method.clone();
-                        method.sig.ident = Ident::new("_on_message_raw", method.sig.ident.span());
+                        method.sig.ident = Ident::new("_websocket_message_raw", method.sig.ident.span());
                         method.vis = Visibility::Inherited;
 
                         quote! {
                             #pound[wasm_bindgen::prelude::wasm_bindgen(js_name = webSocketMessage)]
-                            pub fn _on_message(&mut self, ws: worker_sys::web_sys::WebSocket, message: String) -> js_sys::Promise {
+                            pub fn _websocket_message(&mut self, ws: worker_sys::web_sys::WebSocket, message: JsValue) -> js_sys::Promise {
+                                let ws_message = if let Some(string_message) = message.as_string() {
+                                    worker::WebSocketIncomingMessage::String(string_message)
+                                } else {
+                                    let v = js_sys::Uint8Array::new(&message).to_vec();
+                                    worker::WebSocketIncomingMessage::Binary(v)
+                                }
+                                
                                 // SAFETY:
                                 // On the surface, this is unsound because the Durable Object could be dropped
                                 // while JavaScript still has possession of the future. However,
@@ -144,7 +151,7 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
                                 let static_self: &'static mut Self = unsafe {&mut *(self as *mut _)};
 
                                 wasm_bindgen_futures::future_to_promise(async move {
-                                    static_self._on_message_raw(ws.into(), message).await.map(|_| wasm_bindgen::JsValue::NULL)
+                                    static_self._websocket_message_raw(ws.into(), ws_message).await.map(|_| wasm_bindgen::JsValue::NULL)
                                         .map_err(wasm_bindgen::JsValue::from)
                                 })
                             }
@@ -152,16 +159,16 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
                             #method
                         }
                     },
-                    "on_close" => {
-                        optional_methods.has_on_close = true;
+                    "websocket_close" => {
+                        optional_methods.has_websocket_close = true;
 
                         let mut method = impl_method.clone();
-                        method.sig.ident = Ident::new("_on_close_raw", method.sig.ident.span());
+                        method.sig.ident = Ident::new("_websocket_close_raw", method.sig.ident.span());
                         method.vis = Visibility::Inherited;
 
                         quote! {
                             #pound[wasm_bindgen::prelude::wasm_bindgen(js_name = webSocketClose)]
-                            pub fn _on_close(&mut self, ws: worker_sys::web_sys::WebSocket, code: usize, reason: String, was_clean: bool) -> js_sys::Promise {
+                            pub fn _websocket_close(&mut self, ws: worker_sys::web_sys::WebSocket, code: usize, reason: String, was_clean: bool) -> js_sys::Promise {
                                 // SAFETY:
                                 // On the surface, this is unsound because the Durable Object could be dropped
                                 // while JavaScript still has possession of the future. However,
@@ -171,7 +178,7 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
                                 let static_self: &'static mut Self = unsafe {&mut *(self as *mut _)};
 
                                 wasm_bindgen_futures::future_to_promise(async move {
-                                    static_self._on_close_raw(ws.into(), code, reason, was_clean).await.map(|_| wasm_bindgen::JsValue::NULL)
+                                    static_self._websocket_close_raw(ws.into(), code, reason, was_clean).await.map(|_| wasm_bindgen::JsValue::NULL)
                                         .map_err(wasm_bindgen::JsValue::from)
                                 })
                             }
@@ -179,16 +186,16 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
                             #method
                         }
                     },
-                    "on_error" => {
-                        optional_methods.has_on_error = true;
+                    "websocket_error" => {
+                        optional_methods.has_websocket_error = true;
 
                         let mut method = impl_method.clone();
-                        method.sig.ident = Ident::new("_on_error_raw", method.sig.ident.span());
+                        method.sig.ident = Ident::new("_websocket_error_raw", method.sig.ident.span());
                         method.vis = Visibility::Inherited;
 
                         quote! {
                             #pound[wasm_bindgen::prelude::wasm_bindgen(js_name = webSocketError)]
-                            pub fn _on_error(&mut self, ws: worker_sys::web_sys::WebSocket, error: JsValue) -> js_sys::Promise {
+                            pub fn _websocket_error(&mut self, ws: worker_sys::web_sys::WebSocket, error: JsValue) -> js_sys::Promise {
                                 // SAFETY:
                                 // On the surface, this is unsound because the Durable Object could be dropped
                                 // while JavaScript still has possession of the future. However,
@@ -198,7 +205,7 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
                                 let static_self: &'static mut Self = unsafe {&mut *(self as *mut _)};
 
                                 wasm_bindgen_futures::future_to_promise(async move {
-                                    static_self._on_error_raw(ws.into(), error.into()).await.map(|_| wasm_bindgen::JsValue::NULL)
+                                    static_self._websocket_error_raw(ws.into(), error.into()).await.map(|_| wasm_bindgen::JsValue::NULL)
                                         .map_err(wasm_bindgen::JsValue::from)
                                 })
                             }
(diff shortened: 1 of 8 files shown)


worker/src/durable.rs
```diff
@@ -737,12 +769,44 @@ impl DurableObject for Chatroom {
 }
 ```
 */
+
 #[async_trait(?Send)]
 pub trait DurableObject {
     fn new(state: State, env: Env) -> Self;
-    async fn fetch(&mut self, req: Request) -> Result<Response>;
```
DylanRJohnston, line 743: Opps, missed this when reverting the changes to `async_trait`.

worker-sys/src/types/durable_object/state.rs
```diff
@@ -15,4 +15,21 @@ extern "C" {
...
     pub fn wait_until(this: &DurableObjectState, promise: &js_sys::Promise);
+
+    #[wasm_bindgen(method, js_name=acceptWebSocket)]
+    pub fn accept_websocket(this: &DurableObjectState, ws: web_sys::WebSocket);
+
+    #[wasm_bindgen(method, js_name=acceptWebSocket)]
+    pub fn accept_websocket_with_tags(
+        this: &DurableObjectState,
+        ws: web_sys::WebSocket,
+        tags: Vec<JsValue>,
```
DylanRJohnston, line 26: `wasm_bindgen v0.2.89` supports using `Vec<String>` directly.

worker-macros/src/durable_object.rs
```diff
@@ -109,17 +124,124 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
...
+                                wasm_bindgen_futures::future_to_promise(async move {
+                                    static_self._websocket_error_raw(ws.into(), error.into()).await.map(|_| wasm_bindgen::JsValue::NULL)
+                                        .map_err(wasm_bindgen::JsValue::from)
+                                })
+                            }
+
+                            #method
+                        }
+                    },
+                    ident => panic!("Unsupported method {}, please move extra impl methods to a separate impl definition", ident)
```
DylanRJohnston, line 216: Improved feedback to user if they define extra methods that aren't supported by the macro.

worker-macros/src/durable_object.rs
```diff
@@ -108,18 +125,125 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
...
+                    "websocket_message" => {
+                        optional_methods.has_websocket_message = true;
+
+                        let mut method = impl_method.clone();
+                        method.sig.ident = Ident::new("_websocket_message_raw", method.sig.ident.span());
+                        method.vis = Visibility::Inherited;
+
+                        Ok(quote! {
+                            #pound[wasm_bindgen::prelude::wasm_bindgen(js_name = webSocketMessage)]
+                            pub fn _websocket_message(&mut self, ws: worker_sys::web_sys::WebSocket, message: JsValue) -> js_sys::Promise {
```
eric-seppanen, line 139: In macro context I think we need the fully-qualified `wasm_bindgen::JsValue`, as it might not be in the macro-caller's scope.

worker/src/durable.rs
```diff
@@ -194,6 +194,33 @@ impl State {
...
+    pub fn accept_web_socket(&self, ws: WebSocket) {
+        self.inner.accept_websocket(ws.as_ref().clone())
+    }
+
+    pub fn accept_websocket_with_tags(&self, ws: WebSocket, tags: &[&str]) {
+        let tags = tags.iter().map(|it| (*it).into()).collect();
+
+        self.inner
+            .accept_websocket_with_tags(ws.as_ref().clone(), tags);
+    }
```
eric-seppanen, line 207: Could this be a reference (`&WebSocket`) instead? It's a paper-cut for the caller (who may need to `clone` the websocket), and then this function immediately derefs and clones again...

In fact, I wonder if this could be a reference on the inner layer as well.

worker-macros/src/durable_object.rs
```diff
@@ -108,18 +125,125 @@ pub fn expand_macro(tokens: TokenStream) -> syn::Result<TokenStream> {
...
+                    "websocket_error" => {
+                        optional_methods.has_websocket_error = true;
+
+                        let mut method = impl_method.clone();
+                        method.sig.ident = Ident::new("_websocket_error_raw", method.sig.ident.span());
+                        method.vis = Visibility::Inherited;
+
+                        Ok(quote! {
+                            #pound[wasm_bindgen::prelude::wasm_bindgen(js_name = webSocketError)]
+                            pub fn _websocket_error(&mut self, ws: worker_sys::web_sys::WebSocket, error: JsValue) -> js_sys::Promise {
```
eric-seppanen, line 200: -> `wasm_bindgen::JsValue`

eric-seppanen:
I did a quick functional test of this branch and it works in my project-- I am testing 2 websockets in a durable object, using binary messages and hibernation.

worker/src/durable.rs
DylanRJohnston: Yeah good catch, I'm really not sure about ownership when it comes to `wasm_bindgen`. For what it's worth `WebSocket` is just a single u32 so cloning it a lot is extremely cheap as far as I know.

worker-macros/src/durable_object.rs
DylanRJohnston: Which I had a linter that could help me catch these mistakes.

worker-sys/src/types/durable_object/state.rs
```diff
@@ -15,4 +15,21 @@ extern "C" {
 
     #[wasm_bindgen(method, js_name=waitUntil)]
     pub fn wait_until(this: &DurableObjectState, promise: &js_sys::Promise);
+
+    #[wasm_bindgen(method, js_name=acceptWebSocket)]
+    pub fn accept_websocket(this: &DurableObjectState, ws: &web_sys::WebSocket);
```
timeswind, line 20: > Calling accept() on a WebSocket in an Object will incur duration charges for the entire time the WebSocket is connected. If you prefer, use [state.acceptWebSocket()](https://developers.cloudflare.com/durable-objects/api/websockets/#state-methods-for-websockets) instead, which will stop incurring duration charges once all event handlers finish running.

[https://developers.cloudflare.com/durable-objects/platform/pricing/](https://developers.cloudflare.com/durable-objects/platform/pricing/)

The support of` acceptWebSocket()` is critical to many existing rust-based Durable Object applications since it would dramatically lower the cost incurred by WebSocket connection currently implemented by `accept()` method. Really appreciate your work!

MellowYarker:
Hey @DylanRJohnston, I commented on the previous PR about `state.getTags(ws)`. You [previously asked](https://github.com/cloudflare/workers-rs/pull/434#issuecomment-1894841116) which version of wrangler supported it and I believe it should be [wrangler@3.26.0](https://github.com/cloudflare/workers-sdk/releases/tag/wrangler%403.26.0).

eric-seppanen:
As an experiment I added `getTags` support to my local branch, and the following works for me:
```rust
impl State {
    /// Retrieve tags from a hibernatable websocket
    pub fn get_tags(&self, websocket: &WebSocket) -> Vec<String> {
        let tags = self.inner.get_tags(websocket.as_ref());
        let tags: js_sys::Array = tags.dyn_into().expect("get_tags returned wrong type");

        tags.iter()
            .map(|tag| tag.as_string().expect("get_tags returned non-string value"))
            .collect()
    }
}
```

And some wasm_bindgen glue in worker-sys `state.rs`:
```rust
    #[wasm_bindgen(method, js_name=getTags)]
    pub fn get_tags(this: &DurableObjectState, ws: &web_sys::WebSocket) -> wasm_bindgen::JsValue;
```

Maybe the error handling is unnecessary? Probably `unchecked_into` would be fine; not sure if there's an unchecked way to do `as_string`.

DylanRJohnston:
I think you can change the signature of

```rs
pub fn get_tags(this: &DurableObjectState, ws: &web_sys::WebSocket) -> wasm_bindgen::JsValue;
```
into
```
pub fn get_tags(this: &DurableObjectState, ws: &web_sys::WebSocket) -> Vec<String>
```
and have `wasm_bindgen` do the work of converting it for you.

j-white:
Nice work & a big thanks for putting this together. I was able to switch over to use the Hibernatable Web Sockets API fairly easily with this patch!

DylanRJohnston:
Thanks @j-white, now we just gotta get someone from Cloudflare to look at it so it can get merged. It looks like there's bit a little bit of activity on the repository lately, but mainly documentation fixes. It'd be good if we could get some comms from Cloudflare on the future of the project. Perhaps it's been abandoned internally in favour of working on a WASI native solution? Who knows.

worker-sys/src/types/durable_object/state.rs
```diff
@@ -15,4 +15,21 @@ extern "C" {
...
+
+    #[wasm_bindgen(method, js_name=acceptWebSocket)]
+    pub fn accept_websocket_with_tags(
+        this: &DurableObjectState,
+        ws: &web_sys::WebSocket,
+        tags: Vec<JsValue>,
+    );
+
+    #[wasm_bindgen(method, js_name=getWebSockets)]
+    pub fn get_websockets(this: &DurableObjectState) -> Vec<web_sys::WebSocket>;
```
MellowYarker, line 30: Just curious, why do there need to be two versions of `acceptWebSocket` and `getWebSockets`? The JS and C++ methods all provide an optional/maybe for `tags`, would an `Option` not work here?

worker/src/durable.rs
```diff
@@ -737,12 +768,40 @@ impl DurableObject for Chatroom {
...
         unimplemented!("alarm() handler not implemented")
     }
+
+    #[allow(unused_variables, clippy::diverging_sub_expression)]
+    async fn websocket_message(
+        &mut self,
+        ws: WebSocket,
+        message: WebSocketIncomingMessage,
+    ) -> Result<()> {
+        unimplemented!("websocket_message() handler not implemented")
```
MellowYarker, line 789: So if I'm writing a DO in rust and I use websocket hibernation, I have to implement these otherwise when they're dispatched we panic? If you're aiming to match `Workerd`s behavior we [silently drop the event](https://github.com/cloudflare/workerd/blob/fa0e7bf74b406a241888c87f8327af917a6d7d0d/src/workerd/api/global-scope.c%2B%2B#L526), but I see the value in warning.

MellowYarker (commented):
Thanks for working on this @DylanRJohnston! For tracking purposes, the following Hibernation API methods are still unimplemented in this PR, but this is off to a great start:

- `getTags()`
- `getWebSocketAutoResponse()`
- `setWebSocketAutoResponse()`
- `getWebSocketAutoResponseTimestamp()`
- `getHibernatableWebSocketEventTimeout()`
- `setHibernatableWebSocketEventTimeout()`

My sense is these can be implemented in a later PR, although it would be nice to include `getTags()` here. I'm not usually in this repo, so I won't approve the PR myself, but I'll check in with the runtime team to see who can take a look at this.

worker/src/env.rs
```diff
@@ -79,6 +82,21 @@ impl Env {
     pub fn d1(&self, binding: &str) -> Result<D1Database> {
         self.get_binding(binding)
     }
+
+    pub fn fetch_assets(&self, req: Request) -> crate::Result<Response> {
```
kflansburg, line 86: I'm not 100% familiar with the WebSocket hibernation API, but what is this? Are users expected to have a particular binding?

worker/src/durable.rs
```diff
@@ -737,12 +768,40 @@ impl DurableObject for Chatroom {
...
     fn new(state: State, env: Env) -> Self;
+
     async fn fetch(&mut self, req: Request) -> Result<Response>;
+
     #[allow(clippy::diverging_sub_expression)]
     async fn alarm(&mut self) -> Result<Response> {
         unimplemented!("alarm() handler not implemented")
     }
+
+    #[allow(unused_variables, clippy::diverging_sub_expression)]
```
kflansburg, line 783: I think you can name the variables with `_ws` in the trait to silence `clippy`, and then implementors of the trait can still use whatever name they want.

eric-seppanen: On the other hand, the variable name will appear in the documentation, so it's a tradeoff-- is it better to have a clean variable name in the docs or to avoid unnecessary lint `allow`s?

worker-sys/src/types/durable_object/state.rs
eric-seppanen: In my opinion (I am not the author), it's a style choice but this is probably the right one (in Rust anyway).

It's slightly kinder to the caller to have two functions, and probably more idiomatic in Rust. Passing a parameter `None` hurts the readability of the code, because it's not obvious at a glance what that parameter means.

The rest of this crate seems to mostly follow this style as well, e.g. `ObjectNamespace::unique_id`/`unique_id_with_jurisdiction`, `Storage::list`/`list_with_options`...

worker/src/env.rs
DylanRJohnston: Apologies I actually have no idea where this code even came from. It's been over a month since I wrote this PR but maybe it had something to do with me trying to move to Cloudflare Pages? In any case it definitely doesn't belong in this pull request.

worker/src/durable.rs
DylanRJohnston: As Eric said, that will cause the argument to appear as `_ws` both in documentation and in inline argument name hints.

worker/src/durable.rs
DylanRJohnston: I tend to agree with you, however I'm just matching the existing implementation of `alarm` above.

```rs
#[allow(clippy::diverging_sub_expression)]
    async fn alarm(&mut self) -> Result<Response> {
        unimplemented!("alarm() handler not implemented")
    }
```

If you'd like this approach changed, I think we should also change the `alarm` implementation to match.

worker-sys/src/types/durable_object/state.rs
DylanRJohnston: Yeah it's an unfortunate impedance mismatch between function overloading in JS and the lack thereof in Rust. I'm just trying to match the style of the rest of the repo which manually "un-overloads" the functions as Eric pointed out.

worker/src/env.rs
kflansburg: I think if you remove this, then this can merge. I'm fine with the overloading solution discussed above.

DylanRJohnston: Done.

kflansburg (approved):

kflansburg (dismissed):

kflansburg (approved):

worker/src/durable.rs
```diff
@@ -194,6 +194,32 @@ impl State {
     pub fn _inner(self) -> DurableObjectState {
         self.inner
     }
+
+    pub fn accept_web_socket(&self, ws: &WebSocket) {
```
cowlicks, line 198: Noo!!!!! This should have been changed as suggested in https://github.com/cloudflare/workers-rs/pull/434#discussion_r1452829930

DylanRJohnston: Sorry mate, must have slipped through when I changed the branches over. Feel free to open a PR to change it.

PR #440, opened 2024-01-23
Fix issues with Request.clone in http types

kflansburg:


worker/src/body/body.rs
```diff
@@ -211,6 +174,21 @@ impl Body {
...
     }
+
+    pub(crate) fn into_readable_stream(self) -> Option<web_sys::ReadableStream> {
+        match self.into_inner() {
+            crate::body::BodyInner::Request(req) => req.body(),
+            crate::body::BodyInner::Response(res) => res.body(),
+            crate::body::BodyInner::Regular(s) => Some(
+                wasm_streams::ReadableStream::from_async_read(
+                    crate::body::BoxBodyReader::new(s),
+                    1024,
```
kflansburg, line 185: Idk what to set this default buffer size to

worker/src/http/request.rs
```diff
@@ -131,19 +127,8 @@ pub fn into_wasm(mut req: http::Request<Body>) -> web_sys::Request {
         let _ = r;
     }
 
-    let body = req.into_body();
-    let body = if body.is_none() {
-        None
-    } else {
-        let stream = wasm_streams::ReadableStream::from_stream(body.map(|chunk| {
```
kflansburg, line 138: Ultimately the bug seems to be in `from_stream`. `from_byte_stream` seems to work though. I'm not sure if that means a different `BoxedBody` that isn't `Bytes` will break?

SebastiaanYN: `HttpClone` is only implemented when the body holds `Bytes` so that shouldn't be an issue.

PR #448, opened 2024-02-22
Bump ureq version to 2.9.6

hack3ric:
This upgrades rustls to 0.22 and ring to 0.17, without introducing breaking changes.


kflansburg:
Done in #450

PR #460, opened 2024-03-06
Added support for R2 Checksums to GET and PUT.

spigaz:
Apparently its working fine, I only tested with my production use cases and it seemed ok.

I haven't done any testing, as I do my tests using wasm-bindgen-test, I have something like 600 on a single crate.

I talked with the wasm-bindgen-test folks and they would like to support workerd...

Then I talked to MrBot about this and he said it seemed doable once he gets the vitest miniflare 3 support back as it was on 2.

Are workers-rs tests going to return to wasm-bindgen-test or continue on vitest?

### DIFF
--- a/worker-sys/src/types/r2/checksums.rs
+++ b/worker-sys/src/types/r2/checksums.rs
@@ -23,7 +23,7 @@ impl R2Checksums {
 }
 
 fn get(obj: &Object, key: &str) -> Option<Vec<u8>> {
-    let value = Reflect::get(&obj, &key.into());
+    let value = Reflect::get(obj, &key.into());
     if value.is_err() {
         return None;
     }
@@ -50,5 +50,11 @@ impl From<Object> for R2Checksums {
             sha384: get(&obj, "sha384"),
             sha512: get(&obj, "sha512"),
         }
+    }
+}
+
+impl Default for R2Checksums {
+    fn default() -> Self {
+        Self::new()
     }
 }
--- a/worker-sys/src/types/r2/checksums.rs
+++ b/worker-sys/src/types/r2/checksums.rs
@@ -1,0 +1,54 @@
+use js_sys::{ArrayBuffer, Object, Reflect, Uint8Array};
+use wasm_bindgen::JsCast;
+
+#[derive(Debug, Clone)]
+pub struct R2Checksums {
+    pub md5: Option<Vec<u8>>,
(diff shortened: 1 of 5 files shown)


worker-sys/src/types/r2/checksums.rs
```diff
@@ -0,0 +1,54 @@
+use js_sys::{ArrayBuffer, Object, Reflect, Uint8Array};
+use wasm_bindgen::JsCast;
+
+#[derive(Debug, Clone)]
+pub struct R2Checksums {
```
kflansburg, line 5: I'm wondering if this should be implemented as distinct objects which implement a trait, to enforce that only one can be set at a time.

spigaz: That was my initial idea, because its stated that you can only set a single hash for PUT, but put internally doesn't need to use the R2Checksums, as the hashes are set using methods.

But for GET things were different.

In practice, in my tests I found out that on GET you always had the md5 hash, plus the one you added.

And the documentation states the values as optional ArrayBuffers in a R2Checksums object, so I tried to follow that instead of creating my own api.

https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#checksums

One possibility for R2Checksums was to use the same strategy as for PUT, to have an md5 hash, plus a string identifying the optional hash and the optional hash, using methods instead of fields, but it wasn't the same api and it less clear.

I thought about using a trait, specially for PUT, but just for internal use, I wasn't sure if the extra code complexity was worth it, and I intended to respect the current API as much as possible.

But we can do that no problem, just point me in the right direction.

spigaz: I also thought about using a struct for PUT internally.

```
pub enum R2ChecksumType {
    MD5,
    SHA1,
    SHA256,
    SHA384,
    SHA512,
}

struct R2Checksum {
    checksum_type: ChecksumType,
    checksum: Vec<u8>,
}
```

The trait would be something like

```
pub trait R2Checksum {
  fn algorithm(&self) -> &str; // or R2ChecksumType
  fn checksum(&self) -> Vec<u8>;
}
```

But I wasn't sure the added complexity was worth it, so I tried to minimize impact.

Although I do either option in my own code depending on the context, I organize my code differently, so that the added complexity is acceptable.

kflansburg (approved):

Issue #433, opened 2024-01-10
pinning old version of wasm_bindgen makes dependency management difficult


eric-seppanen:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What version of `workers-rs` are you using?

0.0.18

### What version of `wrangler` are you using?

n/a

### Describe the bug

The `worker` crate pins the exact version of wasm_bindgen:
```
wasm-bindgen = "=0.2.87"
```

This makes it painful to add additional dependencies, because it forces the user to manually hunt around for older versions of crates that permit that exact version of `wasm-bindgen`.

As as example, here's what it's like to create a new project and then start adding other dependencies:
```plaintext
testing$ cargo add worker
    Updating crates.io index
      Adding worker v0.0.18 to dependencies.
             Features:
             - d1
             - queue
    Updating crates.io index
    
testing$ cargo add js-sys
    Updating crates.io index
      Adding js-sys v0.3.66 to dependencies.
    Updating crates.io index
error: failed to select a version for `wasm-bindgen`.
    ... required by package `worker v0.0.18`
    ... which satisfies dependency `worker = "^0.0.18"` of package `testing v0.1.0`
versions that meet the requirements `=0.2.86` are: 0.2.86

all possible versions conflict with previously selected packages.

  previously selected package `wasm-bindgen v0.2.89`
    ... which satisfies dependency `wasm-bindgen = "^0.2.89"` of package `js-sys v0.3.66`
    ... which satisfies dependency `js-sys = "^0.3.66"` of package `testing v0.1.0`

failed to select a version for `wasm-bindgen` which could resolve this conflict
```

The error isn't terribly clear, but the only solution is to find a version of `js-sys` that will accept the same version of `wasm-bindgen` that `worker` requires. Let's try the previous version:
```plaintext
testing$ cargo add js-sys@0.3.65
    Updating crates.io index
      Adding js-sys v0.3.65 to dependencies.
    Updating crates.io index
error: failed to select a version for `wasm-bindgen`.
    ... required by package `js-sys v0.3.65`
    ... which satisfies dependency `js-sys = "^0.3.65"` of package `testing v0.1.0`
versions that meet the requirements `^0.2.88` are: 0.2.89

all possible versions conflict with previously selected packages.

  previously selected package `wasm-bindgen v0.2.86`
    ... which satisfies dependency `wasm-bindgen = "=0.2.86"` of package `worker v0.0.18`
[...]

PR #472, opened 2024-03-09
update/use workspace dependencies (wasm_bindgen, etc)

NicoZweifel:
Notes:
- related/created from https://github.com/cloudflare/workers-rs/issues/433
- updating depencencies/using workspace dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1611,9 +1611,9 @@ dependencies = [
 
 [[package]]
 name = "tempfile"
-version = "3.10.0"
+version = "3.10.1"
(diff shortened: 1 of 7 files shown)


kflansburg:
This would be awesome. There are still a few places where packages are not using the workspace dependency (worker-sys/Cargo.toml#13).

NicoZweifel:
> This would be awesome. There are still a few places where packages are not using the workspace dependency (worker-sys/Cargo.toml#13).

I think I got them all now, some of the leftovers require installing features, that aren't needed in other locations so I left them relatively unchanged.

PR #475, opened 2024-03-11
feat: Tokio with D1 example

franz101:
Simple example for using d1, based on the docs but working end2end


sam-rusty:
in your example, you didn't use `tokio-postres` at all. what's the point then?

kflansburg:
I think D1 is fairly well documented, and I'm not sure that this example adds anything. At the very least it doesn't seem related to `tokio`.

PR #480, opened 2024-03-15
[WIP] Convert global fetcher to use `http` crate

kflansburg:
This was identified as an inconsistency in the original PR (#477). This is one option for fixing it (converting the existing API), but another that I will try is to improve interoperability so that we can just recommend that people use `reqwest` directly and remove this API entirely. Will open another PR with that option. 

Note that there is one breaking change: the request object supplied to `fetch_with_request` is no longer a reference.

### DIFF
--- a/worker-sandbox/src/lib.rs
+++ b/worker-sandbox/src/lib.rs
@@ -3,6 +3,8 @@ use futures_util::{future::Either, StreamExt, TryStreamExt};
 use rand::Rng;
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "http")]
+use std::convert::TryFrom;
+#[cfg(feature = "http")]
 use std::convert::TryInto;
 use std::{
     sync::{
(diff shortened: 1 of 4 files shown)


worker-sandbox/src/lib.rs
```diff
@@ -21,6 +23,34 @@ mod r2;
 mod test;
 mod utils;
 
+#[cfg(feature = "http")]
```
kflansburg, line 26: These just convert the types between what fetch expects (depends on feature flag) and what the test harness router expects (`worker::Request`). I think pretty soon we should have the `http` version of the test harness use `axum` and remove this.

spigaz: If the only reason you see for that API is to be used by the tests, then they should be dropped.

Regarding tests I'm bit more opinionated, as I believe we should move to wasm-bindgen-test and move unit tests closer to the source.

As MrBBot no longer told me to wait, so I started getting my feet wet...
https://github.com/rustwasm/wasm-bindgen/issues/3891

So most of the tests would be direct tests instead of the current model.

PR #483, opened 2024-03-19
impl Send/Sync for Queue

avsaase:
I'm in the process of migrating a worker to use `axum` and this missing impl prevents using `Queue` in a state that can be attached to a router. Another issue is a missing `Send`/`Sync` impl on `KvStore` but that lives in the `worker-kv` repo, outside of the cloudflare org. Should I make a PR there as well?

### DIFF
--- a/worker/src/queue.rs
+++ b/worker/src/queue.rs
@@ -145,6 +145,9 @@ impl<'a, T> std::iter::ExactSizeIterator for MessageIter<'a, T> where T: Deseria
 
 pub struct Queue(EdgeQueue);
 
+unsafe impl Send for Queue {}
+unsafe impl Sync for Queue {}
+
 impl EnvBinding for Queue {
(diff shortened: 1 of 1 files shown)


kflansburg:
Yep, `workers-kv` was another big hurdle I ran into. A PR there would be appreciated.

kflansburg:
> Yep, `workers-kv` was another big hurdle I ran into. A PR there would be appreciated.

@avsaase, I spoked with @zebp and we should probably just move `workers-kv` to be a crate in the `workers-rs` Cargo workspace.

avsaase:
> @avsaase, I spoked with @zebp and we should probably just move `workers-kv` to be a crate in the `workers-rs` Cargo workspace.

Yes that makes sense. I was surprised it wasn't part of the workspace already. 

I already opened the PR: https://github.com/zebp/worker-kv/pull/14. It can either be merged before moving the code of it can be closed.

PR #484, opened 2024-03-19
implement get_tags for hibernatable websockets

eric-seppanen:
get_tags allows a durable object to read metadata describing the websocket.

It's useful when a durable object is managing multiple websockets and needs to discover which one sent a message that caused a hibernating websocket to wake.


eric-seppanen:
This is a step towards supporting all the hibernatable websockets APIs. See #436 for some previous discussion.

kflansburg (approved):

Issue #486, opened 2024-03-20
[Feature] Enable DurableObject Storage's `transaction` API


TannerRogalsky:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

The function that creates a transaction for a DurableObject's Storage is currently commented out. See here: https://github.com/cloudflare/workers-rs/blob/3db61a422ec970ccde0e109bdd18aac61d32db70/worker/src/durable.rs#L445-L464

The comment about nightly features is no longer relevant and I have a local changeset that enables this functionality with the existing dependency versions.

The documentation on this API indicates that it shouldn't generally be necessary to invoke this but there are still some cases where it is useful: https://developers.cloudflare.com/durable-objects/api/transactional-storage-api/#supported-options-2

PR #487, opened 2024-03-20
Expose DurableObject Storage's transaction function

TannerRogalsky:
Associated issue: #486 

The comment about nightly features is no longer relevant and I have a local changeset that enables this functionality with the existing dependency versions.

The documentation on this API indicates that it shouldn't generally be necessary to invoke this but there are still some cases where it is useful: https://developers.cloudflare.com/durable-objects/api/transactional-storage-api/#supported-options-2


kflansburg (approved):
I think probably we will want to follow up with some tests here.

Issue #492, opened 2024-03-24
Make workers-kv a crate in this workspace


kflansburg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Copy code from `worker-kv` and place it in this Workspace, copy any tests and examples. 

https://github.com/zebp/worker-kv

Also make `KvStore` to be `Send` for `http` feature.

Kakapio:
Please close issue.

PR #494, opened 2024-03-25
Add worker-kv crate to repository.

Kakapio:
WIP of #492.
@zebp I would appreciate a rough roadmap of this if you don't mind fleshing out the specifics a bit more. My understanding is the only task at this time is to remove whatever custom HTTP types you've been using and replace them with `crates/http` library calls.


examples/kv/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+js-sys = "0.3.50"
+serde_json = "1.0.64"
+wasm-bindgen = "0.2.73"
+wasm-bindgen-futures = "0.4.23"
+worker-kv = { path = "../" }
```
kflansburg, line 15: This path needs to be updated

worker-kv/tests/integration.rs
```diff
@@ -0,0 +1,103 @@
+use std::{
```
kflansburg, line 1: If you have bandwidth, could you try getting these tests working with Actions in this repo? It can be a separate workflow for now. 

I'm not sure I understand where / how the `CF_API_TOKEN` is used, but I think we should be able to remove dependencies on a real account and just use Miniflare.

kflansburg (changes requested):

worker-kv/tests/integration.rs
Kakapio: All tests in examples/ or just the worker-kv one? The included worker-kv test has some sort of issue, something to do with bad references to `nix`. I am working on figuring out what exactly it is. [Here is the log](https://pastebin.com/XBELw6Dv).

kflansburg: Not sure what you mean, I think the `workers-kv` test should run in Actions alongside the other tests. 

Unfortunately I have not seen that issue before.

Kakapio: Ah, it turns out that building on Windows has some issues. I resolved it on MacOS. What sort of environment does the Cloudflare team use? WSL? MacOS?

Kakapio: > Not sure what you mean, I think the `workers-kv` test should run in Actions alongside the other tests.

Do any of the other tests already included in Actions use Miniflare? Because if not, I will likely have to make a separate Workflow. Workers-kv requires miniflare which requires an `npm install` to grab the proper dependencies.

kflansburg: I use MacOS, I'm not sure what others use.

Kakapio:
I believe I have resolved the formatting issue if you can give the Action another whirl @kflansburg

kflansburg:
Is the plan to do the tests in a follow up PR?

Kakapio:
> Is the plan to do the tests in a follow up PR?

Yes. It is my first time working on Actions so I want to make sure I do not pollute the branch.

kflansburg (approved):
Ok, sounds good

Kakapio:
> Ok, sounds good

Final change, cleaned up the formatting issue. Should be mergable.

kflansburg (approved):

Issue #441, opened 2024-02-07
[BUG] TCP Connect does not error if connection not able to be created


cmmarslender:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What version of `workers-rs` are you using?

main

### What version of `wrangler` are you using?

3.26.0

### Describe the bug

The tldr of this issue is a request to make connect() fail if the connection could not be established due to a closed port, or some other reason. See below for details

It seems that socket `connect()` doesn't actually return an error if a connection is unable to be created for some reason (port closed, etc). Trying to build a worker that as quickly and simply as possible can let a user know if a port on their IP is open or not.

Generally, I'd expect an error from the connect() call if the port is closed, but in the workers env, this doesn't seem to hold true. Here's the code that I would expect to fail on a closed port that just continues past a closed port:
```
let mut socket = Socket::builder()
        .secure_transport(SecureTransport::Off)
        .connect(&connecting_ip, port)
        .map_err(|_| worker::Error::from("Could not create socket"))?;
```

For the most part, I can reliably tell if the port is open or closed by attempting to write some data AND then read a byte from the socket after the connect call (write isn't enough, and also doesn't seem to fail if the port is closed - have to add the read step). This isn't particularly ideal though, since I have to rely on binding `setTimeout` into rust and setting a timeout on the read to determine if its not going to succeed. This method also doesn't work for all protocols that I need to check for open ports.

I was curious to see how this worked in a non-workers rust environment, and was able to reliably determine if a port is open at the connect call, using the following code:

```
use std::net::{SocketAddr, TcpStream};
use std::str::FromStr;
use std::time::Duration;

fn main() {
    let remote_ip = "1.2.3.4";
    let port = 123;
    let timeout = Duration::from_secs(1);

    // Construct the remote address from the IP and port
    let remote_addr = match SocketAddr::from_str(&format!("{}:{}", remote_ip, port)) {
        Ok(addr) => addr,
        Err(e) => {
            eprintln!("Error constructing remote address: {}", e);
            return;
        }
    };

    // Attempt to connect with a timeout
    match TcpStream::connect_timeout(&remote_addr, timeout) {
        Ok(_) => println!("Port {} is open on {}", port, remote_ip),
        Err(e) => println!("Failed to connect to port {} on {}: {}", port, remote_ip, e),
    }
}
```

kflansburg:
[...]

PR #509, opened 2024-03-28
Add Socket.opened implementation

Kakapio:
#441 
I'm a bit confused regarding the actual implementation of Socket.opened itself. I created all the necessary bindings, same as the ones [already implemented](https://developers.cloudflare.com/workers/runtime-apis/tcp-sockets/#Socket), but I'm not sure if I'm done. The `Socket` struct confuses me. I know the `inner` is the bound JS object itself, and the rest are mostly streams and options. What do Options such as `close` do? They are used in Tokio traits such as `AsyncWrite`. It looks like to me that particular one returns whether the Socket has closed yet, and the implementation also indicates that.  I tried to decipher what I think is the [original implementation](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/sockets.h) of `opened` but have not been successful in understanding it. I added an `opened` Option to the `Socket` struct but am not sure of what to do with it.

Would it be incorrect to just implement it in `AsyncRead.poll_read()` ?

I _think_ I just have to use reflection to grab the `opened` value on `poll_read()`?

### DIFF
--- a/worker/src/socket.rs
+++ b/worker/src/socket.rs
@@ -20,38 +20,33 @@ use web_sys::{
     ReadableStream, ReadableStreamDefaultReader, WritableStream, WritableStreamDefaultWriter,
 };
 
+#[derive(Default)]
 enum Reading {
+    #[default]
     None,
     Pending(JsFuture, ReadableStreamDefaultReader),
     Ready(Vec<u8>),
 }
 
-impl Default for Reading {
-    fn default() -> Self {
-        Self::None
-    }
-}
-
+#[derive(Default)]
 enum Writing {
     Pending(JsFuture, WritableStreamDefaultWriter, usize),
+    #[default]
     None,
 }
 
-impl Default for Writing {
-    fn default() -> Self {
-        Self::None
-    }
-}
-
+#[derive(Default)]
 enum Closing {
     Pending(JsFuture),
+    #[default]
     None,
 }
 
-impl Default for Closing {
-    fn default() -> Self {
-        Self::None
-    }
+#[derive(Default)]
+enum Opening {
+    Pending(JsFuture),
+    #[default]
+    None,
 }
 
 /// Represents an outbound TCP connection from your Worker.
@@ -62,6 +57,7 @@ pub struct Socket {
     write: Option<Writing>,
     read: Option<Reading>,
     close: Option<Closing>,
+    opened: Option<Opening>,
 }
 
 // This can only be done because workers are single threaded.
@@ -79,6 +75,7 @@ impl Socket {
             read: None,
             write: None,
             close: None,
+            opened: None,
         }
     }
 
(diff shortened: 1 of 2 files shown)


kflansburg:
> #441 I'm a bit confused regarding the actual implementation of Socket.opened itself. I created all the necessary bindings, same as the ones [already implemented](https://developers.cloudflare.com/workers/runtime-apis/tcp-sockets/#Socket), but I'm not sure if I'm done. The `Socket` struct confuses me. I know the `inner` is the bound JS object itself, and the rest are mostly streams and options. What do Options such as `close` do? They are used in Tokio traits such as `AsyncWrite`. It looks like to me that particular one returns whether the Socket has closed yet, and the implementation also indicates that. I tried to decipher what I think is the [original implementation](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/sockets.h) of `opened` but have not been successful in understanding it. I added an `opened` Option to the `Socket` struct but am not sure of what to do with it.
> 
> Would it be incorrect to just implement it in `AsyncRead.poll_read()` ?
> 
> I _think_ I just have to use reflection to grab the `opened` value on `poll_read()`?

Have you seen https://developers.cloudflare.com/workers/runtime-apis/tcp-sockets/#socket?
This was the primary thing that I consulted when I wrote the original Rust wrappers.

worker/src/socket.rs
```diff
@@ -79,6 +75,7 @@ impl Socket {
             read: None,
             write: None,
             close: None,
+            opened: None,
```
kflansburg, line 78: I'm not sure that this is necessary, the read / write / close properties I think are used for locking / unlocking the internal reader of the socket. Whereas this never gets set / accessed I think?

kflansburg (commented):
Looks reasonable to me, we probably need to think about adding tests for this API.

worker/src/socket.rs
Kakapio: What tests would you like me to write? A general list is okay.

kflansburg: Unfortunately I'm not sure it will be straightforward to mock sockets. We didn't do any tests when we first added this API. 

But testing for "not opened" might be the easiest. Just open a socket to somewhere that shouldn't be valid (maybe some random localhost port) and verify that opened promise resolves and indicates failure?

Kakapio: I attempted a test like so:

```
#[test]
    fn test_is_opened() {
        println!("hi im running the special test :)");
        let socket = Socket::builder()
        .secure_transport(SecureTransport::StartTls)
        .connect("database_url", 5432);

        assert!(true);
    }
```

and saw that this test has the same error as one of the two failing tests, `cannot call wasm-bindgen imported functions on non-wasm targets`. Zeb mentioned you guys are going to work on adding Vitest for Workers to workers-rs, so I'm assuming getting these working isn't a high priority?

I made an attempt and I am honestly not sure about being able to get these working without a lot of hassle.

kflansburg: We already use vitest / miniflare. Check out the `worker-sandbox` directory. 

If you run `npm install` and `npm run test` then it should run the test suite.

Kakapio: Gotcha. The Miniflare API doesn't provide access to TCP sockets as far as I know. I think you were on the money with sockets being extremely difficult to test, I don't see a way to accomplish this without a lot of API changes. I'd be happy to take your suggestion if you have any ideas.

kflansburg (approved):

worker/src/socket.rs
kflansburg: I'll propose something in a follow up PR

Issue #490, opened 2024-03-23
[Feature] Contribution Guidance


Kakapio:
Hello!

I am interested in contributing to workers-rs but wanted to get a run-down from someone on the team as to what features require development. I was hoping for someone more experienced to point me to a good first issue, or a part of the repository that needs more testing. I understand that a lot of the development team here is at Cloudflare, so I am missing a ton of insider knowledge. Sorry to make an issue for this, but I didn't know a better way to reach out.

Thank you!
Kakapio 😄

kflansburg:
Hey @Kakapio, thanks for you interest. I have labeled a few issues as `good first issue`. Most of these relate to documentation or adding individual JavaScript methods to existing bindings. 

The repository is also undergoing a larger project to move from using custom HTTP types to using Rust's `http` crate. I think there may be some opportunities for parallel work-streams if you are interested.

Kakapio:
> The repository is also undergoing a larger project to move from using custom HTTP types to using Rust's `http` crate. I think there may be some opportunities for parallel work-streams if you are interested.

Would love to hear more.

kflansburg:
This is just sort of a general effort, not documented in a ticket, but primarily various types need to be converted to use `http` types where possible (behind crate flag `http`). They also need to be made `Send` because `axum` expects this. Here is a potential good ticket for this https://github.com/cloudflare/workers-rs/issues/492

Kakapio:
> This is just sort of a general effort, not documented in a ticket, but primarily various types need to be converted to use `http` types where possible (behind crate flag `http`). They also need to be made `Send` because `axum` expects this. Here is a potential good ticket for this #492

Is there anywhere in particular you would like me to start looking? Replacing the custom types with http sounds fun. I'm happy to give it a go.

kflansburg:
So the next thing I was thinking of looking into was what our guidance will be for making HTTP requests. With the old types, we have the global `fetch` API, but there are some rough edges and I think in the new model people may prefer to just use something like `reqwest` or `hyper`. 

So I was thinking the next step would be to remove the Fetcher in `global.rs` when the `http` feature is enabled, and then see how ergonomic it is to use one of those libraries in the test suite. We may end up maintaining an API similar to `fetch` that uses `reqwest` under the hood. 

One rub that I think I saw is that `reqwest` does not appear to use `http` 1.0 yet? Not sure what the status / roadmap is for this.

avsaase:
[...]

PR #510, opened 2024-03-28
Use Reqwest instead of custom HTTP handler

Kakapio:
As discussed in #490 

I intend to recreate the Fetch API using Reqwest to add support for the HTTP crate.

### DIFF
--- a/worker/src/global.rs
+++ b/worker/src/global.rs
@@ -1,23 +1,25 @@
 use std::ops::Deref;
+use std::time::Duration;
 
+use reqwest::{Body, Client, Error, Response};
+use tokio::time;
 use wasm_bindgen::JsCast;
 use wasm_bindgen_futures::JsFuture;
 
 use crate::{
-    request::Request as WorkerRequest, response::Response as WorkerResponse, AbortSignal, Result,
+     AbortSignal, Result,
 };
 
 /// Construct a Fetch call from a URL string or a Request object. Call its `send` method to execute
 /// the request.
 pub enum Fetch {
     Url(url::Url),
-    Request(WorkerRequest),
+    Request(Body),
 }
 
-#[cfg(not(feature = "http"))]
 impl Fetch {
     /// Execute a Fetch call and receive a Response.
-    pub async fn send(&self) -> Result<WorkerResponse> {
+    pub async fn send(&self) -> Result<Response> {
         match self {
             Fetch::Url(url) => fetch_with_str(url.as_ref(), None).await,
             Fetch::Request(req) => fetch_with_request(req, None).await,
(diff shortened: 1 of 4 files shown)


worker/src/global.rs
```diff
@@ -1,37 +1,75 @@
...
+    pub async fn send_with_signal(&self, signal: &AbortSignal) -> Result<Response> {
         match self {
             Fetch::Url(url) => fetch_with_str(url.as_ref(), Some(signal)).await,
             Fetch::Request(req) => fetch_with_request(req, Some(signal)).await,
         }
     }
 }
 
+//#[cfg(feature = "http")]
+async fn fetch_with_str(url: &str, signal: Option<&AbortSignal>) -> Result<Response> {
```
kflansburg, line 39: Might have to upstream some changes to pass in the abort signal https://github.com/seanmonstar/reqwest/blob/68a3f5803b03393938c39a3794a5a9e5f5264796/src/wasm/client.rs#L221

worker/src/global.rs
```diff
@@ -1,37 +1,75 @@
...
+    pub async fn send(&self) -> Result<Response> {
         match self {
             Fetch::Url(url) => fetch_with_str(url.as_ref(), None).await,
             Fetch::Request(req) => fetch_with_request(req, None).await,
         }
     }
 
     /// Execute a Fetch call and receive a Response.
-    pub async fn send_with_signal(&self, signal: &AbortSignal) -> Result<WorkerResponse> {
+    pub async fn send_with_signal(&self, signal: &AbortSignal) -> Result<Response> {
```
kflansburg, line 30: I wonder if we want to convert `reqwest::Response` to `http::Response<Body>`. I think its not totally straightforward, but possible.

kflansburg (commented):
Let's make sure to keep API with `http` disabled unchanged.

worker/src/global.rs
Kakapio: This might be a dumb question, but why does the `reqwest `repo contain `.signal` functionality and the Cargo package doesn't? I don't see it in the docs either when it's clearly in the repository.
https://docs.rs/reqwest/latest/reqwest/struct.Client.html

worker/src/global.rs
Kakapio: What would be the purpose of that? Would it help with keeping the data format more flexible? I'm curious.

kflansburg: Well, right now if they wanted to call `fetch` and then return the value, it would be the wrong type unless it was converted to `http::Response`.

worker/src/global.rs
Kakapio: Would it be acceptable to just use `Drop` and provide a handler? I am not sure if we necessarily need direct access to the signal.

Something like:
```rs
// Asynchronous loop to continuously check for cancellation signal
    loop {
        tokio::select! {
            _ = sleep(Duration::from_secs(1)) => {
                println!("Async task running...");
            }
            _ = cancel_receiver.recv() => {
                println!("Cancellation signal received. Dropping task...");
                break; // Break out of the loop when cancellation signal is received
            }
        }
    }```

kflansburg: I'm not sure what the tradeoffs are here. Using a JavaScript Abort Signal lets the runtime cancel the request rather which I think might be more efficient, but if there is no way to plumb this through it may be ok to go this route.

PR #512, opened 2024-03-29
Add semver checks and sanity-check build to create-release-pr

kflansburg:
Adds [cargo-semver-checks](https://crates.io/crates/cargo-semver-checks) when creating a new release. Since we are still on `0.x.y`, these checks do not detect anything by default. I have forced the checker to treat every release as a `minor` version so that it will highlight any breaking changes. 

Since we still want to be able to ship breaking changes occasionally, I have set these steps to not fail the build, and instead will review them manually. 

I've also added a sanity check build step to make sure there are no issues introduced by `cargo update`. Ideally the full test suite would still run on the PR, but I will have to figure out how to do that later.

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -29,6 +29,25 @@ jobs:
           cargo update
           NEW_VERSION=$(cargo metadata --no-deps --format-version 1 | jq -rc '.packages[] | select(.name=="worker") | .version')
           echo "version=$NEW_VERSION" >> "$GITHUB_OUTPUT"
+      - name: Check semver
+        uses: obi1kenobi/cargo-semver-checks-action@v2
+        continue-on-error: true
+        with:
+          package: worker
+          verbose: true
+          release-type: minor
+      - name: Check semver (http)
+        uses: obi1kenobi/cargo-semver-checks-action@v2
+        continue-on-error: true
+        with:
+          package: worker
+          feature-group: all-features
+          verbose: true
+          release-type: minor
+      - name: Build workspace
+        run: cargo build --workspace
+      - name: Build workspace (http)
+        run: cargo build --package worker-sandbox --all-features
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v6
         with:


kflansburg:
cc @avsaase

obi1kenobi:
Hey! I'm the `cargo-semver-checks` maintainer — thanks for trying it out and please let me know if you run into any issues!

I noticed a couple of minor things in the PR that I wanted to mention.

> Since we are still on 0.x.y, these checks do not detect anything by default.

This is not completely accurate. Both cargo and cargo-semver-checks implement semver with an "ignore leading zeroes" approach, so `0.1.0 -> 0.2.0` is considered major and `0.1.0 -> 0.1.1` is considered a minor bump.

This guide uses the terms “major” and “minor” assuming this relates to a “1.0.0” release or later. Initial development releases starting with “0.y.z” can treat changes in “y” as a major release, and “z” as a minor release. “0.0.z” releases are always major changes. This is because Cargo uses the convention that only changes in the left-most non-zero component are considered incompatible.
*from https://doc.rust-lang.org/cargo/reference/semver.html*

The other thing I noticed is that the two cargo-semver-checks runs differ in whether they use `feature-group: all-features` or not. This might not do what you intended — the two runs seem to me like they'd end up being identical.

This is because `cargo-semver-checks` [uses a heuristic to enable all non-internal, non-nightly features](https://github.com/obi1kenobi/cargo-semver-checks?tab=readme-ov-file#what-features-does-cargo-semver-checks-enable-in-the-tested-crates) as part of semver-checking. The crate being checked has no features that would have been excluded by the heuristic (and which would have been re-added by the `all-features` group), so all features are used both times.

I'm guessing you might have meant to make one of them be `feature-group: default-features` instead.

I'd love to know if my guess here is accurate. If so, it's probably a good idea for me to add a warning to `cargo-semver-checks` about the ineffective `all-features` flag selection, since the confusion is very understandable and likely to happen again. I like to keep a "no sharp edges unless absolutely necessary" policy 😅

kflansburg:
I see, we use `0.0.z` though (maybe we shouldn't) so when I ran this locally it did ignore the changes I was trying to detect. 

Regarding the features, I think you may be right. What we want is to test with and without the `http` feature enabled, so `default-features` is probably the way to go. 

Thanks for taking a look!

obi1kenobi:
Ah yes, in `0.0.z` versions, everything is a major change since leading zeroes are ignored. Using an explicit `minor` release type makes sense :+1:

Please open issues or just ping me directly if you ever run into any issues, or if you have ideas for how to make `cargo-semver-checks` even better!

PR #518, opened 2024-03-30
Add socket tests

kflansburg:
Add tests for:
* Socket fails to connect and `opened` promise returns error
* Socket connects, echos data, and closes. 

cc @Kakapio


Kakapio:
Oh, so it was possible! There's a lot going on here, I'll do my best to decipher it...

PR #520, opened 2024-04-01
Make chrono-tz an optional dependency

### LINKED ISSUE #519: [Feature] Make `chrono-tz` an optional dependency

### Description

I was looking though a compiled wasm binary using `wasm-objdump` and I noticed there is a lot of data in the data section related to time zones. In our own code we use `chrono-tz ` with the `filter-by-regex` feature to select which time zones are compiled (see [here](https://github.com/chronotope/chrono-tz?tab=readme-ov-file#limiting-the-timezone-table-to-zones-of-interest)) but this feature is not enabled in the `worker` crate. I suspect this causes the filter to not work because in the data section of my binary I see time zones that should have been filtered out.

The only place where `chrono-tz` is used is here:
https://github.com/cloudflare/workers-rs/blob/79ce476af039229f30989594b1167e8fbe058c74/worker/src/cf.rs#L153-L157

This makes me think there's an easy opportunity for some binary size saving by either 
1. adding a (potentially default) feature that enables the `chrono-tz` dependency and guards this function
2. enabling the `filter-by-regex` feature and change the return type of this function to an `Option` (maybe that's a good idea anyway?). 
3. getting rid of the function altogether is also an option.

I suppose a combination of 1 and 2 is also possible but probably overkill. 

What do you think? I'm happy to make a PR.

kflansburg:
Closes #519

### DIFF
--- a/worker/src/cf.rs
+++ b/worker/src/cf.rs
@@ -150,7 +150,8 @@ impl Cf {
         self.inner.region_code()
     }
 
-    /// Timezone of the incoming request
+    /// **Requires** `timezone` feature. Timezone of the incoming request
+    #[cfg(feature = "timezone")]
     pub fn timezone(&self) -> impl chrono::TimeZone {
         let tz = self.inner.timezone();
         tz.parse::<chrono_tz::Tz>().unwrap()
--- a/worker/src/cf.rs
+++ b/worker/src/cf.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "timezone")]
+use crate::Result;
+
 /// In addition to the methods on the `Request` struct, the `Cf` struct on an inbound Request contains information about the request provided by Cloudflare’s edge.
 ///
 /// [Details](https://developers.cloudflare.com/workers/runtime-apis/request#incomingrequestcfproperties)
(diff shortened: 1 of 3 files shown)


worker/src/cf.rs
```diff
@@ -150,7 +150,8 @@ impl Cf {
         self.inner.region_code()
     }
 
-    /// Timezone of the incoming request
+    /// **Requires** `timezone` feature. Timezone of the incoming request
+    #[cfg(feature = "timezone")]
     pub fn timezone(&self) -> impl chrono::TimeZone {
```
avsaase, line 155: I think the return type here should be changed to `Option<impl chrono::TimeZone`. If the user directly depends on `chrono-tz` with the `filter-by-regex` feature then parsing will fail if the time zone has been filtered out.

avsaase: I wonder if this function should be removed altogether. It returns an opaque type that's not usable in any way without bringing in another dependency. At that point you might as well leave the time zone parsing to the user.

kflansburg: Does this change enable you to actually use the filter, or should I also enable that filter in `worker`?

avsaase: The filter works as long as somewhere in the dependency graph the `filter-by-regex` feature is enabled and the environment variable is passed to cargo. This is outside the control of the `worker` create so it should not assume all time zones are available.

kflansburg: Ok, I've changed this method to return a `Result`

avsaase (approved):

PR #523, opened 2024-04-01
Fix issue with data corruption in web_sys::WebSocket

### LINKED ISSUE #379: [BUG] WebSocket sends the wrong data

0.0.18

3.6.0

### Describe the bug

`WebSocket::send_with_bytes` reproducibly corrupts outgoing binary data in some situations. I have included a repro where `[00 01 02]` becomes `[65 00 01]`.

Repro repo: https://github.com/paulgb/cloudflare-repro/tree/main

I been able to reproduce this on Mac and Linux.

### Steps To Reproduce

1. Clone `https://github.com/paulgb/cloudflare-repro.git`
2. `npm i && npm run dev`
3. open another terminal and `cd` into `client`
4. `npm i && npx ts-node ./client.ts`

Observe that the output is `received: <Buffer 65 00 01>` rather than `received: <Buffer 01 02 03>`.

kflansburg:
Closes #379

### DIFF
--- a/worker/src/websocket.rs
+++ b/worker/src/websocket.rs
@@ -1,6 +1,7 @@
 use crate::{Error, Method, Request, Result};
 use futures_channel::mpsc::UnboundedReceiver;
 use futures_util::Stream;
+use js_sys::Uint8Array;
 use serde::Serialize;
 use url::Url;
 use worker_sys::ext::WebSocketExt;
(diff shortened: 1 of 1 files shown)


eric-seppanen:
I have been running for a few months with an identical patch, so I can confirm this works to suppress the problem.

Does this imply that an extra copy of the data is performed? It's not obvious to me what the conversion to `Uint8Array` implies a copy that might have performance impact.

I added some other concerns [here](https://github.com/cloudflare/workers-rs/issues/379#issuecomment-2030253935).

kflansburg:
I've added a comment for posterity. If we find similar bugs, we will address them.

PR #524, opened 2024-04-01
Return 500 instead of panicking when returning error to fetch handler

avsaase:
This PR returns a generic 500 response to the client, instead of panicking, when an error is returned to the fetch handler. IMO panicking defeats the purpose of Rust's error handling model and propagating errors.

The `respond_with_errors` option avoids a panic but blindly returns the error to the client which is not good practice.

I can also add a new option if you want to keep the existing behavior. Or add a new panic option in case some people really want to do a panic.

### DIFF
--- a/worker-macros/src/event.rs
+++ b/worker-macros/src/event.rs
@@ -57,7 +57,7 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream, http: bool) -> TokenSt
                     }
                 }
                 false => {
-                    quote! { panic!("{}", e) }
+                    quote! { ::worker::Response::error("INTERNAL SERVER ERROR", 500).unwrap().into() }
                 }
             };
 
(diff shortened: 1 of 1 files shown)


worker-macros/src/event.rs
```diff
@@ -50,15 +50,10 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream, http: bool) -> TokenSt
...
-                        ::worker::Response::error(e.to_string(), 500).unwrap().into()
-                    }
-                }
-                false => {
-                    quote! { panic!("{}", e) }
-                }
+            let error_handling = if respond_with_errors {
+                quote! { ::worker::Response::error(e.to_string(), 500).unwrap().into() }
+            } else {
+                quote! { ::worker::Response::error("INTERNAL SERVER ERROR", 500).unwrap().into() }
```
kflansburg, line 56: Should this `console_error`? I think it makes sense to not return the error to the client, but I'm wondering if the it is always desirable to log it somewhere.

kflansburg (approved):

avsaase:
Technically this is a breaking change but I don't think with cargo-semver-checks will pick up on it.

PR #527, opened 2024-04-02
Allow returning any error that implements `std::error::Error`

kflansburg:
This allows the `fetch` handler to return any error type that implements `std::error::Error`. This will allow the user to choose how they want to handle errors (use `worker::Error`, `anyhow`, `thiserror`, etc.). 

This also changes some behavior slightly:
1. If `respond_with_error` is enabled, this now uses `Display` to stringify the error instead of `to_string`. 
2. The error logged is now stringified with `Debug` and with `console_error` instead of `console_log`. 

cc @avsaase

### DIFF
--- a/worker-macros/src/event.rs
+++ b/worker-macros/src/event.rs
@@ -65,12 +65,28 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
                     ctx: ::worker::worker_sys::Context
                 ) -> ::worker::worker_sys::web_sys::Response {
                     let ctx = worker::Context::new(ctx);
-                    let result = #input_fn_ident(::worker::FromRequest::from_raw(req), env, ctx).await.map(::worker::IntoResponse::into_raw);
-                    // get the worker::Result<worker::Response> by calling the original fn
-                    match result {
-                        Ok(res) => res,
+                    match ::worker::FromRequest::from_raw(req) {
+                        Ok(req) => {
+                            let result = #input_fn_ident(req, env, ctx).await;
+                            // get the worker::Result<worker::Response> by calling the original fn
+                            match result {
+                                Ok(raw_res) => {
+                                    match ::worker::IntoResponse::into_raw(raw_res) {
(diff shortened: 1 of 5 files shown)


worker-macros/src/event.rs
```diff
@@ -51,7 +51,7 @@ pub fn expand_macro(attr: TokenStream, item: TokenStream, http: bool) -> TokenSt
             input_fn.sig.ident = input_fn_ident.clone();
 
             let error_handling = if respond_with_errors {
-                quote! { ::worker::Response::error(e.to_string(), 500).unwrap().into() }
+                quote! { ::worker::Response::error(format!("{}", e), 500).unwrap().into() }
```
avsaase, line 54: Just FYI, this is equivalent to the previous code. `ToString` is implemented for all types that implement `Display`: https://doc.rust-lang.org/std/string/trait.ToString.html

avsaase (commented):
Should `worker::Error` also implement the `Error` trait?

dakom:
Just wondering - would it make sense to get rid of `respond_with_errors`? 

I guess that question is more related to the change in https://github.com/cloudflare/workers-rs/pull/524, but at the end of the day - it's now pretty trivial for application code redact all errors before returning from the handler (and it's a bit unexpected for that redaction to be the default)

Even if y'all agree, not sure what the policy is about breaking backwards compatibility is though, consider it a curiosity question rather than a hard push :)

kflansburg:
> Just wondering - would it make sense to get rid of `respond_with_errors`?
> 
> I guess that question is more related to the change in #524, but at the end of the day - it's now pretty trivial for application code to redact all errors before returning from the handler (and it's a bit unexpected for that redaction to be the default)
> 
> Even if y'all agree, not sure what the policy is about breaking backwards compatibility though, consider it a curiosity question rather than a hard push :)

I think its still useful for people to control whether and Error is returned to client. Its true that they may control this now via axum and IntoResponse for their error type, but I dont think that will apply to everyone.

Al3676:
Oké

PR #528, opened 2024-04-03
Add KV Testing

Kakapio:
I had a lot of difficulty getting the tests working in this repo. I had issues with not including the tests in workers-kv as a member of the main `Cargo.toml`. There were also issues with the versioning of the various nested `Cargo.toml` files. I think it would probably be prudent to separate the `workers-kv/tests` into its own member. I have not yet seen rust-based tests anywhere else in this repo so I wanted to get feedback on that.

I will make a follow-up pull request for updated dependencies and separation of testing modules. I will also look into updating the tests to use Miniflare 3 instead of Miniflare 1.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -82,3 +82,7 @@ jobs:
       - name: Run tests (http)
         shell: bash
         run: npm run test-http
+
+      - name: Run worker-kv tests
+        shell: bash
+        run: cd worker-kv/tests && npm install && cargo test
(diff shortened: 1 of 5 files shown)


kflansburg:
There are only a handful of endpoints here, and this very closely matches the pattern used in the main test suite, would it be simpler to just move these test cases over to `worker-sandbox`?

guybedford:
All tests seem to run in the sandbox folder now, improvements to testing infrastructure are always welcome though!

PR #529, opened 2024-04-03
Rust-friendly DigestStream API

### LINKED ISSUE #383: [Feature] add support for crypto.DigestStream

### Description

According to the docs, workers have access to a non-standard Crypto type called "DigestStream". It would be nice if this crate provided ergonomic access to this class as well as a nice way to compute the digest while doing something else with a request stream.

dakom:
Added an example showing that it works too :)

* Fixes #383


worker/src/digest_stream.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum DigestStreamAlgorithm {
+    Sha1,
+    Sha256,
+    Sha384,
+    Sha512,
+}
+
+#[wasm_bindgen]
+extern "C" {
```
kflansburg, line 29: This would typically go in `worker-sys`

worker/src/lib.rs
```diff
@@ -162,6 +162,7 @@ mod cors;
 pub mod d1;
 mod date;
 mod delay;
+pub mod digest_stream;
```
kflansburg, line 165: Typically, we would re-export this. I'm not sure that I like this pattern, so we can keep it like this, but I think the module name should be `crypto`.

worker/src/digest_stream.rs
```diff
@@ -0,0 +1,66 @@
...
+use web_sys::WritableStream;
+
+impl DigestStream {
+    #[allow(dead_code)]
+    pub fn new(algo: DigestStreamAlgorithm) -> Self {
+        Self::internal_new(algo.as_str())
+    }
+
+    #[allow(dead_code)]
+    pub async fn digest(&self) -> Result<Uint8Array, JsValue> {
```
kflansburg, line 13: I think this should return `Result<Vec<u8>, worker::Error>`. If you need access to the JavaScript value, then maybe add `digest_raw`?

worker/src/digest_stream.rs
dakom: ah, I think I see the pattern now... okay, will fix it up

worker/src/digest_stream.rs
dakom: d'oh, yup, and this is supported automatically with `?` :)

worker/src/digest_stream.rs
dakom: done in https://github.com/cloudflare/workers-rs/commit/2bfa37e50425dc92d2af2ced66a1ecf7f2eab944

worker/src/crypto.rs
```diff
@@ -0,0 +1,65 @@
...
+/// let _ = body.pipe_to(digest_stream.raw());
+///
+/// let bytes:Vec<u8> = digest_stream.digest().await.unwrap().to_vec();
+/// ```
+pub struct DigestStream {
+    inner: worker_sys::DigestStream,
+}
+
+impl DigestStream {
+    #[allow(dead_code)]
```
kflansburg, line 29: Are these actually dead? Since you publicly export this module, it should be fine.

worker/src/crypto.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub fn new(algo: DigestStreamAlgorithm) -> Self {
+        Self {
+            inner: worker_sys::DigestStream::new(algo.as_str()),
+        }
+    }
+
+    #[allow(dead_code)]
+    pub async fn digest(&self) -> Result<Uint8Array, crate::Error> {
+        let buffer: ArrayBuffer = JsFuture::from(self.inner.digest()).await?.unchecked_into();
```
kflansburg, line 38: You may need to do this to avoid `Sync` / `Send` issues. 
```suggestion
        let fut = SendFuture(JsFuture::from(self.inner.digest()));
        let buffer: ArrayBuffer = fut.await?.unchecked_into();
```

examples/digest/yarn.lock
```diff
@@ -0,0 +1,870 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kflansburg, line 1: Can this file be removed?

kflansburg (changes requested):

examples/digest/yarn.lock
dakom: done in https://github.com/cloudflare/workers-rs/pull/529/commits/8a50b68017e9e00d61aa3addd1134c62b23e0ec4

worker/src/crypto.rs
dakom: interesting... I haven't used `SendFuture` and don't quite understand it since wasm can't actually send futures between threads... I suppose it's for making the futures work with "multithreaded" (but not really) frameworks?

anyway - no objection from me, committing

worker/src/crypto.rs
dakom: you're right, removed in https://github.com/cloudflare/workers-rs/pull/529/commits/0527af1f904db199415207bc2044f63c12a0b5e8 :)

worker/src/crypto.rs
kflansburg: Yep, exactly, we need to tell Rust that its safe in this context.

kflansburg (approved):

PR #543, opened 2024-04-10
AI prototype with new bindings

kflansburg:


smallbraingames:
What is the status of this? looking to use it

kflansburg:
This pretty much works, however types need to be defined for each of the various model types and their respective input / output formats. Is there a particular model you plan to use?

I'm also just trying to understand the value of this. What is the use case for invoking AI in Rust? Are there crates you would also be using that are better than what JavaScript offers?

smallbraingames:
I actually just stopped using rust and went back to ts. Using WASM for what I want

But was using the whisper model

BrandonDyer64:
Any updates on this? I'd like to start using the image-to-text and translation models in my worker.

gakonst:
Hey @kflansburg also interested in this -- are you looking for input / do you want someone to take this over?

kflansburg:
I believe that an alternative PR merged recently https://github.com/cloudflare/workers-rs/pull/679

Issue #545, opened 2024-04-12
🤾‍♀️ [Feature] Most interop methods should be marked with `catch`


Jasper-Bekkers:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Given how prevalent it is in the node.js side of things to handle errors through exceptions, I think we should mark most methods in the worker-sys crate as `catch`. From the documentation, emphasis mine.

> By default wasm-bindgen **will take no action** when wasm calls a JS function which ends up throwing an exception. The wasm spec right now doesn't support stack unwinding and as a result Rust code will not execute destructors. This can unfortunately cause memory leaks in Rust right now, but as soon as wasm implements catching exceptions we'll be sure to add support as well!

https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/catch.html

Not handling catching these exceptions leads to the (quite frustrating) error message when developing.

> A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.                      
> X [ERROR] Uncaught (in response) Error: The script will never generate a response.                    

Obviously the error message is somewhat correct: the wasm module terminated without any reason as to why. Diving deeper and doing some debugging I found that this is mostly just because I'm passing bad inputs into the api's which are then detected by the node.js code, and bubbled up to us but we silently ignore them.  

However, getting `A hanging Promise was canceled.` messages because - say - my `part_number` is 0 when it should be 1 isn't helpful.

This would be quite a sweeping change though, and most likely cause a lot of `.unwrap` or `.expect` calls to have to be added to the wrapping library, but it would make development quite a bit nicer for the folks using this library so I'd like to discuss possible ways forward. I don't think `wasm-bindgen` has a catch-all way of dealing with this unfortunately (not that I could find at least).

Jasper-Bekkers:
Some stats: currently there are 166 methods exposed through `worker-sys` and 39 of those have the `catch` attribute.

Jasper-Bekkers:
https://github.com/cloudflare/workers-rs/pull/546 Is an initial attempt at trying to fix these issues.

Jasper-Bekkers:
Example from my follow up #547. With these changes if you pass an object with `part_number` into `complete` you'll get this error message:

```
[...]

PR #546, opened 2024-04-12
🤽 `worker-sys` Make all methods `catch`

Jasper-Bekkers:
This is quite a big PR trying to address https://github.com/cloudflare/workers-rs/issues/545 that should serve as a starting point/discussion point wrt to using `catch` project wide in all `method`, `constructor` and `static_method_of` calls.

This is a blanket replacement and fixup of all calls into `worker-sys` without changing the higher level API. This means that we introduce a few `.unwrap` calls and that I'll piggyback on methods that already were returning a `Result` so we can `?` unwrap all the errors.

Going forward, if we choose to go down this route, I'd suggest enforcing this as a project-wide rule.

### DIFF
--- a/worker-sandbox/src/d1.rs
+++ b/worker-sandbox/src/d1.rs
@@ -96,7 +96,7 @@ pub async fn exec(mut req: Request, env: Env, _data: SomeSharedData) -> Result<R
         .await
         .expect("doesn't exist");
 
-    Response::ok(result.count().unwrap_or_default().to_string())
+    Response::ok(result.count()?.unwrap_or_default().to_string())
 }
 
 #[worker::send]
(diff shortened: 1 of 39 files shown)


Jasper-Bekkers:
> My only concern are the unwraps. I suppose this is still strictly better than before because the panic will contain more useful error messages, but I do wonder if we should bite the bullet and make the relevant methods all return Result.

Yeah my thought initially was that I don't want to change the public facing API, I wanted to leave that up to the project maintainers. I think the unwraps are fine in the places where they are since they're mostly in getters and like you say they'll contain relevant information rather then swallowing the error messages.

kflansburg (approved):

PR #547, opened 2024-04-12
 💃Multi part upload

Jasper-Bekkers:
This sits on top of my previous PR (but could be rebased).

For a more reasonable diff check out afed6e2f28a6a80d7d8264a899ab78967217867b

Currently users could never call `complete` on `MultipartUpload` correctly - or rather - they could call `complete` but they couldn't construct the `UploadedPart` that it requires as a parameter. Given that the flow typically is that these `UploadedPart`'s are serialized over the wire (rather then stored for later use after a call to `upload_part`) and then passed to `complete` this causes a problem. This PR changes `R2UploadedPart` to be more like `R2Checksums` so that a user calling into the API can construct the types themselves and feed them into `complete`.


Jasper-Bekkers:
Rebased on main.

worker-sys/src/types/r2/uploaded_part.rs
```diff
@@ -1,14 +1,42 @@
-use wasm_bindgen::prelude::*;
+use js_sys::{Object, Reflect};
+use wasm_bindgen::JsValue;
 
-#[wasm_bindgen]
-extern "C" {
-    #[wasm_bindgen(extends=js_sys::Object)]
-    #[derive(Debug, Clone, PartialEq, Eq)]
-    pub type R2UploadedPart;
+#[derive(Debug, Clone)]
```
kflansburg, line 4: Is it possible to do something like [this](https://rustwasm.github.io/wasm-bindgen/reference/types/exported-rust-types.html) here? 

Or alternatively, could you just add a constructor to the old implementation? https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/constructor.html

Jasper-Bekkers: Constructor was what I tried first, that didn't work because it for some reason unbeknownst to Mr wouldn't understand and the fields when passed to `complete`.

I'll give the link a tie

kflansburg:
Closing in favor of #552

PR #551, opened 2024-04-18
Experimental RPC Support

kflansburg:
This introduces RPC support in Rust with some limitations:
* Currently just uses `wasm-bindgen` directly, as we understand use cases more we can introduce an `#[rpc]` macro to simplify things. 
* I haven't been able to get function passing to work. That will have to be done as a follow up.


examples/rpc-client/src/lib.rs
```diff
@@ -0,0 +1,41 @@
+use wasm_bindgen::prelude::*;
+use wasm_bindgen::JsValue;
+
+use wasm_bindgen_futures::JsFuture;
+use worker::*;
+
+#[wasm_bindgen]
```
kflansburg, line 7: Ideally all of this could be code-gen from capnp or some other specification

spigaz: A general specification makes a lot of sense because of compatibility with implementations in JS.

But an implementation based on the tool rustdocs uses, like pavex does, might be more friendlier as it has the types from the rust side to generate a client crate.

On another perspective, I'm not sure if using the WASM Component Model, we couldn't target the same interface, allowing you to link the client, with the proxy or the real component, avoiding serialisation, a complete bypass.

In theory I think so, in practice, I believe so, if I recall correctly, it might be just a question of linking to different implementations at load time.

But the same tool can be used to generate capnp or the WIT files.

Its even possible to think of when you deploy a Service / DO, its public interface is processed and at runtime its linked directly to it or to a proxy generated on the fly.

So yes, in theory, the same trickery could allow direct C++ -> JS / Rust / Python, Rust -> Rust or Rust -> Proxy -> Rust.

Although I believe it should be feasible right now, much of the path seems the same, even if it isn't just yet for some reason.

In a recent talk at wasmIO they mentioned that C++ support was close by.
But regarding the Component Model in V8, I payed attention, but I didn't capture any mention, and they talked a lot about wasm at Google, V8, wasm gc...

kflansburg: Which tool is being used by rustdocs/pavex? They are both large projects. 

I have considered defining the types in Rust, but that makes it very difficult to support multiple languages. The idea for using WIT is very interesting. I will give it some thought.

spigaz: He uses rustdoc-json and rustdoc-types, the same used by cargo-semver-checks and cargo-check-external-types.

He explains how it does it at minute 41m https://www.youtube.com/watch?v=cMea6IMRk2s

He also uses a tool to generate the SDK crate, cargo px, he explains it at min 35.

The ideal would be an equivalent for C++, but if there isn't one, and that's the blocker, the types in rust can work as a proxy. If necessary you can use the comments to do some kind of annotation, so that when you process them you have all the info you require.

For the runtime proxy, there is always the option to generate it from the wit files, I'm just not sure if generating in rust would limit non rust languages.

kflansburg:
I've prototyped using WIT to define the interface for the RPC client. I wasn't able to find a quick way to leverage `wit-bindgen`, so for now I just wrote my own. All of the primitive types are supported, but I will need some follow up work to support `records` which I think can be supported using `serde_wasm_bindgen`. 

This seems like a pretty nice developer experience but I worry about edge cases of both WIT and Workers' RPC system. 

Just to be clear on the plan here, I'm focusing on supporting "schemaless" RPC with this PR. This will bridge through JS and is intended to get people using RPC as quickly as possible. 

After that, there are plans to support "schemaful" RPC as follows:

1. Host methods to send generic RPC methods over capnproto are exposed directly to Wasm guest, bypassing JavaScript. 
2. Wasm guests share capnp definition and use capnp codegen to include ser/de code in Wasm. The runtime considers these messages to be opaque and does not know the schema.

spigaz:
I have been gathering additional information...

The idea of providing a WIT interface for a DO / Service, and just allow the client to use it directly should work.
This pattern shows up frequently in examples of interoperability.
The only challenge seems to be to generate the WIT stub that actually serialises and makes the remote request, yes, what you are already doing.

The idea of providing a temporary WIT interface based on workers-rs till C++ support is available, should also work just fine.
It would allow support for other languages already supporting WIT like C#, Go and Swift (WasmKit) for instance.
When the support is available in C++, you just stop loading the workers-rs implementation.

lewsmith:
Changes to the shim.js here - https://github.com/cloudflare/workers-rs/commit/f3eb1eb535d9b248615db531490b77629338a439#r141667118 - has broken `@cloudflare/vitest-pool-workers` integrations using `SELF`, causing vitest to fail with the error `Handler does not export a fetch() function`.

Having to revert to v0.0.10 to workaround.

See original bug report/debugging here - https://github.com/cloudflare/workers-sdk/issues/5726.

kflansburg:
> Changes to the shim.js here - [f3eb1eb#r141667118](https://github.com/cloudflare/workers-rs/commit/f3eb1eb535d9b248615db531490b77629338a439#r141667118) - has broken `@cloudflare/vitest-pool-workers` integrations using `SELF`, causing vitest to fail with the error `Handler does not export a fetch() function`.
> 
> Having to revert to v0.0.10 to workaround.
> 
> See original bug report/debugging here - [cloudflare/workers-sdk#5726](https://github.com/cloudflare/workers-sdk/issues/5726).

Thanks for the heads up. Since this change follows the pattern recommended for [RPC workers](https://developers.cloudflare.com/workers/runtime-apis/rpc/#example), which may be expanded in the future to enable other workers functionality, I think this should be fixed in `vitest-pool-workers`. I believe you should be able to continue using the older version of `worker-build` for the foreseeable future, that package does not change frequently.

und3fined:
@kflansburg sorry ping, your rpc-server sample is basic, can you write a example for [​​Class Instances](https://developers.cloudflare.com/workers/runtime-apis/rpc#class-instances) or [Functions](https://developers.cloudflare.com/workers/runtime-apis/rpc#functions). It's definitely more handy than `fn add()`

kflansburg:
> @kflansburg sorry for ping, your rpc-server sample is basic, can you write a example for [​​Class Instances](https://developers.cloudflare.com/workers/runtime-apis/rpc#class-instances) or [Functions](https://developers.cloudflare.com/workers/runtime-apis/rpc#functions). It's definitely more handy than `fn add()`
> 
> So, my plan is to set up an rpc server, generate some other services in JS, and then make calls to the rpc server.

Unfortunately this functionality has not yet been worked out, as noted in our [documentation](https://docs.rs/worker/latest/worker/#rpc-support).

und3fined:
> > @kflansburg sorry for ping, your rpc-server sample is basic, can you write a example for [​​Class Instances](https://developers.cloudflare.com/workers/runtime-apis/rpc#class-instances) or [Functions](https://developers.cloudflare.com/workers/runtime-apis/rpc#functions). It's definitely more handy than `fn add()`
> > So, my plan is to set up an rpc server, generate some other services in JS, and then make calls to the rpc server.
> 
> Unfortunately this functionality has not yet been worked out, as noted in our [documentation](https://docs.rs/worker/latest/worker/#rpc-support).

Sure bro, i see. But any PoC make it work?

kflansburg:
> Sure bro, i see. But any PoC make it work?

No, I tried it out and just know that it won't be easy.

TannerRogalsky:
I haven't tracked down exactly what the issue is but this change has broken my ability to connect websockets directly to a worker in local development.

TannerRogalsky:
Yeah, I'm sorry for such an underdeveloped report. I believe this is reproducible simply by running a worker locally and attempting to connect to it via a websocket client.

Is the version of miniflare used pinned to this crate? My experience is that rolling back to `0.0.10` of `worker-build` with no other changes fixed my connectivity problems.

If you're not able to reproduce this report with the existing websocket examples in local development then I will attempt to further narrow down what's actually causing my problem.

kflansburg:
I just think there might be gaps in local dev supporting the new `WorkersEntrypoint` interface, similar to https://github.com/cloudflare/workers-rs/pull/551#issuecomment-2093940194. 

I will give this a shot and see what I can find.

kflansburg:
I don't appear to be able to reproduce this. Using `wrangler==3.63.1`, `worker-build==0.1.0`, `workerd==2024-07-01`. 

I'm able to test a simple WebSocket example:

```rust
    let pair = WebSocketPair::new()?;
    let server = pair.server;
    server.accept()?;

    wasm_bindgen_futures::spawn_local(async move {
        let mut event_stream = server.events().expect("could not open stream");
        while let Some(event) = event_stream.next().await {
            match event.expect("received error in websocket") {
                WebsocketEvent::Message(msg) => console_log!("{:#?}", msg),
                WebsocketEvent::Close(_) => console_log!("Closed!"),
            }
        }
    });

    Response::from_websocket(pair.client)
```

and then I'm able to send and see messages logged by the server with `websocat ws://localhost:8787`.

TannerRogalsky:
Alright, thanks for trying. I'll try to figure out what the actual root of the problem is this week. I suspect it's related to that link you shared yesterday.

PR #552, opened 2024-04-19
feat: Allow instantiating an UploadPart from its constituent parts

### LINKED ISSUE #316: [BUG] `UploadedPart` is unserializable/constructable therefore cannot be passed to MultipartUpload::complete in a subsequent request

0.0.15

### Describe the bug

Because MultipartUpload::complete requires instances of `UploadedPart`, we need to be able to construct or deserialize the `UploadedPart` iterable list from the request body. Ideally a method would be available to pass the request or its body directly to a constructor to avoid having to manage deserialization in my worker code. Multipart uploads to R2 do not appear to be supported until this is fixed.

### Steps To Reproduce

_No response_

lwansbrough:
Fixes #316

Objective here was to provide an easy way to create an `UploadedPart` -- currently in my worker code I'm deserializing a custom object into a `part_number` and an `etag`, which I then use to construct the `UploadedPart` to be passed to `MultipartUpload::complete`

### DIFF
--- a/worker/src/r2/mod.rs
+++ b/worker/src/r2/mod.rs
@@ -330,12 +330,13 @@ pub struct UploadedPart {
 
 impl UploadedPart {
     pub fn new(part_number: u16, etag: String) -> Self {
-        let obj: JsValue = js_sys::Object::new().into();
+        let obj = js_sys::Object::new();
         Reflect::set(&obj, &JsValue::from_str("partNumber"), &JsValue::from_f64(part_number as f64)).unwrap();
         Reflect::set(&obj, &JsValue::from_str("etag"), &JsValue::from_str(&etag)).unwrap();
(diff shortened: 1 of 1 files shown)


kflansburg:
@Jasper-Bekkers does this work in place of your PR?

lwansbrough:
Sorry, I pushed this before testing. 😬 I don't think the JsValues need to be wrapped in a promise.

Jasper-Bekkers:
> @Jasper-Bekkers does this work in place of your PR (#547)?

It should,  yeah.

Jasper-Bekkers:
@lwansbrough it looks like the CI is failing on formatting, would you mind updating the PR so it can be merged?

kflansburg (approved):

PR #564, opened 2024-05-02
Use u64 for R2 range requests

### LINKED ISSUE #560: [BUG] Range requests in R2 are limited to 32 bit offset/length

0.2.0

3.53.0

### Describe the bug

The `worker::r2::builder::Range` enum is defined like this:
```rust
pub enum Range {
    OffsetWithLength { offset: u32, length: u32 },
    OffsetWithOptionalLength { offset: u32, length: Option<u32> },
    OptionalOffsetWithLength { offset: Option<u32>, length: u32 },
    Suffix { suffix: u32 },
}
```

Using `u32` for all the quantities makes the interface completely unable to handle files more than 4 GiB in size. I've done a bit of research, and it seems the code casts 32-bit numbers to `f64` for JS (essentially supporting 52 bits), and those are converted into `u64` in the workerd C++ code.

I can see multiple ways to fix this to varying degrees:
1. Simply changing Rust high-level definition to `u64`, low-level `R2Range` definition to `f64`, and erroring out during conversion if the number does not fit into `f64` without precision loss. Will fix the practical problem with the least effort, but this is certainly a hack. Values read from the response could be silently wrong if they exceed 52 bits, though (should not be a practical issue).
2. There seems to be a (somehow [undocumented](https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2getoptions)) way to pass a `Headers` instance with pre-made `Range` header. The Rust code could format the header manually from `u64` values, bypassing the JS `f64` dance entirely.
3. Support `BigInt` in the JS interface to pass through `u64` losslessly?

As a side note, the current `Range` Rust definition seems very weird to me: there is way too much overlap between the first three variants, and some semantics are not clear to me. What does `OptionalOffsetWithLength{None, L}` even mean? Is it the same as `OffsetWithLength{0, L}`/`OffsetWithOptionalLength{0, Some(L)}`? Or `Suffix{L}`? I think that the Range semantics could be made simpler, and it could be done simultaneously with the required breaking change to fix the data types. Perhaps this could work?
```rust
pub enum Range {
    OffsetWithLength { offset: u64, length: u64 },
    FromOffset { offset: u64 },
    Suffix { length: u64 },
}
```

### Steps To Reproduce

_No response_

kflansburg:
Closes #560

Waiting for guidance on the behavior of `limit` without `offset`. For now I'd like to continue to match the JavaScript behavior.

### DIFF
--- a/worker/src/r2/builder.rs
+++ b/worker/src/r2/builder.rs
@@ -93,34 +93,44 @@ impl From<Conditional> for JsObject {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Range {
-    OffsetWithLength { offset: u32, length: u32 },
-    OffsetWithOptionalLength { offset: u32, length: Option<u32> },
-    OptionalOffsetWithLength { offset: Option<u32>, length: u32 },
-    Suffix { suffix: u32 },
+    OffsetWithLength { offset: u64, length: u64 },
+    OffsetWithOptionalLength { offset: u64, length: Option<u64> },
+    OptionalOffsetWithLength { offset: Option<u64>, length: u64 },
+    Suffix { suffix: u64 },
+}
+
+const MAX_INT: u64 = (2 ^ 53) - 1;
+
+fn check_range_precision(value: u64) -> f64 {
+    assert!(
+        value <= MAX_INT,
+        "Integer precision loss when converting to JavaScript number"
+    );
+    value as f64
 }
 
 impl From<Range> for JsObject {
     fn from(val: Range) -> Self {
         match val {
             Range::OffsetWithLength { offset, length } => js_object! {
-                "offset" => Some(offset),
-                "length" => Some(length),
+                "offset" => Some(check_range_precision(offset)),
+                "length" => Some(check_range_precision(length)),
                 "suffix" => JsValue::UNDEFINED,
             },
             Range::OffsetWithOptionalLength { offset, length } => js_object! {
-                "offset" => Some(offset),
-                "length" => length,
+                "offset" => Some(check_range_precision(offset)),
+                "length" => length.map(check_range_precision),
                 "suffix" => JsValue::UNDEFINED,
             },
             Range::OptionalOffsetWithLength { offset, length } => js_object! {
-                "offset" => offset,
-                "length" => Some(length),
+                "offset" => offset.map(check_range_precision),
+                "length" => Some(check_range_precision(length)),
                 "suffix" => JsValue::UNDEFINED,
             },
             Range::Suffix { suffix } => js_object! {
                 "offset" => JsValue::UNDEFINED,
                 "length" => JsValue::UNDEFINED,
-                "suffix" => Some(suffix),
+                "suffix" => Some(check_range_precision(suffix)),
             },
         }
     }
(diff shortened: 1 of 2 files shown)


kflansburg:
Looks like Rust `1.78` introduced some new clippy lints. I've fixed ours, but looks like `wasm-bindgen` may need to fix some as well. https://github.com/rustwasm/wasm-bindgen/issues/3945

kflansburg:
@himikof How does this look? I'm still waiting to hear back on the intended behavior of the range API.

worker/src/r2/builder.rs
```diff
@@ -93,34 +93,44 @@ impl From<Conditional> for JsObject {
...
-    OffsetWithOptionalLength { offset: u32, length: Option<u32> },
-    OptionalOffsetWithLength { offset: Option<u32>, length: u32 },
-    Suffix { suffix: u32 },
+    OffsetWithLength { offset: u64, length: u64 },
+    OffsetWithOptionalLength { offset: u64, length: Option<u64> },
+    OptionalOffsetWithLength { offset: Option<u64>, length: u64 },
+    Suffix { suffix: u64 },
+}
+
+const MAX_INT: u64 = (2 ^ 53) - 1;
```
himikof, line 102:
```suggestion
const MAX_SAFE_INTEGER: u64 = js_sys::Number::MAX_SAFE_INTEGER as u64;
```
This way, the assert semantics will be obvious.

himikof:
This change should fix the practical issue, allowing to use the API with large files. 

But I still think that the API does not really match the JS from the user point of view, and does not feel like a proper Rust API. In JS, you just specify either some non-empty subset of {`offset`, `length`}, or `suffix` when specifying a request range, and just access the 3 fields (checking them for `undefined`) when reading a response range.
But in Rust, you must explicitly select one of the enum Variants in the first case (leading to questions like "what is the difference?", "what if I chose the wrong one?"), and even worse, handle all of them in the second case (leading to code duplication). I do not think that to match the JS behavior you need to copy the TypeScript type annotation (which is really an implementation detail many users won't care about) into the Rust API, making it very explicit for all users. Note how the `impl TryFrom<R2RangeSys> for Range` never constructs `OffsetWithOptionalLength` with `Some` length, or `OptionalOffsetWithLength` with `Some` offset: those fields are not needed to match JS semantics at all.

Furthermore, this is actually not some Cloudlare-specific API: I would expect it to match the well-known semantics of simple HTTP `range` header (with only one range). You would expect it to have 3 non-overlapping possibilities: you can requesting a specific range (offset+length), a suffix from the given offset, or a suffix of given length. It is well-known that the `range` header has this structure, so not having the API reflect this leads to unnecessary surprise.

As this PR already is a breaking change for this API (no implicit integer conversions in Rust), I ask to reconsider changing the Rust enum shape to remove this overlap, and better match the use cases/expectations.

kflansburg:
> As this PR already is a breaking change for this API (no implicit integer conversions in Rust), I ask to reconsider changing the Rust enum shape to remove this overlap, and better match the use cases/expectations.

I'm waiting to hear back from the team on the actual behavior of these options. Without this information, I do not want to make a decision here.

kflansburg:
@himikof I've updated `Range` to have more clear options / documentation. I did keep 4 options, although you could achieve `Prefix` using `OffsetWithLength { offset: 0, length N }`, I think this is more clear.

PR #566, opened 2024-05-03
Hyperdrive binding

avsaase:


worker-sys/src/types/hyperdrive.rs
```diff
@@ -0,0 +1,26 @@
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends=js_sys::Object)]
+    #[derive(Clone, PartialEq, Eq)]
+    pub type Hyperdrive;
+
+    #[wasm_bindgen(method, getter, js_name=connectionString)]
+    pub fn connection_string(this: &Hyperdrive) -> String;
```
avsaase, line 10: Open questions
1. Is `method` the correct attribute?
2.  Are these property accessors fallible? Should these return a `Result`?

worker-sys/src/types/hyperdrive.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    #[wasm_bindgen(method, getter, js_name=host)]
+    pub fn host(this: &Hyperdrive) -> String;
+
+    #[wasm_bindgen(method, getter, js_name=port)]
+    pub fn port(this: &Hyperdrive) -> u16;
```
avsaase, line 16: On the JS side this is a `number` (https://github.com/cloudflare/workerd/blob/a1055534197598c01b9f8b2d0d4d8f7e27de13c5/types/defines/hyperdrive.d.ts#L32). Can I access it as a `u16` on the Rust side?

avsaase (commented):
This is the first that I work with `wasm_bindgen` so I am not sure on some of the details.

kflansburg:
This is exciting, could this work with `sqlx`? I assume you would connect directly to the database for development but could use Hyperdrive in production.

avsaase:
I found [this blog post](https://kian.org.uk/using-cloudflare-hyperdrive-with-tokio-postgres-in-workers/) explaining how to use Hyperdrive  with `tokio-postgress`. Unfortunately it depends on a branch who's PR has since been closed by the author. I haven't looking into what the problem with upstream `tokio-postgress` is and if the same limitations apply to `sqlx`.

worker-sys/src/types/hyperdrive.rs
kflansburg: I do believe that `method` is correct. I don't think these attributes could fail, although I suppose it couldn't hurt to catch it.

worker-sys/src/types/hyperdrive.rs
kflansburg: I'm pretty sure the answer is yes, except for the larger integer types that map to BigInt.

kflansburg (approved):

kflansburg:
> I found [this blog post](https://kian.org.uk/using-cloudflare-hyperdrive-with-tokio-postgres-in-workers/) explaining how to use Hyperdrive with `tokio-postgress`. Unfortunately it depends on a branch who's PR has since been closed by the author. I haven't looking into what the problem with upstream `tokio-postgress` is and if the same limitations apply to `sqlx`.

Hmmm, ok yeah I think I also remember that `sqlx` makes it difficult to support the custom startTls implementation that our Socket API uses, but maybe Hyperdrive gets around that since it originates TLS.

avsaase:
Thanks for merging. Where you able to confirm that this works? I don't have Hyperdrive set up to test it.

kflansburg:
> Thanks for merging. Where you able to confirm that this works? I don't have Hyperdrive set up to test it.

No, I will give it a shot this evening.

PR #568, opened 2024-05-04
Add `encodeBody` and `cf` to ResponseInit. Adopt builder pattern.

### LINKED ISSUE #567: [BUG] No way to return pre-compressed data with Content-Encoding from worker-rs due to missing encodeBody option

0.2.0

3.53.0

### Describe the bug

According to https://stackoverflow.com/a/64849685 and [encodeBody documentation](https://developers.cloudflare.com/workers/runtime-apis/response/#properties), the only way to return already-gzipped data from worker (without decompressing and compressing again) with `Content-Encoding: gzip` set (for the browser to decompress it automatically) is to set the `encodeBody` Response init option to `"manual"`:
```js
// Make a new response with the same body but using manual encoding.
resp = new Response(resp.body, {
  status: resp.status,
  headers: resp.headers,
  encodeBody: "manual"
});

// Modify headers and return.
resp.headers.set("Content-Encoding", "gzip");
return resp;
```

If `encodeBody: "manual"` is missing, Worker runtime will silently compress the body again, returning gzip-in-gzip data to the browser.

I have failed to find any reasonable way to set this option in `worker-rs`, because it can only be passed through `ResponseInit` object, and the current worker code:
1. has no support for passing anything apart from `status`, `headers` and `websocket` through `ResponseInit`;
2. has no support for using custom `ResponseInit` instances;
3. constructs the `ResponseInit` object too late (after the `fetch` handler has completed processing), so the handler cannot customize it further.

Also, the [encodeBody documentation](https://developers.cloudflare.com/workers/runtime-apis/response/#properties) is a bit confusing: it makes it seem like `encodeBody` is a property on `Response`, but according to my reading of workerd code, it can be only passed during `Response` construction through `init` (`ResponseInit.encodeBody` property, essentially), and is not exposed on the `Response` object in any way.

### Steps To Reproduce

_No response_

kflansburg:
This expands `ResponseInit` to support all of the options available in JavaScript. Additionally, the builder pattern is adopted so that users do not have to specify all of these new fields as we add them (breaking change). There are a few questions I want to answer before merging this. 

Closes #567

### DIFF
--- a/worker-sandbox/src/cache.rs
+++ b/worker-sandbox/src/cache.rs
@@ -2,7 +2,7 @@ use super::SomeSharedData;
 use futures_util::stream::StreamExt;
 use rand::Rng;
 use std::time::Duration;
-use worker::{console_log, Cache, Date, Delay, Env, Request, Response, Result};
+use worker::{console_log, Cache, Date, Delay, Env, Request, Response, ResponseBuilder, Result};
 
 fn key(req: &Request) -> Result<Option<String>> {
     let uri = req.url()?;
@@ -24,12 +24,11 @@ pub async fn handle_cache_example(
         Ok(resp)
     } else {
         console_log!("Cache MISS!");
-        let mut resp =
-            Response::from_json(&serde_json::json!({ "timestamp": Date::now().as_millis() }))?;
-
         // Cache API respects Cache-Control headers. Setting s-max-age to 10
         // will limit the response to be in cache for 10 seconds max
-        resp.headers_mut().set("cache-control", "s-maxage=10")?;
+        let mut resp = ResponseBuilder::new()
+            .with_header("cache-control", "s-maxage=10")?
+            .from_json(&serde_json::json!({ "timestamp": Date::now().as_millis() }))?;
         cache.put(key, resp.cloned()?).await?;
         Ok(resp)
     }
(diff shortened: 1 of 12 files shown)


worker/src/response.rs
```diff
@@ -356,9 +347,73 @@ fn no_using_invalid_error_status_code() {
...
+    pub fn builder() -> ResponseInitBuilder {
+        ResponseInitBuilder::new()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ResponseInitBuilder {
+    status_code: u16,
+    headers: Headers,
+    websocket: Option<WebSocket>,
```
kflansburg, line 367: Is it actually valid to clone a `WebSocket`. In https://github.com/cloudflare/workers-rs/commit/9042648b8a5ce6879a1f36ff9b278ad7258d7c15, I see `WebSocket` explicitly cloned, but I also see an explicit check to prevent cloning `WebSocket` in the same commit.

worker/src/response.rs
```diff
@@ -306,33 +290,40 @@ impl Response {
...
+        &mut self.init.headers
     }
 
     /// Clones the response so it can be used multiple times.
     pub fn cloned(&mut self) -> Result<Self> {
-        if self.websocket.is_some() {
+        if self.init.websocket.is_some() {
             return Err(Error::RustError("WebSockets cannot be cloned".into()));
         }
+        // This information is lost when we make the roundtrip though web_sys::Response
```
kflansburg, line 320: Waiting to confirm if these properties are available on `Request`, which would allow for the round trip.

worker-sandbox/src/cache.rs
```diff
@@ -2,7 +2,7 @@ use super::SomeSharedData;
 use futures_util::stream::StreamExt;
```
kflansburg, line 2: These changes to the other files aren't __necessary__ (there are no breaking changes to `Response`), but are done to adopt the new builder pattern.

worker/src/response.rs
```diff
@@ -355,19 +320,209 @@ fn no_using_invalid_error_status_code() {
...
+    }
+
+    /// Sets this response's cors headers from the `Cors` struct.
+    /// Example usage:
+    /// ```
+    /// use worker::*;
+    /// fn fetch() -> worker::Result<Response> {
+    ///     let cors = Cors::default();
+    ///     Response::empty()?
+    ///         .with_cors(&cors)
```
himikof, line 370: The example usage does not match the method (`ResponseBuilder::with_cors`).

worker/src/response.rs
zebp: It is valid to clone the `WebSocket` struct since you're just cloning a *reference* to a JavaScript WebSocket object, but when we call `.clone()` on `EdgeResponse` we're actually calling [a JS clone method](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone). In workerd this'll throw if the `ResponseInit` contains a websocket, so we check before hand to ensure this doesn't throw an Error in JS.

PR #569, opened 2024-05-04
Update` tokio-postgres` example to demonstrate Hyperdrive

kflansburg:


kflansburg:
@xortive, might need your help to get this working. I believe that using `StartTLS` is the right way to go based on `node-postgres` code. I'm pretty confident that this example will not use a prepared statement. 

I get the following errors:
```
db error: FATAL: Internal error.
Error: Network connection lost.
```

cc @avsaase

kflansburg:
I've tried disabling TLS entirely, as well as constructing the `Config` object without the full connection string (just username, password, and database). Same error both times. Could be a bug in the `workers-rs` Socket `Stream` implementation, but difficult to narrow things down.

KianNH:
> I'm pretty confident that this example will not use a prepared statement.

`simple_query` won't use prepared statements but `query` will, I believe. The issue is that tokio-postgres doesn't support unnamed statements and Hyperdrive doesn't support prepared statements - so `query` won't work.

I had this issue in my [blog](https://kian.org.uk/using-cloudflare-hyperdrive-with-tokio-postgres-in-workers/) about using `tokio-postgres` in `workers-rs`. I didn't use Tls though, that's likely a separate issue, but I think you will run into issues using `query` regardless.

kflansburg:
> `simple_query` won't use prepared statements but `query` will, I believe. The issue is that tokio-postgres doesn't support unnamed statements and Hyperdrive doesn't support prepared statements - so `query` won't work.

This fixed it, thanks!

PR #584, opened 2024-06-07
Expose `Hyperdrive` from the worker crate

ramnivas:
This allows users to pass around the instance of `Hyperdrive`.


kflansburg:
Huh, surprised that Rust didn't complain about the private type in a public API `Env::hyperdrive`.

kflansburg (approved):

PR #591, opened 2024-07-04
Add a binding to meta object for `D1Result`.

langyo:
Thanks to the official staff for reviewing. This PR is based on #337 .

When I tried to use `sea-orm` proxy connection, I found that the `D1Result` object could not obtain other information except data, so I traced the source code.

In fact, `worker-sys` has already declared the `meta` function, but `D1Result` under `worker` did not further extract and serialize this part of data. I made some simple work that `D1Result` can obtain this part of the content.

[Demo is here](https://github.com/langyo/cloudflare_worker_rust_demo).

### DIFF
--- a/worker/src/d1/mod.rs
+++ b/worker/src/d1/mod.rs
@@ -387,10 +387,10 @@ impl D1Result {
 
     /// Return the meta data in this result.
     ///
-    /// Returns `None` if it's not exists.
+    /// Returns `None` if `meta` field is not populated.
     pub fn meta(&self) -> Result<Option<D1ResultMeta>> {
         if let Ok(meta) = self.0.meta() {
-            let meta: D1ResultMeta = serde_wasm_bindgen::from_value(meta.into()).unwrap();
+            let meta: D1ResultMeta = serde_wasm_bindgen::from_value(meta.into())?;
             Ok(Some(meta))
         } else {
             Ok(None)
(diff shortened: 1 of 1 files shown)


worker/src/d1/mod.rs
```diff
@@ -372,6 +384,18 @@ impl D1Result {
             Ok(Vec::new())
         }
     }
+
+    /// Return the meta data in this result.
+    ///
+    /// Returns `None` if it's not exists.
```
kflansburg, line 390: I'm actually not sure when this would not be included in the result.
```suggestion
    /// Returns `None` if `meta` field is not populated.
```

worker/src/d1/mod.rs
```diff
@@ -372,6 +384,18 @@ impl D1Result {
...
+    pub fn meta(&self) -> Result<Option<D1ResultMeta>> {
+        if let Ok(meta) = self.0.meta() {
+            let meta: D1ResultMeta = serde_wasm_bindgen::from_value(meta.into()).unwrap();
```
kflansburg, line 393: Could you remove this unwrap?

kflansburg (changes requested):

worker/src/d1/mod.rs
langyo: > Could you remove this unwrap?

Okay, I neglected it.

kflansburg (approved):

Issue #423, opened 2023-12-04
[BUG] set-cookie twice / header folds


ostenbom:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What version of `workers-rs` are you using?

0.0.18

### What version of `wrangler` are you using?

0.0.18

### Describe the bug

When calling a cloudflare worker in `workers-rs`, the `set-cookie` header "folds" and creates a single `set-cookie = cookie1=value1, cookie2=value2`.

According to [cloudflare docs](https://developers.cloudflare.com/workers/runtime-apis/headers/#differences) and the [IETF RFC](https://www.rfc-editor.org/rfc/rfc6265), the set-cookie header should not fold.

We do not have the same behaviour in our javascript workers.

It's a tricky case when looking at the headers interface, but maybe a similar `getAll` interface is needed here?

This makes it slightly more tricky to act as a reverse-proxy and forward requests from our worker.

### Steps To Reproduce

1. Make a cf workers-rs worker
2. Log request.headers
3. `curl -H "Set-Cookie: cookie1=value1" -H "Set-Cookie: cookie2=value2" http://0.0.0.0:8787`
4. Observe `set-cookie = cookie1=value1, cookie2=value2`

nakamura-shuta:
Should I implement the get_all function in Headers (headers.rs) with the following specification?

```
Return all values of the `Set-Cookie` header in the `Headers` object with
Returns them as a Result<Vec<String>> type.
Also return an error if the name is not `Set-Cookie`.
```

PR #597, opened 2024-07-18
Implement get_all function to return non-folding set-cookie headers

nakamura-shuta:
issue : https://github.com/cloudflare/workers-rs/issues/423

Implement get_all method for Headers using wasm-bindgen

This PR implements the get_all method for Headers with the following approach:

1. Define the get_all method in worker_sys::ext::headers::Headers using wasm-bindgen
2. Create a HeadersExt trait in worker_sys to extend web_sys::Headers functionality
3. Implement get_all in worker::Headers using the worker_sys implementation

Key changes:
- Add a new file worker-sys/src/ext/headers.rs to define Headers type and get_all method
- Create HeadersExt trait to extend web_sys::Headers with get_all method
- Update worker/src/headers.rs to use the new HeadersExt trait
- Modify the get_all implementation in worker::Headers to use the new wasm-bindgen based method

Specification of the get_all method:
- Retrieves all values of a given header name
- Returns the result as Result<Vec<String>>
- Uses wasm-bindgen to directly call the JavaScript Headers.getAll() method

This implementation allows for retrieving multiple values for any header, not just 'Set-Cookie', 
providing more flexibility and aligning with the Web API's Headers.getAll() method.


kflansburg:
Why limit only to `set-cookie` header?

nakamura-shuta:
The workerd code I used as reference for implementation made an error when a header other than `set-cookie` was specified, so I implemented it as is.

https://github.com/cloudflare/workerd/blob/a265d5c22bf2a6bc6394aa54db517ff86d6c788c/src/workerd/api/http.c%2B%2B#L229-L242

kflansburg:
I see, I can ask internally about this. 

If this is already implemented in the runtime, would it make sense to just add the `getAll` method to the wasm_bindgen Header extension type here https://github.com/cloudflare/workers-rs/blob/main/worker-sys/src/ext/headers.rs?

nakamura-shuta:
Yes, that seems to be the simplest way.
I will try to implement it that way.

kflansburg:
Can we invoke the JavaScript `getAll` method via `wasm-bindgen`, rather than re-implementing it in Rust?

worker-sys/src/ext/headers.rs
```diff
@@ -0,0 +1,20 @@
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
```
kflansburg, line 3: Does this need to be buried in a `glue` module like some of the other extensions?

kflansburg (commented):
This looks great! Were you able to test this?

worker-sys/src/ext/headers.rs
nakamura-shuta: Yes, I have included it in the glue module to match the others. Included in the glue module along with others

nakamura-shuta:
@kflansburg

> This looks great! Were you able to test this?

I created the following application under the example directory and verified its operation with curl.

```
/// create sample application. examples/cookie/src/lib.rs
use worker::*;

// Main entry point for the worker
#[event(fetch)]
pub async fn main(req: Request, env: Env, _ctx: worker::Context) -> Result<Response> {
    // Set up a router and define a route for "/test_cookies"
    Router::with_data(())
        .get_async("/test_cookies", test_cookies)
        .run(req, env)
        .await
}

// Handler function for the "/test_cookies" route
async fn test_cookies(req: Request, _ctx: RouteContext<()>) -> Result<Response> {
    // Retrieve all Set-Cookie headers from the input request
    let input_cookies = req.headers().get_all("Set-Cookie")?;

    // Create a new response
    let mut res = Response::ok("Cookies processed")?;

    // Add the input Set-Cookie headers to the new response
    for cookie in &input_cookies {
        // Append each Set-Cookie header to the response
        res.headers_mut().append("Set-Cookie", cookie)?;
    }

    // Log the request and response headers for debugging
    console_log!("Request Set-Cookie headers: {:?}", input_cookies);
    console_log!("Response headers: {:?}", res.headers());

    // Return the response
    Ok(res)
}
```

```
# curl command in console

% curl 'http://localhost:8787/test_cookies' \
  -H 'Set-Cookie: a=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT' \
  -H 'Set-Cookie: b=2; Expires=Wed, 21 Oct 2015 07:28:00 GMT' \
  -v
* Host localhost:8787 was resolved.
> GET /test_cookies HTTP/1.1
> ・・・・・・・
> Set-Cookie: a=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT
> Set-Cookie: b=2; Expires=Wed, 21 Oct 2015 07:28:00 GMT
>
< HTTP/1.1 200 OK
< Content-Length: 17
< Content-Type: text/plain; charset=utf-8
< Set-Cookie: a=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT
< Set-Cookie: b=2; Expires=Wed, 21 Oct 2015 07:28:00 GMT
<
* Connection #0 to host localhost left intact
Cookies processed
```

```
# wrangler dev console

% cd workers-rs/examples/cookie
% npx wrangler dev

[wrangler:inf] Ready on http://localhost:8787
Request Set-Cookie headers: ["a=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT", "b=2; Expires=Wed, 21 Oct 2015 07:28:00 GMT"]
Response headers: Headers {
content-type = text/plain; charset=utf-8
set-cookie = a=1; Expires=Wed, 21 Oct 2015 07:28:00 GMT
set-cookie = b=2; Expires=Wed, 21 Oct 2015 07:28:00 GMT
}
[wrangler:inf] GET /test_cookies 200 OK (12ms)
```

PR #605, opened 2024-07-29
Relax type of callback arguments to Router methods

compiler-errors:
This PR relaxes the type of the callback

> `func: fn(Request, RouteContext<D>) -> T`

into

> `func: impl Fn(Request, RouteContext<D>) -> T + 'static`

which is an argument-position impl trait type (APIT), since I don't see any reason for requiring the passed-in callback is a function pointer, since we're immediately boxing it up and erasing the type.

Specifically, this ensures that you can pass a (currently nightly-only) async closure (`async ||`), which currently does *not* support being coerced to a fn ptr type. I'd like to avoid needing to implement that functionality that if possible.

### DIFF
--- a/worker/src/router.rs
+++ b/worker/src/router.rs
@@ -190,7 +190,11 @@ impl<'a, D: 'a> Router<'a, D> {
 
     /// Register an HTTP handler that will exclusively respond to HEAD requests. Enables the use of
     /// `async/await` syntax in the callback.
-    pub fn head_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
+    pub fn head_async<T>(
+        mut self,
+        pattern: &str,
+        func: impl Fn(Request, RouteContext<D>) -> T + 'a,
+    ) -> Self
     where
         T: Future<Output = Result<Response>> + 'a,
     {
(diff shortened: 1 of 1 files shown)


worker/src/router.rs
```diff
@@ -190,7 +190,11 @@ impl<'a, D: 'a> Router<'a, D> {
 
     /// Register an HTTP handler that will exclusively respond to HEAD requests. Enables the use of
     /// `async/await` syntax in the callback.
-    pub fn head_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
+    pub fn head_async<T>(
+        mut self,
+        pattern: &str,
+        func: impl Fn(Request, RouteContext<D>) -> T + 'static,
+    ) -> Self
     where
```
compiler-errors, line 198: Alternatively, this doesn't need to be an APIT, but could also just be a generic parameter.

```suggestion
    pub fn head_async<T>(
        mut self,
        pattern: &str,
        func: F,
    ) -> Self
    where
        F: Fn(Request, RouteContext<D>) -> T + 'static,
```

worker/src/router.rs
```diff
@@ -190,7 +190,11 @@ impl<'a, D: 'a> Router<'a, D> {
 
     /// Register an HTTP handler that will exclusively respond to HEAD requests. Enables the use of
     /// `async/await` syntax in the callback.
-    pub fn head_async<T>(mut self, pattern: &str, func: fn(Request, RouteContext<D>) -> T) -> Self
+    pub fn head_async<T>(
+        mut self,
+        pattern: &str,
+        func: impl Fn(Request, RouteContext<D>) -> T + 'static,
```
kflansburg, line 196: I guess my only concern is if the `'static` bound will break existing code in any circumstance.

PR #624, opened 2024-08-23
Add email worker and send email support 

### LINKED ISSUE #274: [Feature] Add email worker support

### Description

Please add email event/worker support

LuisDuarte1:
Closes #274 

Adds Email Worker support in workers-rs in addition to Send Email support


worker-build/src/js/shim.js
```diff
@@ -20,6 +20,12 @@ class Entrypoint extends WorkerEntrypoint {
     async scheduled(event) {
         return await imports.scheduled(event, this.env, this.ctx)
     }
+
+    // For some reason, email events doesn't seem to use WorkerEntrypoint so we get the env and ctx from
+    //  from the function itself.
+    async email(message, _env, _ctx) {
```
kflansburg, line 26: @ObsidianMinor do you know why email worker wouldn't use the new interface? Will cron triggers have the same issue?

worker-sys/src/types/email.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends=js_sys::Object)]
+    #[derive(Clone, PartialEq, Eq)]
+    pub type EmailMessage;
+
+    // TODO(lduarte): see if also accepting string is needed
+    #[wasm_bindgen(constructor, catch)]
+    pub fn new(from: &str, to: &str, raw: &str) -> Result<EmailMessage, JsValue>;
```
kflansburg, line 13: Should we also have a `new_from_stream` method that takes `ReadableStream`?

worker-sys/src/types/email.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    #[wasm_bindgen(method, getter)]
+    pub fn from(this: &EmailMessage) -> String;
+
+    #[wasm_bindgen(method, getter)]
+    pub fn to(this: &EmailMessage) -> String;
+}
+
+#[wasm_bindgen]
+extern "C" {
```
kflansburg, line 23: Should we add `headers` field? Should these getters also be on `EmailMessage`?

worker-sys/src/types/email.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub fn raw_size(this: &ForwardableEmailMessage) -> u32;
+
+    #[wasm_bindgen(method, catch, js_name=setReject)]
+    pub fn set_reject(this: &ForwardableEmailMessage, reason: &str) -> Result<(), JsValue>;
+
+    #[wasm_bindgen(method, catch)]
+    pub fn forward(
+        this: &ForwardableEmailMessage,
+        rcpt_to: &str,
+        headers: Headers,
```
kflansburg, line 48: This field should probably be optional, or maybe have `forward` and `forward_with_headers`. Or is there no difference between `headers` being `undefined` or `{}` in JavaScript API?

worker/src/email.rs
```diff
@@ -0,0 +1,207 @@
+use futures_util::TryStreamExt;
+use wasm_bindgen::JsCast;
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::JsFuture;
+use worker_sys::EmailMessage as EmailMessageExt;
```
kflansburg, line 5: I would typically call this `EmailMessageSys`.

worker/src/email.rs
```diff
@@ -0,0 +1,207 @@
...
+        self.0.set_reject(reason)
+    }
+
+    pub async fn forward(&self, rcpt_to: &str, headers: Option<Headers>) -> Result<(), JsValue> {
+        JsFuture::from(self.0.forward(rcpt_to, headers.unwrap_or_default().0)?).await?;
+        Ok(())
+    }
+
+    pub async fn reply(&self, message: EmailMessage) -> Result<(), JsValue> {
+        JsFuture::from(self.0.reply(message.0)?).await?;
```
kflansburg, line 150: You may want to wrap these async calls as follows, otherwise the method's Future will not be Send:
```rust
let promise = self.0.reply(message.0)?;
let fut = SendFuture::new(JsFuture::from(promise));
fut.await?;
```

Basically, you don't want to hold `JsFuture` across `await` boundary.

worker/src/email.rs
```diff
@@ -0,0 +1,207 @@
...
+use crate::EnvBinding;
+use crate::Error;
+use crate::Headers;
+
+pub struct EmailMessage(EmailMessageExt);
+
+unsafe impl Send for EmailMessage {}
+unsafe impl Sync for EmailMessage {}
+
+impl JsCast for EmailMessage {
```
kflansburg, line 19: I may be wrong, but you may not need all of this boilerplate for types that do not need to implement `EnvBinding`.

worker-build/src/js/shim.js
LuisDuarte1: I still haven't figured it out yet - for some reason `env` is still not probably propagated into the rust layer, and therefore I can't seem to use bindings with it :sweat_smile:

LuisDuarte1: It also quite difficult to debug because email workers can't be run locally :sweat_smile:

worker/src/email.rs
DougAnderson444: `Send` and `Sync` are autotraits, you can test whether they are automatically implemented by adding a test such as

```rust
fn is_normal<T: Sized + Send + Sync + Unpin>() {}

#[test]
fn normal_types() {
    is_normal::<EmailMessage>();
}
```

nudded:
As this PR seemed to be dead, I've got a working implementation and would love some comments.(https://github.com/cloudflare/workers-rs/pull/715)

PR #627, opened 2024-08-29
BREAKING: Implement more of the R2 API

fisherdarling:
This PR adds two missing parts of the R2 API:

- An (accidentally) undocmented `startAfter` option in R2's list objects call
- R2 PutObject Conditional Support. 
  - This is a BREAKING change. When a provided conditional to an R2 PutObject fails, the resultant Object could be `null`, or `None` in Rust. The returned object from a PUT is not a `Option<Object>`.

This API is successfully being used to implement a compare-and-swap (CAS) primitive for an internal project. We ended up not needing the list changes, but I thought it'd be useful nonetheless.

Tasks:
- [ ] clean up workers-rs tests

### DIFF
--- a/worker/src/r2/builder.rs
+++ b/worker/src/r2/builder.rs
@@ -65,7 +65,7 @@ impl GetOptionsBuilder<'_> {
     }
 }
 
-/// You can pass an [Conditional] object to [GetOptionsBuilder]. If the condition check fails,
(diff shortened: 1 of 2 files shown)


guybedford (commented):
This seems sensible to me and breaking changes like this are okay to make in major releases.

It would of course be nice to see tests though.

PR #639, opened 2024-09-20
Update semgrep.yml

hrushikeshdeshpande:
Updating Semgrep.yml file - Semgrep is a tool that will be used to scan Cloudflare's public repos for Supply chain, code and secrets. This work is part of Application & Product Security team's initiative.

In case of any questions, please reach out to "Hrushikesh Deshpande" on cf internal chat.


kflansburg (approved):
A PR for the original commit would have been appreciated...

PR #656, opened 2024-10-18
Workers Assets Binding

### LINKED ISSUE #644: Introduce Workers Assets binding

### Description

Allow access to static assets from user worker.

BrandonDyer64:
closes #644 

I went down a rabbit hole creating an entire new `Assets` env binding just to find out it's a `Fetcher`. In fact, you can use assets without this PR by just calling `env.get_binding::<Fetcher>("ASSETS")`.

Testing:
- Run `npx wrangler dev` in `worker-sandbox`
- Go to `/asset/test.txt`
- See `TEST` in browser
- Go to `/asset/doesnotexist.txt`
- See 404

### DIFF
--- a/worker-sandbox/src/assets.rs
+++ b/worker-sandbox/src/assets.rs
@@ -1,0 +1,17 @@
+#[cfg(not(feature = "http"))]
+pub async fn handle_asset(
+    req: worker::Request,
+    env: worker::Env,
+    _data: crate::SomeSharedData,
+) -> worker::Result<worker::Response> {
+    use worker::Url;
+
+    let url: Url = req.url()?;
+    let name: String = url.path_segments().unwrap().nth(1).unwrap().to_string();
+    let url: String = ["https://dummyurl.com/", &name].concat();
+    Ok(env
+        .assets("ASSETS")
+        .expect("ASSETS BINDING")
+        .fetch(url, None)
+        .await?)
+}
--- a/worker-sandbox/src/lib.rs
+++ b/worker-sandbox/src/lib.rs
@@ -6,7 +6,9 @@ use std::sync::{
 #[cfg(feature = "http")]
 use tower_service::Service;
 use worker::*;
+
 mod alarm;
+mod assets;
 mod cache;
 mod counter;
 mod d1;
--- a/worker-sandbox/src/router.rs
+++ b/worker-sandbox/src/router.rs
@@ -1,5 +1,5 @@
 use crate::{
-    alarm, cache, d1, fetch, form, kv, queue, r2, request, service, socket, user, ws,
+    alarm, assets, cache, d1, fetch, form, kv, queue, r2, request, service, socket, user, ws,
     SomeSharedData, GLOBAL_STATE,
 };
 #[cfg(feature = "http")]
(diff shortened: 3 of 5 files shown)


BrandonDyer64:
Maybe this isn't enough. I'm doing some testing and setting the following in my `wrangler.toml` will serve the assets, but not expose the binding:
```toml
[assets]
binding = "ASSETS"
directory = "./public/"
```
There's nothing related to assets in the "Your worker has access to the following bindings:" readout.

kflansburg:
Sorry that you went in circles on this. I think this looks correct. Have you tried deploying your worker and testing that? I will see if I can reproduce the issue in local development.

kflansburg:
This worked locally for me. You do need to specify a dummy URL (I think all Fetchers behave this way).

BrandonDyer64:
> You do need to specify a dummy URL

Ah, that was it. Adding `https://example.com` to the beginning of my URL fixed it for me.

BrandonDyer64:
Couple questions/feedback on this:

1. I don't see the `ASSETS` binding in the `Your worker has access to the following bindings:` printout. That seems like a bug, and it's why it took me so long to find a solution. I was chasing ghosts.
2. What's your recommendation for getting the `fetch()` response body as bytes? I had to import `http_body_util::BodyExt` and write this, just to get basic data out:
    - ```rust
      let bytes: Vec<u8> = response.into_body().collect().await?.to_bytes().to_vec();
      ```

kflansburg:
> Couple questions/feedback on this:
> 
> 1. I don't see the `ASSETS` binding in the `Your worker has access to the following bindings:` printout. That seems like a bug, and it's why it took me so long to find a solution. I was chasing ghosts.
> 2. What's your recommendation for getting the `fetch()` response body as bytes? I had to import `http_body_util::BodyExt` and write this, just to get basic data out:
>    
>    * ```rust
>      let bytes: Vec<u8> = response.into_body().collect().await?.to_bytes().to_vec();
>      ```

I think 1. may be a bug in `wrangler` cc @petebacondarwin 

For 2, I assume you are using the `http` crate feature? I think the main way to consume it is as a stream like you are doing, but I suppose we could add some utility methods to the concrete [`Body`](https://github.com/cloudflare/workers-rs/blob/636d76047fb41076de0f648ddb57c5f192bedbb0/worker/src/http/body.rs#L24) type you get from our `Fetcher::fetch`.

BrandonDyer64:
> we could add some utility methods to the concrete [`Body`](https://github.com/cloudflare/workers-rs/blob/636d76047fb41076de0f648ddb57c5f192bedbb0/worker/src/http/body.rs#L24) type you get from our `Fetcher::fetch`

I'd love to see an `async fn to_bytes() -> Vec<u8>` and `async fn to_json<T>() -> T` on `Body` if possible. It would make life a lot easier.

BrandonDyer64:
I'm not sure why the tests are failing in the CI. Started happening after my first commit, which just added a single method to `Env`.

BrandonDyer64:
Can anyone review this?

kflansburg:
I'm no longer with Cloudflare, so I cannot approve. I think test failures are related to the Rust 1.82 issue [here](https://github.com/cloudflare/workers-rs/issues/668). We may need to pin the Rust version in our CI. 

Otherwise, I think this is a good test to add. I would like to see it also tested with `http` flag enabled, we are trying to keep those two test suites identical and will probably eventually make `http` the default (or at least that was my original plan). 

cc @harrishancock

zebp (approved):
Sorry for the delay! It'd be nice to have this tested but otherwise LGTM, I'm going to merge into a non-main branch so I can add tests and update miniflare since I think the testing for this might be a bit more involved than the scope of this PR.

PR #657, opened 2024-10-18
Add cron worker example

pawurb:
This PR adds minimal example for using cron worker, together with logging. I could not find it anywhere. 

Screenshot 2024-10-18 at 12 02 53


examples/cron/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+use worker::*;
+
+#[event(scheduled)]
+async fn scheduled(_evt: ScheduledEvent, _env: Env, _ctx: ScheduleContext) {
+    wasm_rs_dbg::dbg!("Hello cron!");
```
zebp, line 5: Typically we just use the [`console_log`](https://docs.rs/worker/latest/worker/macro.console_log.html) macro to avoid adding an extra dependency.

examples/cron/wrangler.toml
```diff
@@ -0,0 +1,13 @@
+name = "cron-worker"
+main = "build/worker/shim.mjs"
+compatibility_date = "2023-03-22"
+
+[build]
+command = "cargo install worker-build && worker-build --release"
+
+[triggers]
+crons = ["*/5 * * * *"]
```
zebp, line 9: Would be nice to have a comment above to explain what the example cron does and link to the Cloudflare [cron docs](https://developers.cloudflare.com/workers/configuration/cron-triggers/).

zebp (changes requested):
2 small issues but otherwise lgtm

PR #670, opened 2024-11-17
fix: support for externref

### LINKED ISSUE #668: 🐛 BUG: Error with Rust 1.82.0 in Wrangler
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What version(s) of the tool(s) are you using?

wrangler 3.81.0, rustc 1.82.0

### What version of Node are you using?

20.10.0

### What operating system and version are you using?

Ubuntu 22.04

### Observed behavior
When using Rust version 1.82.0, running npx wrangler dev or npx wrangler deploy can result in an error related to WebAssembly imports. The error message may vary slightly but typically includes.
✘ [ERROR] service core:user:rs-worker: Uncaught LinkError: WebAssembly.Instance(): Import cloudflare/workers-sdk#2 "./index_bg.js" "__wbg_fetch_bc7c8e27076a5c84": function import requires a callable at null.<anonymous> (shim.js:41:9)
✘ [ERROR] The Workers runtime failed to start. There is likely additional logging output above.

When downgrading to Rust version 1.81.0, the problem disappears and everything works as expected.

### Expected behavior
The runtime should start successfully when using npx wrangler dev or npx wrangler deploy without producing errors in the console.

### Steps to reproduce
Please provide the following:
- [A minimal working subset of your worker code](https://github.com/cloudflare/workers-rs/tree/main/examples/router)
- Run your project with different versions of Rust (e.g., Rust 1.81.0 and Rust 1.82.0) to observe the differences in behavior.

### Please provide a link to a minimal reproduction

_No response_

### Please provide any relevant error logs

✘ [ERROR] service core:user:rs-worker: Uncaught LinkError: WebAssembly.Instance(): Import cloudflare/workers-sdk#2 "./index_bg.js" "__wbg_fetch_bc7c8e27076a5c84": function import requires a callable at null.<anonymous> (shim.js:41:9)
✘ [ERROR] The Workers runtime failed to start. There is likely additional logging output above.

zebp:
Fixes #668 

1. When reference types are enabled, which is the default in Rust 1.82, `wasm-bindgen` will generate JS where some wasm imports that were previously functions get generated as variables due to the difference in the JS/wasm interop barrier. After we bundled with esbuild we ran into the issue where a wasm import was defined after the instance was initialized, which wasn't a problem for the pre-reference type builds since functions get automatically hoisted in JS.

2. Previously we weren't calling `wasm-bindgen`'s `__wbindgen_start` function but instead called our `start` function for any function with `#[event(start)]`, but when reference types are enabled this function setups the externref table. This was causing us to change the value in the externref table for the `UNDEFINED` constant (index 128) to an http request causing our tests to fail. This was not fun to debug.

### DIFF
--- a/worker-build/src/js/glue.js
+++ b/worker-build/src/js/glue.js
@@ -1,5 +1,0 @@
-import wasmModule from "./index.wasm";
-import * as imports from "./index_bg.js";
-
-const instance = new WebAssembly.Instance(wasmModule, { "./index_bg.js": imports });
-export default instance.exports;
--- a/worker-build/src/js/shim.js
+++ b/worker-build/src/js/shim.js
@@ -10,44 +10,44 @@ const instance = new WebAssembly.Instance(wasmModule, {
 imports.__wbg_set_wasm(instance.exports);
 
 // Run the worker's initialization function.
-imports.start?.();
+instance.exports.__wbindgen_start?.();
 
 export { wasmModule };
 
 class Entrypoint extends WorkerEntrypoint {
-    async fetch(request) {
-        let response = imports.fetch(request, this.env, this.ctx);
-        $WAIT_UNTIL_RESPONSE
-        return await response;
-    }
+	async fetch(request) {
+		let response = imports.fetch(request, this.env, this.ctx);
+		$WAIT_UNTIL_RESPONSE;
+		return await response;
+	}
 
-    async queue(batch) {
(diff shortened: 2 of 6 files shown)


kflansburg (approved):
Nice!

zebp:
very very strange bug with reference types, it seems that we always get the first value passed to wasm with some options in r2 calls?

image

PR #671, opened 2024-11-19
chore: bump worker-build

zebp:
Bumps `worker-build` for a release since it doesn't seem to be covered by the existing release workflow, which i checked by running `cargo release version patch`.


zebp:
Manually tested this PR while pinning the Rust version to `1.81` due to #668 where everything worked as expected. Merging this in the meantime while I debug #670.

PR #673, opened 2024-11-24
docs: remove the leading dollar symbol that gets copied when pressing the "Copy" button

saikatdas0790:


guybedford (approved):

guybedford:
Agreed this is always annoying.

PR #675, opened 2024-11-28
Improved Durable Objects macro

### LINKED ISSUE #432: Using `#[durable_object]` more than once causes a compile error

0.0.18

3.22.3

### Describe the bug

Attempting to declare multiple `#[durable_object]` structs in the same module always results in a compile error.

Furthermore, there is [no documentation on the macro](https://docs.rs/worker/0.0.18/worker/attr.durable_object.html) so it's not clear how it is supposed to be used.

### Steps To Reproduce

Example code:
```rust
use worker::{durable_object, Env, Request, Response};

#[durable_object]
pub struct One {
    counter: u32,
}

#[durable_object]
impl DurableObject for One {
        Self {
            counter: 0,
        }
    }

    async fn fetch(&mut self, req: Request) -> worker::Result<Response> {
        Response::ok("one")
    }
}

#[durable_object]
pub struct Two {
    counter: u32,
}

#[durable_object]
impl DurableObject for Two {
        Self {
            counter: 0,
        }
    }

    async fn fetch(&mut self, req: Request) -> worker::Result<Response> {
        Response::ok("one")
    }
}

```

`wrangler dev` or `wrangler deploy` always results in a rust compiler error, because the macro apparently emits the same symbols each time it's invoked.

```txt
error[E0428]: the name `__Need_Durable_Object_Trait_Impl_With_durable_object_Attribute` is defined multiple times
  --> src/durable.rs:90:1
   |
72 | #[durable_object]
   | ----------------- previous definition of the trait `__Need_Durable_Object_Trait_Impl_With_durable_object_Attribute` here
...
90 | #[durable_object]
   | ^^^^^^^^^^^^^^^^^ `__Need_Durable_Object_Trait_Impl_With_durable_object_Attribute` redefined here
   |
   = note: `__Need_Durable_Object_Trait_Impl_With_durable_object_Attribute` must be defined only once in the type namespace of this module
   = note: this error originates in the attribute macro `durable_object` (in Nightly builds, run with -Z macro-backtrace for more info)

```

kanarus:
## Problem

Currently, this crate's support for Durable Objects *works*, but its developer experience can be better:

```rust
use worker::{durable_object, DurableObject};
// 1. We MUST introduce them into the scope for this code to compile.
use worker::{async_trait, wasm_bindgen};

#[durable_object]
struct MyObject {
    state: worker::State,
}

// 2. This attribute to impl-block prevents editor's completions of trait methods.
// 3. And this uses `async_trait` which is unnecessary, bothering layer now.
#[durable_object]
// 4. Due to current implementation of the attribute, hover and code jump on
// this ↓ `DurableObject` doesn't work.
impl DurableObject for MyObject {
    //...
}

// 5. Defining multiple Durable Objects causes compile error due to the wrong design of Durable Objects support
#[durable_object]
struct AnotherObject;

#[durable_object]
impl DurableObject for AnotherObject { /* ... */ }
``` 

<details> <summary>Details with Pictures</summary>

- 5 is reported in #432

- 1 makes error messages difficult to understand for those who avoid `use worker::*;`:
  - `worker` exports a lot of items from the top-level, so *-import causes serious namespace pollution and many developer will like to avoid it

Screenshot from 2024-11-28 16-19-53

- 3 leads to very unfriendly error messages for unexpected usage like:

Screenshot from 2024-11-28 16-07-45

- 2, 4: completion or hover doesn't work:

Screenshot from 2024-11-28 16-25-45

</details>

## Solution

This PR resolves all of them by re-designing Durable Objects support. See *worker-macros/src/durable_object.rs*  for how new `#[DurableObject]` works internally.

usage:

```rust
// 1. We only have to import `durable_object` and `DurableObject`.
use worker::{durable_object, DurableObject};

#[durable_object]
struct MyObject {
    state: worker::State,
}

// 2,3,4. Now we don't need annoying attribute on impl-block.
// Of course, hover and code jump and completion works perfectly.
impl DurableObject for MyObject {
    //...
}

// 5. This compiles. No problem.
#[DurableObject]
pub struct AnotherObject;

impl DurableObject for AnotherObject {
    //...
}
```

<details> <summary>Details with Pictures</summary>

- ( 1, 2, 3, 4 ) Of course, completion, hover and code jump work perfectly

Screenshot from 2024-11-28 16-02-37

Screenshot from 2024-11-28 16-12-55

Screenshot from 2024-11-28 16-39-13

- ( 5 ) More than one Durable Objects don't cause problem ( so closes #432 ).

Screenshot from 2024-11-28 16-11-47

- Very clear error message for unexpected usage: 

_if attribute is missing:_
Screenshot from 2024-11-28 16-01-16

_if trait impl is missing:_
Screenshot from 2024-11-28 16-01-58

</details>

~~Additionally, even if we merge this PR *immediately*, current codes still **compile and work**, just with "deprecated" warnings, due to some effort for backward compatibility:~~ Merge note: this was converted into a backwards incompatible change to avoid introducing non-idiomatic camel-casing convention.

Screenshot from 2024-11-28 16-44-14


adhami3310:
I also ran into the multiple durable objects problem, would be great if this can get merged o7

guybedford:
This is fantastic, thanks for working on it.

I've gone ahead and pushed up a rebase here and also update it to be a breaking change with a clear compile error if the `impl` attribute is not removed.

Further feedback welcome, otherwise I will aim to go ahead and get this one release with the 0.6 update.

guybedford (dismissed):

guybedford (approved):

PR #677, opened 2024-12-05
Support object variables in wrangler toml.

mendess:
Example:

```toml
[vars]
VAR = { a = 1, b = "hey" }

[vars.VAR]
a = 1
b = "hey"
```


worker/src/env.rs
```diff
@@ -42,12 +43,24 @@ impl Env {
...
+    /// See: <https://developers.cloudflare.com/workers/configuration/environment-variables/>
     pub fn var(&self, binding: &str) -> Result<Var> {
         self.get_binding::<Var>(binding)
     }
 
+    /// Get an environment variable defined in the [vars] section of your wrangler.toml that is
+    /// defined as an object.
+    ///
+    /// See: <https://developers.cloudflare.com/workers/configuration/environment-variables/>
+    pub fn object_var<T: DeserializeOwned>(&self, binding: &str) -> Result<T> {
```
zebp, line 58: Would be great to add a test for this. Can you add a var in [worker-sandbox's wrangler.toml](https://github.com/cloudflare/workers-rs/blob/main/worker-sandbox/wrangler.toml), add a route [here](https://github.com/cloudflare/workers-rs/blob/main/worker-sandbox/src/router.rs) that does a basic sanity check, and then write a basic vitest test like [this](https://github.com/cloudflare/workers-rs/blob/main/worker-sandbox/tests/r2.spec.ts).

zebp (changes requested):

zebp (approved):

Issue #678, opened 2024-12-08
[BUG] D1 binding doesn't support Null values


v-rogg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What version of `workers-rs` are you using?

0.4.2

### What version of `wrangler` are you using?

3.93.0

### Describe the bug

I have a DB with a nullable column. The following code works if the value is **not** `NULL`. As soon as the value is `NULL` I get this error:
`Error: Error: invalid type: JsValue(null), expected String - Cause: Error: invalid type: JsValue(null), expected String` 

```
#[derive(Deserialize, Serialize)]
struct ExampleObject {
    id: String,
    nullable: Option<String>,
}

#[event(fetch)]
async fn fetch(req: Request, env: Env, _ctx: Context) -> Result<Response> {
    Router::new()
        .get_async("/", |req, ctx| async move {
            let db = ctx.env.d1("DB")?;

            let query = db.prepare("SELECT id FROM example");
            let result = query.first::<ExampleObject>(None).await?;
            match result {
                Some(res) => Response::from_json(&res),
                None => Response::error("", 404),
            }
        })
        .run(req, env)
        .await
}
```

With a similar setup, trying to insert a `NULL` value I get this error:
`Error: Error: D1_TYPE_ERROR: Type 'object' not supported for value '[object ExecutionContext]' - Cause: Error: D1_TYPE_ERROR: Type 'object' not supported for value '[object ExecutionContext]'`

Currently a workout could be to use `.raw_js_values()` instead of `.first()`, `.all()`, `.raw()`

spigaz:
I have spent a few hours on the problem of persisting none values as nulls.

It seems that its a problem during serialization of the arguments, but in practice it doesn't seem so, the macro seems to be working correctly, converting none to wasm_bindgen::JsValue::NULL.

But the wasm_bindgen::JsValue::NULL instead of returning a valid null value is returning the environment variable with all the secrets and do on, yes this object:

```
D1_TYPE_ERROR: Type 'object' not supported for value '[object ExecutionContext]
```

just try
```rust
console_error!("Null value: {:?}", wasm_bindgen::JsValue::NULL);
```

I have no idea on how to fix this yet...

spigaz:
I detected the issue while testing serialisation for a none value, because the None was converted to that object instead of to null.

```rust
#[worker::send]
pub async fn serialize_none(_req: Request, _env: Env, _data: SomeSharedData) -> Result<Response> {
[...]

PR #680, opened 2024-12-10
D1 Binding - Null Support - Added tests to isolate the issue and fixed first to return None

spigaz:
As I mentioned on issue #678 the main problem seems to be with the JsValue::Null that for some reason, was somehow overwritten by the Env variable.

So on serialisation it serialises the env variable content, and it seems that on serialisation it still uses the same env variable to check if its a null or not.

In practice this means:
- When it tries to serialise None, it sends the full env as an object
- When it tries to deserialise, it doesn't recognise as null because its different from JsValue::NULL, and complains in the context of expecting a Some value.

When I try to reconstruct manually it deserialises just fine, because here the JsValue::NULL is the one referencing the Env and not JsValue(null).
```rust
let js_value = Object::new();
Reflect::set(&js_value, &JsValue::from_str("id"), &JsValue::from_f64(1.0)).unwrap();
Reflect::set(&js_value, &JsValue::from_str("name"), &JsValue::NULL).unwrap();
Reflect::set(&js_value, &JsValue::from_str("age"), &JsValue::NULL).unwrap();

let js_value: JsValue = js_value.into();
let value: NullablePerson = serde_wasm_bindgen::from_value(js_value).unwrap();
```

- [x] Added test cases for all of that, including for manual serialisation and deserialisation and even for JsNull.is_null() that fails, but I wasn't able to fix it yet.

@kflansburg Gave me a hint to check for different versions of wasm-bindgen in the hope of finding a regression.

I plan to investigate that, but it might not be that simple, I tested in my app running on wrangler and on wasm_bindgen_test and in both JsValue(null) is just fine, yes with the latest dependencies.

But the latest dependencies didn't helped in the workers-rs tests, although I don't use D1 on that app... And use web_sys...

Anyway anyone is free to give it a look, that's why I'm sharing the tests.

- [x] Added a fix and a test case for the first statement, that failed instead of returning None.


spigaz:
I was able to reproduce it just fine extracting the worker-sandbox and using it as an external project, I'm able to reproduce it just fine using either wrangler or vitest.

I wasn't having much success trying to trigger the issue in other apps, that use d1, axum or even http::Request.

Next step will be to try to isolate what triggers the issue in the sandbox.

guybedford:
It would be great to review this again now with a rebase if it is still relevant.

spigaz:
I made them about 6 months ago, to help trace down a problem someone was having with D1 and nulls.

It ended up being a corruption of the JsValue::NULL if I'm not mistaken, that was breaking the serialisation, surfacing as a D1 Null error.

As there were no tests either for none of the intermediate steps, I left them, but the location could be improved...

The tests are all running now, so do as you think its best...

I just had to run them locally with ```-- --test-timeout=10000```.
Because for some reason, some to take longer to start locally.

guybedford:
Thanks for explaining the background context here. I didn't realise it was already resolved, but since better coverage always seems worthwhile, I'll go ahead and land this anyway.

guybedford (approved):

PR #691, opened 2025-01-07
Add support for Analytics Engine

### LINKED ISSUE #280: [Feature] Workers Analytics Engine Support

### Description

Are there any plans to support Workers Analytics Engine in the near future?

### LINKED ISSUE #327: [Feature] Workers Analytics Engine

### Description

https://developers.cloudflare.com/analytics/analytics-engine/

cmackenzie1:
This one was a little tricky to add and requires mocking out the analytics engine binding using a miniflare wrapped binding that logs each data point to console.

Closes #280 
Closes #327


worker/src/analytics_engine.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub fn new() -> Self {
+        Self {
+            indexes: Vec::new(),
+            doubles: Vec::new(),
+            blobs: Vec::new(),
+            _phantom: PhantomData,
+        }
+    }
+
+    pub fn indexes(mut self, index: impl Into<Vec<String>>) -> Self {
```
zebp, line 96: For convenience, it'd be nice to have singular versions of all these methods that just added a single item to the inner vec.

worker/src/analytics_engine.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+// Marker traits to constrain T
+pub trait BlobType {}
+impl BlobType for String {}
+impl BlobType for Vec<u8> {}
+
+#[derive(Clone, Debug)]
+pub struct AnalyticsEngineDataPoint<T>
```
zebp, line 70: I think these should be type erased and under the hood the fields are JS arrays we can cheaply clone. This API as is feels a little painful always having to explicitly define a blob type that is only relevant when you're building.

worker/src/analytics_engine.rs
```diff
@@ -0,0 +1,186 @@
...
+where
+    T: BlobType,
+{
+    indexes: Vec<String>,
+    doubles: Vec<f64>,
+    blobs: Vec<T>,
+}
+
+pub struct AnalyticsEngineDataPointBuilder<T: BlobType> {
+    indexes: Vec<String>,
```
zebp, line 80: These should also just be JS Arrays we're settings/appending to.

worker/src/analytics_engine.rs
```diff
@@ -0,0 +1,186 @@
...
+        self.indexes = index.into();
+        self
+    }
+
+    pub fn doubles(mut self, doubles: impl Into<Vec<f64>>) -> Self {
+        self.doubles = doubles.into();
+        self
+    }
+
+    pub fn blobs(mut self, blobs: impl Into<Vec<T>>) -> Self {
```
zebp, line 106: If we make this `impl Into<Vec<impl BlobType>>` we can prevent the BlobType generic from leaking into the other types.

worker/src/analytics_engine.rs
```diff
@@ -0,0 +1,186 @@
...
+        self.blobs = blobs.into();
+        self
+    }
+
```
zebp, line 110: Also for convenience, a `write_to` method that accepts a `AnalyticsEngineDataset` so I can write it to WAE without having to then build and do a subsequent function call.

```rs
let ae = env.analytics_engine("whatever")?;
AnalyticsEngineDataPoint::builder()
    .doubles([1, 2, 3])
    .write_to(&ae)?;
```

worker/src/analytics_engine.rs
```diff
@@ -0,0 +1,186 @@
...
+{
+    indexes: Vec<String>,
+    doubles: Vec<f64>,
+    blobs: Vec<T>,
+}
+
+pub struct AnalyticsEngineDataPointBuilder<T: BlobType> {
+    indexes: Vec<String>,
+    doubles: Vec<f64>,
+    blobs: Vec<T>,
```
zebp, line 82: Notably the values in blobs can be an `Array<bytes | string>` not `Array<bytes> | Array<string>` so our builder methods can mix and match.

cmackenzie1:
@zebp since the type is `Array<bytes | string>` I simplified the types to use an `enum` instead, removing all the generic parameters. Added the convenience function `write_to` as well as `doubles`, `blobs` which set the entire array and then `add_blob` and `add_double` which append to the current array of values.

zebp (approved):
LGTM other than some parameter types that I think could be less strict, going to merge into another branch so I can loosen the type requirements

Issue #700, opened 2025-01-29
[BUG] Binding cannot be cast to the type RateLimiter from Ratelimit


maxdeviant:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What version of `workers-rs` are you using?

0.5.0

### What version of `wrangler` are you using?

3.106.0

### Describe the bug

When trying to retrieve a `RateLimiter` with `Env::get_binding` the following error occurs:

```
Binding cannot be cast to the type RateLimiter from Ratelimit
```

### Steps To Reproduce

I have the following in my `wrangler.toml`:

```toml
[[unsafe.bindings]]
name = "MY_RATE_LIMITER"
type = "ratelimit"
namespace_id = "1001"
simple = { limit = 10, period = 60}
```

And the rate limiter is being retrieved like so:

```rs
// ...

let rate_limiter = cx
    .env
    .get_binding::<RateLimiter>("MY_RATE_LIMITER")?;

let rate_limit_outcome = rate_limiter
    .limit(format!("user_id:{}", claims.user_id))
    .await?;
if !rate_limit_outcome.success {
    return Response::error(
        "Rate limit exceeded.",
        StatusCode::TOO_MANY_REQUESTS.as_u16(),
    );
}
```

Related GitHub Discussion: https://github.com/cloudflare/workers-rs/discussions/692

PR #701, opened 2025-01-29
Fix `RateLimiter` binding name

maxdeviant:
This PR fixes an issue where the `RateLimiter` could not be retrieved with `Env::get_binding`, as described in #700.

When trying to retrieve the rate limiter I was getting the following error:

```
Binding cannot be cast to the type RateLimiter from RateLimit
```

From what I can tell, the underlying issue is that the runtime constructor name is actually `Ratelimit`, based on this code here: https://github.com/cloudflare/workers-rs/blob/38af58acc4e54b29c73336c1720188f3c3e86cc4/worker/src/env.rs#L123

I was able to fix this by updating the type name from `RateLimiter` to `Ratelimit` in the `EnvBinding` impl.

If there is a different way this should be fixed, I am happy to pursue other avenues.

### DIFF
--- a/worker/src/rate_limit.rs
+++ b/worker/src/rate_limit.rs
@@ -20,7 +20,7 @@ unsafe impl Send for RateLimiter {}
 unsafe impl Sync for RateLimiter {}
 
 impl EnvBinding for RateLimiter {
-    const TYPE_NAME: &'static str = "RateLimiter";
+    const TYPE_NAME: &'static str = "Ratelimit";
 }
 impl RateLimiter {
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for sharing the PR. It would be great to see a test here and a rebase. I've opened https://github.com/cloudflare/workers-rs/pull/755 to follow-up on landing as well now.

guybedford:
Landed this one in https://github.com/cloudflare/workers-rs/pull/755.

maxdeviant:
> Landed this one in #755.

Thank you!

PR #708, opened 2025-02-15
Update params name for ScheduledTime::from

und3fined:
The term 'offset' refers to a generic name for parameters, which can make it challenging for code to understand.

With new code is super clear and easy to understand.


guybedford (dismissed):

worker/src/durable.rs
```diff
@@ -708,9 +708,9 @@ impl ScheduledTime {
 }
 
 impl From<i64> for ScheduledTime {
-    fn from(offset: i64) -> Self {
+    fn from(in_millis: i64) -> Self {
```
guybedford: This might actually be slightly better worded:

```suggestion
    fn from(offset_ms: i64) -> Self {
```

worker/src/durable.rs
```diff
@@ -708,9 +708,9 @@ impl ScheduledTime {
...
         ScheduledTime {
-            init: ScheduledTimeInit::Offset(offset as f64),
+            init: ScheduledTimeInit::Offset(in_millis as f64),
```
guybedford:
```suggestion
            init: ScheduledTimeInit::Offset(offset_ms as f64),
```

guybedford:
I ended up rolling this `offset_ms` change into the other PR I was working on. Do we still need the name change in this duration case as well?

und3fined:
@guybedford Okay, pls close this pull request, thanks

PR #715, opened 2025-03-15
feat(EmailWorkers): Implement email worker functionality

### LINKED ISSUE #274: [Feature] Add email worker support

nudded:
Closes #274 

Looking for initial comments so I can wrap up this implementation. I've successfully validated reply and forward functionality. 

example handler: 

```rust
use uuid::Uuid;
use worker::*;

#[event(email)]
async fn main(message: EmailMessage, _env: Env, _ctx: Context) -> Result<()> {
    let message_id = message.headers().get("Message-ID")?.unwrap();

    let msg = format!(
        "From: Cloudflare bot <{}>
To: Toon <{}>
In-Reply-To: {}
Message-ID: <{}@redacted.com>
Subject: Email well received!

I've parsed the mail!

",
        message.to_email(),
        message.from_email(),
        message_id,
        Uuid::new_v4()
    );

    let new_message = EmailMessage::new(&message.to_email(), &message.from_email(), &msg)?;

    message.reply(new_message).await?;
    Ok(())
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2086,6 +2086,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "receive-email-on-workers"
+version = "0.1.0"
+dependencies = [
+ "mail-builder",
+ "worker",
+]
+
+[[package]]
 name = "redox_syscall"
 version = "0.5.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/examples/receive-email/Cargo.toml
+++ b/examples/receive-email/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "receive-email-on-workers"
+version = "0.1.0"
+edition = "2021"
+
+[package.metadata.release]
+release = false
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+# Default feature `gethostname` pulls in a crate that doesn't build for
+# `wasm32-unknown-unknown`, so disable it. The remaining core API is enough
+# to assemble a message as long as we set `date` and `message_id` ourselves
+# (the auto-generated ones rely on `SystemTime::now()` / `gethostname`).
+mail-builder = { version = "0.4", default-features = false }
+worker.workspace = true
--- a/examples/receive-email/README.md
+++ b/examples/receive-email/README.md
@@ -1,0 +1,51 @@
+# Receiving email on Cloudflare Workers
+
+Example of an [Email Worker](https://developers.cloudflare.com/email-routing/email-workers/) that auto-replies with `message received` to every inbound message.
+
+The handler is wired up via `#[event(email)]`. It receives a [`worker::InboundEmail`](https://docs.rs/worker/latest/worker/struct.InboundEmail.html), pulls `Message-ID` and `Subject` off the original headers, assembles a reply with [`mail-builder`](https://crates.io/crates/mail-builder), and hands it to [`InboundEmail::reply`](https://docs.rs/worker/latest/worker/struct.InboundEmail.html#method.reply).
+
+Unlike `[[send_email]]`, inbound delivery is not declared in `wrangler.toml` — once deployed, attach an address to this worker from the **Email → Email Routing** section of the Cloudflare dashboard.
+
+## Local development
+
+`wrangler dev --local` exposes a `/cdn-cgi/handler/email` endpoint that simulates an inbound delivery. The reply is written to a local `.eml` file rather than actually being sent (see the [Cloudflare docs](https://developers.cloudflare.com/email-routing/email-workers/local-development/)).
+
+```bash
+npm install
+npm run dev
+# then, in another shell:
+curl -X POST 'http://localhost:8787/cdn-cgi/handler/email' \
+  --url-query 'from=sender@example.com' \
+  --url-query 'to=recipient@example.com' \
+  --header 'Content-Type: application/octet-stream' \
+  --data-binary @sample.eml
+```
+
+`sample.eml` is any RFC 5322 message, e.g.:
+
+```
+From: sender@example.com
+To: recipient@example.com
+Subject: hello
+Message-ID: <abc@example.com>
+
+hi there
+```
+
+## Deploying
+
+```bash
+npm run deploy
+```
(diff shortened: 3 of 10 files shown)


DougAnderson444:
Does this part need to be a specific format? Or can it be any string?

```rust
let msg = format!(
        "From: Cloudflare bot <{}>
To: Toon <{}>
In-Reply-To: {}
Message-ID: <{}@redacted.com>
Subject: Email well received!

I've parsed the mail!

"
```

What I am wondering is if it needs a specific format, maybe it should use the type system to enforce it?

If it's just any old string, then disregard

DougAnderson444:
What about an API like this:

```rust
use uuid::Uuid;
use worker::*;

#[event(email)]
async fn main(message: EmailMessage, _env: Env, _ctx: Context) -> Result<()> {

    let new_message = EmailMessage::try_from(
        RawEmailMessage::builder()
            .from_name("From Name")
            .from_email(&message.to_email())
            .to_name("To Name")
            .to_email(&message.from_email())
            .subject("Email well received!")
            .date("Sat, 15 Mar 2025 22:06:02 +0000")
            .message_id(format!("{}@redacted.com", Uuid::new_v4()))
            .in_reply_to(message.headers().get("Message-ID")?.unwrap())
            .message("I've parsed the mail!")
            .build(),
    );

    message.reply(new_message).await?;

    Ok(())
}
```

With the type system enforcing mandatory fields at compile time and removing any risk of user formatting errors?

DougAnderson444:
Actually, on second thought, maybe `msg` should just be left a a string and if the user wants to use a `build` helper, it can exist in userland as a separate crate.

DougAnderson444:
This looked promising: https://docs.rs/mail-builder/latest/mail_builder/index.html

nudded:
> This looked promising: https://docs.rs/mail-builder/latest/mail_builder/index.html

Tried this, got a CPU limit exceeded error on Cloudflare. (I think it's related to it adding the `Date` header, but did not want to spend much more time debugging)

jeholliday:
I tried out these changes, and I was able to receive and reply to emails just fine. I also used the mail_parser and mail_builder crates, and I found that they worked pretty well in combination with workers. The one thing I noticed is that you have to explicitly set a date and message id when building a message because otherwise the defaults will panic when it tries to get the date or generate a random id.

It would be nice if functionality for the [SendEmail binding](https://developers.cloudflare.com/email-routing/email-workers/send-email-workers/) could also be added. I tried adding it myself I was also able to send emails that way. I just reused the binding for an EmailMessage and it worked to call send().

devnull03:
> I tried out these changes, and I was able to receive and reply to emails just fine. I also used the mail_parser and mail_builder crates, and I found that they worked pretty well in combination with workers. The one thing I noticed is that you have to explicitly set a date and message id when building a message because otherwise the defaults will panic when it tries to get the date or generate a random id.
> 
> It would be nice if functionality for the [SendEmail binding](https://developers.cloudflare.com/email-routing/email-workers/send-email-workers/) could also be added. I tried adding it myself I was also able to send emails that way. I just reused the binding for an EmailMessage and it worked to call send().

@jeholliday would you be able to share what you wrote to implement the `SendEmail binding`? and the other stuff too. It would be really helpful. I am also trying to mainly use the send to other email addresses functionality instead of replying to received email.

jeholliday:
> @jeholliday would you be able to share what you wrote to implement the `SendEmail binding`? and the other stuff too. It would be really helpful. I am also trying to mainly use the send to other email addresses functionality instead of replying to received email.

@devnull03 Sure, sorry it took a few days. I have committed the changes needed to get the SendEmail binding working in https://github.com/jeholliday/workers-rs/commit/f05b750e46d63a4f71ff244d29234421dddbc953

This is an excerpt from where I am successfully using it:
```rust
use mail_builder::MessageBuilder;
use uuid::Uuid;
use worker::*;

async fn send_email(subject: &str, body: &str, env: &Env) -> Result<()> {
    let from = "<from_email>";
    let to = "<to_email>";

    let msg = MessageBuilder::new()
        .from(from)
        .to(to)
        .subject(subject)
        .date(Date::now().as_millis() / 1000)
        .message_id(format!("{}@<my_domain>", Uuid::new_v4()))
        .text_body(body)
        .write_to_string()
        .unwrap();

    let msg = EmailMessage::new(from, to, &msg)?;
    let seb = env.send_email("SEB")?;
    seb.send(msg).await?;

    Ok(())
}
```

This uses a binding named SEB from my `workers.toml`:
```toml
send_email = [
    { name = "SEB", destination_address = "<to_email>" },
]
```

relvacode:
I was able to get this to work with MessageBuilder for replying to incoming e-mails but one thing to note is that `In-Reply-To` must be set and you cannot use `.in_reply_to()` with the incoming message ID header directly, as MessageBuilder adds surrounding `<>`.

Instead, you need to set the header manually using a raw header value.

```rust
let msg_id = msg.headers().get("Message-ID")?.unwrap();
let reply_msg = MessageBuilder::new()
    .header("In-Reply-To", HeaderType::from(Raw::new(msg_id)))
```

guybedford (dismissed):
Thanks for the update on this one @nudded.

guybedford:
@nudded I simplified the wrapper struct to just implement the main two convenience methods you added as `raw_byte_stream()` and `raw_bytes`. PTAL and let me know how that seems to you.

nudded:
> @nudded I simplified the wrapper struct to just implement the main two convenience methods you added as `raw_byte_stream()` and `raw_bytes`. PTAL and let me know how that seems to you.

Looks good to me! I was adding an example as well, but I could add that in a follow up PR as well.

guybedford:
> I was adding an example as well, but I could add that in a follow up PR as well.

Example would be great, please go ahead - I can wait to land this.

nudded:
@guybedford added example and had to do a slight tweak to the macros. I deployed it and validated it (it's not the prettiest example, but it does work) 

(comments about local development can be removed, but perhaps useful for future readers)

guybedford (dismissed):

guybedford (approved):

PR #719, opened 2025-04-02
Add note to readme about runtime-agnostic primitives from tokio

lukevalenta:
I'd mistakenly assumed when first reading this warning that I couldn't use anything from `tokio` (even through the existing text does just say the tokio async runtime).  Add a note to make it clear that runtime-agnostic tokio crates are still usable.


irvinebroque (approved):
@danlapid @mikenomitch - came out of blog post conversation for dev week — think this is accurate

danlapid (approved):

danlapid:
I actually don't have permissions here
@zebp do you?

guybedford (approved):

PR #725, opened 2025-04-30
Update cargo-generate.toml for leptos template

loveJesus:
Fix to be able to run cargo generate for leptos which was throwing me the following error
Error: Substitution skipped, found invalid syntax in
        assets/favicon.ico
        public/favicon.ico

Hallelujah


guybedford:
Thanks for the PR, landed in https://github.com/cloudflare/workers-rs/pull/740.

PR #726, opened 2025-05-09
Added support for DO RPC

spigaz:
I'm not sure its the best approach but apparently it works, hopefully without regressions.

If I understood correctly, DOs required an extra shim to establish the connection to DurableObject from cloudflare:workers to get over this:

```
Error: TypeError: The receiving Durable Object does not support RPC, because its class was not declared with `extends DurableObject`. In order to enable RPC, make sure your class extends the special class `DurableObject`, which can be imported from the module "cloudflare:workers".
```

### Modular approach

Right now, it gets the DOs information from Wrangler.toml.

But it allows to add support easily to other Wrangler formats and other shims, including for service workers.

The DOs shim is only injected when DOs are detected, so no overhead on cold starts if not necessary.

I updated the DurableObject fetch method to be optional.

### Examples

I already added a couple of examples:
- One with a RPC Server and a RPC Client
- One with the non RPC version

I already had more examples in the pipeline, but the PR was getting too big.

### The server side usage is quite simple

But support is still limited to native types and client side remains as it was.

```rust
use wasm_bindgen::prelude::wasm_bindgen;
use worker::*;

#[durable_object]
pub struct RPCServer {
    counter: u32,
}

#[wasm_bindgen]
impl RPCServer {
    #[wasm_bindgen]
    pub fn add(&mut self, a: u32, b: u32) -> u32 {
        self.counter += 1;
        a + b + self.counter
    }
}

#[durable_object]
impl DurableObject for RPCServer {
    fn new(state: State, _env: Env) -> Self {
        Self { counter: 0 }
    }
}

```


worker-build/src/js/durable_objects_shim.js
```diff
@@ -0,0 +1,27 @@
+import { DurableObject } from "cloudflare:workers";
+
+const successfullyWrappedDONames = [];
+const globalStorePrefix = "__DO_WRAPPED_"; // For globalThis storage
+
+if (typeof __WORKER_BUILD_DO_NAMES__ !== 'undefined' && Array.isArray(__WORKER_BUILD_DO_NAMES__)) {
+    __WORKER_BUILD_DO_NAMES__.forEach(className => {
+        const OriginalClass = imports[className];
+        if (typeof OriginalClass === 'function' && OriginalClass.prototype) {
+            console.log(`[shim.js] Wrapping DO: ${className}`);
```
guybedford, line 10: Does it make sense to always log this?

guybedford (commented):
This looks great, sorry it's taken so long to review. I would very much like to land this soon though. I know the example gets run here, but it would help to have an explicit test as well for this behaviour including perhaps some of the edge cases and error cases.

worker-build/src/js/durable_objects_shim.js
spigaz: I just added, because it could be useful as a debug information in this particular stage, for people trying to make sure they were using the correct version, as that issue sometimes does happen.

I can remove it if you want...

spigaz:
> but it would help to have an explicit test as well for this behaviour including perhaps some of the edge cases and error cases.

I thought about that, even about adding more examples, but the PR was getting big and I wasn't even sure it was going to get approved.

Besides the tests weren't being valued properly.

But its okay, just let me know the test cases you want me to cover that I'll do my best.

worker-build/src/js/durable_objects_shim.js
guybedford: I think it's always nice to be able to turn these kinds of messages off. I'm still also familiarizing with the codebase - do we do messages like this in many other places? Perhaps it's okay just to get the warning messages when implementations are not found?

guybedford:
The failing test can be fixed with this change - https://github.com/cloudflare/workers-rs/pull/737/files#diff-dadc77a902a1d8dd154afd35c4c3dd86a6b287c5fd6c806f105973a1dd1ec7e8.

It looks like everything is passing except for the formatting otherwise.

In terms of increasing test coverage:

* Is it possible to move the simplest example into the `worker-sandbox` folder as a case there with tests against it. I think the examples should be more about readable code then.
* In terms of additional edge cases it would be nice to test more variations of the RPC shapes beyond just numbers. What happens when unsupported types are used? Testing known errors / invalid paths would also be great to see if possible.

worker-build/src/js/durable_objects_shim.js
spigaz: As I mentioned, I added it mainly because people trying it, could be running a different version of worker-build and having issues because of that.

It was an easy way for anyone having issues to trace back the problem.

Outside of wrangler, obviously it doesn't make any sense.

In the context of wrangler, to be honest, I feel that it outputs too much irrelevant information already.

But I do support your perspective, in general, there is no point for it, as once its working, I don't remember any reason to output it.

spigaz:
The idea of the sample was to get what Kevin did working on DOs as a starter, as I'm not sure how far he went with RPC support.

I'll give it a look and try to understand what the current boundaries are.

bandogora:
Will there be any movement on this in the near future? I'd like to use RPC for my DOs rather than creating a pseudo router in the `fetch` function.

PR #723, opened 2025-04-14
Add basic support for sqlite-backed durable objects

maxbrunsfeld:
Fixes https://github.com/cloudflare/workers-rs/issues/645

Hi, we want to use sqlite-backed durable objects using Rust, so we added some bindings to the Sql Storage API. Let us know if there are changes you'd like to see. Thanks.

PR #727, opened 2025-05-14
Add support for sqlite backed durable objects.

### LINKED ISSUE #645: Introduce SQLite API for Durable Objects

### Description

https://developers.cloudflare.com/durable-objects/api/storage-api/#sqlexec

connyay:
Fixes #645

### DIFF
--- a/worker-sandbox/src/sql_counter.rs
+++ b/worker-sandbox/src/sql_counter.rs
@@ -1,0 +1,59 @@
+use worker::*;
+use wasm_bindgen::JsValue;
+
+/// A simple SQLite-backed counter stored in Durable Object storage.
+///
+/// Each Durable Object instance owns its own private SQLite database.  We keep a
+/// single table `counter` with one row that stores the current value.
+#[durable_object]
+pub struct SqlCounter {
+    sql: SqlStorage,
+}
+
+#[durable_object]
+impl DurableObject for SqlCounter {
+    fn new(state: State, _env: Env) -> Self {
+        let sql = state.storage().sql();
+        // Create table if it does not exist.  Note: `exec` is synchronous.
+        sql.exec("CREATE TABLE IF NOT EXISTS counter(value INTEGER);", Vec::new())
+            .expect("create table");
+        Self { sql }
+    }
+
+    async fn fetch(&mut self, _req: Request) -> Result<Response> {
+        // Read current value (if any)
+        #[derive(serde::Deserialize)]
+        struct Row {
+            value: i64,
+        }
+
+        let rows: Vec<Row> = self
+            .sql
+            .exec("SELECT value FROM counter LIMIT 1;", Vec::new())?
+            .to_array()?;
+        let current = rows.get(0).map(|r| r.value).unwrap_or(0);
+        let next = current + 1;
+
+        // Upsert new value – simplest way: delete and insert again.
+        self.sql.exec("DELETE FROM counter;", Vec::new())?;
+        let val_js = JsValue::from_f64(next as f64);
+        self.sql
+            .exec("INSERT INTO counter(value) VALUES (?);", vec![val_js])?;
+
+        Response::ok(format!("SQL counter is now {}", next))
+    }
+}
+
+#[worker::send]
+/// Route handler that proxies a request to our SqlCounter Durable Object with id derived from the
+/// path `/sql-counter/{name}` (so every name gets its own instance).
+pub async fn handle_sql_counter(req: Request, env: Env, _data: super::SomeSharedData) -> Result<Response> {
+    let uri = req.url()?;
+    let mut segments = uri.path_segments().unwrap();
+    // skip "sql-counter"
+    let _ = segments.next();
+    let name = segments.next().unwrap_or("default");
+    let namespace = env.durable_object("SQL_COUNTER")?;
+    let stub = namespace.id_from_name(name)?.get_stub()?;
+    stub.fetch_with_str("https://fake-host/").await
+}
--- a/worker/src/sql.rs
+++ b/worker/src/sql.rs
@@ -1,0 +1,102 @@
+use js_sys::{Array};
+use wasm_bindgen::JsValue;
+use worker_sys::types::{SqlStorage as SqlStorageSys, SqlStorageCursor as SqlStorageCursorSys};
+
+use serde::de::DeserializeOwned;
+use serde_wasm_bindgen as swb;
+
+use crate::Error;
+use crate::Result;
+
+/// Wrapper around the Workers `SqlStorage` interface exposed at `ctx.storage.sql`.
+///
+/// The API is intentionally minimal for now – additional helper utilities can be built on top
+/// as needed.
+#[derive(Clone)]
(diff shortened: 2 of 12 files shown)


connyay:
alt to https://github.com/cloudflare/workers-rs/pull/723 with some additional bindgen methods, a test, and a sandbox example.

~This PR uses jsvalue pretty liberally and omits SqlStorageValue. Happy to make a change if needed, but types are already wonky in sqlite and the rust/wasm layers make it even more confusing. Would love from someone from cloudflare to weigh in there. Ideally could skip things like https://github.com/cloudflare/workers-rs/pull/723/files#r2045394966 if we are just ysing jsvalues? 🤷‍♂️~

connyay:
Ended up adding SqlStorageValue after attempting to use my fork. Definitely improves ergonomics.

worker-sandbox/src/sql_counter.rs
```diff
@@ -0,0 +1,61 @@
...
+impl DurableObject for SqlCounter {
+    fn new(state: State, _env: Env) -> Self {
+        let sql = state.storage().sql();
+        // Create table if it does not exist.  Note: `exec` is synchronous.
+        sql.exec("CREATE TABLE IF NOT EXISTS counter(value INTEGER);", None)
+            .expect("create table");
+        Self { sql }
+    }
+
+    async fn fetch(&mut self, _req: Request) -> Result<Response> {
```
guybedford: This is now:

```suggestion
    async fn fetch(&self, _req: Request) -> Result<Response> {
```

guybedford (dismissed):
This is great, thank you.

worker-sandbox/src/sql_counter.rs
guybedford: corrected.

worker/src/sql.rs
```diff
@@ -0,0 +1,216 @@
...
+/// A cursor returned from `SqlStorage::exec`.
+#[derive(Clone, Debug)]
+pub struct SqlCursor {
+    inner: SqlStorageCursorSys,
+}
+
+unsafe impl Send for SqlCursor {}
+unsafe impl Sync for SqlCursor {}
+
+impl SqlCursor {
```
guybedford, line 314: Seems this is missing:
* Iterator
* resusedCachedQueryForTest
* raw
* next

Any reason not to consider adding any of these as well?

worker/src/sql.rs
```diff
@@ -0,0 +1,216 @@
...
+/// as needed.
+#[derive(Clone, Debug)]
+pub struct SqlStorage {
+    inner: SqlStorageSys,
+}
+
+unsafe impl Send for SqlStorage {}
+unsafe impl Sync for SqlStorage {}
+
+impl SqlStorage {
```
guybedford, line 175: Seems this one is missing:

* prepare
* ingest
* setMaxPageCountForTest

Again, any reason to exclude these? It might just be worth investigating implementation / leaving a comment that these are pending implementation otherwise.

connyay: got a lil spooked by https://github.com/cloudflare/workerd/pull/2722 https://github.com/cloudflare/workerd/blob/3d177d3/src/workerd/api/sql.h#L43-L52

left a TODO and reference to the sql.h. let me know how you'd like to proceed

worker/src/sql.rs
connyay: added raw,next, and Iterator impl. not 100% sure I got those right. 

left a todo and ref for resusedCachedQueryForTest

worker/src/sql.rs
```diff
@@ -213,4 +213,43 @@ impl SqlCursor {
...
+    /// to be used as a JavaScript iterator.
+    pub fn next(&self) -> js_sys::Object {
+        self.inner.next()
+    }
+
+    /// Returns an iterator where each row is an array rather than an object.
+    ///
+    /// This method provides a more efficient way to iterate over results when you
+    /// only need the raw values without column names.
+    pub fn raw(&self) -> js_sys::Iterator {
```
guybedford: Rather than having these return `js_sys` types I think it would be good to have these written as first-class Rust APIs, so `next()` with a signature more like the Rust iterator case returning `Iterator<Item = T>` and `raw()` returning `Iterator<Item = Vec<SqlStorageValue>>` I believe.

guybedford (commented):
Thanks so much for following up on this. Marking the TODOs for experimental features makes sense to me. I'd be happy to keep working this one through if we can get it in for the coming release.

README.md
```diff
@@ -271,6 +271,67 @@ tag = "v1" # Should be unique for each entry
...
+impl DurableObject for SqlCounter {
+    fn new(state: State, _env: Env) -> Self {
+        let sql = state.storage().sql();
+        // Create table if it does not exist
+        sql.exec("CREATE TABLE IF NOT EXISTS counter(value INTEGER);", None)
+            .expect("create table");
+        Self { sql }
+    }
+
+    async fn fetch(&mut self, _req: Request) -> Result<Response> {
```
guybedford:
```suggestion
    async fn fetch(&self, _req: Request) -> Result<Response> {
```

connyay:
sry. I was only iterating on `npm run test` 👼

worker/src/sql.rs
```diff
@@ -0,0 +1,365 @@
...
+/// This enum represents the types that can be safely passed to SQL queries
+/// while maintaining type safety and proper conversion to JavaScript values.
+#[derive(Debug, Clone, PartialEq)]
+pub enum SqlStorageValue {
+    /// SQL NULL value
+    Null,
+    /// Boolean value
+    Boolean(bool),
+    /// 32-bit signed integer (safe for JavaScript Number conversion)
+    Integer(i32),
```
guybedford: Thinking about this one again, I think it would make more sense to use `i64` and then have a guard when setting to determine that it is between `Number.MIN_SAFE_INTEGER` and `Number.MAX_SAFE_INTEGER` for JS. That gets us over 6 byte integers at least, which would be a big improvement over 4 byte integers.

connyay: hobbled something together. feels a bit messy/dangerous

+1 on https://github.com/cloudflare/workerd/issues/4195

guybedford: Yeah I see what you mean about this. I'd quite like to (a) know what the upgrade path to bigint would look like and be sure this isn't now inhibiting that and (b) investigate the bigint approach more.

I guess for the bigint upgrade we would want `Integer(i32)` along with `BigInteger(i64)`?

guybedford: Or, would we just have `Integer(i64)` where the JS API semantics are that:

* Conversion to JS will use BigInt when outside i32 range
* Conversion to JS will always use BigInt if that's supported safely
* Conversion from JS would always support either

I guess thinking this through some more the above seems absolutely fine to me from an interop perspective without needing any other special handling, so that `Integer(i64)` from the start is the correct end design, since it would be more painful on the bigint upgrade path to have an opt-in variant based on range.

guybedford:
Thanks again for your work on this!

PR #728, opened 2025-05-16
feat: Add support for DO websocket_auto_response.

connyay:
Noticed this was missing from the rust apis.
Basic bindgen, miniflare test, and sandbox example.


worker-sandbox/src/test/auto_response.rs
```diff
@@ -0,0 +1,49 @@
...
+impl DurableObject for AutoResponseObject {
+    fn new(state: State, _env: Env) -> Self {
+        Self { state }
+    }
+
+    async fn fetch(&mut self, req: Request) -> Result<Response> {
+        match req.path().as_str() {
+            "/set" => {
+                // Configure ping -> pong auto-response for all websockets bound to this DO.
+                let pair = worker_sys::WebSocketRequestResponsePair::new("ping", "pong")?;
```
guybedford: Rather than requiring the `worker_sys` usage explicitly here (it's not otherwise a dependency), let's instead expose `WebSocketRequestResponsePair` through `worker::` directly.

connyay: let me know if this matches what you had in mind https://github.com/cloudflare/workers-rs/pull/728/commits/865a5a360df9afa24748fece3f602471bcb0a922 👀

guybedford (approved):

PR #737, opened 2025-05-29
wip: shared durable object approach

guybedford:
This provides a possible implementation for a `SharedDurableObject` trait which does not implement the `&mut self` pattern but instead implements `&self` to allow multiple concurrent borrowers.

This then pushes the concurrency problem for durable object access down into field cells, where any standard Rust techniques can then be used to manage mutable field access - `Mutex` / `RwLock` / `RefCell` etc.

The test here uses the `RefCell` approach for mutable fields, where so long as mutable borrows are not held over async points there will never be a panic.

While formally we are single threaded, blanket turning off concurrency checks on mutable access may have implications for undefined Rust compiler behaviours. So instead this approach aligns with Rust conventions, even though threaded code is not in play.

I think these idioms should be perfectly comfortable to Rust devs (far more than relaxing mutable behaviours), even if they don't fully apply in this single threaded JS embedding environment it remains closer to the proper way of doing things without much ergonomic cost.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -952,6 +952,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
 
 [[package]]
+name = "gloo-timers"
+version = "0.3.0"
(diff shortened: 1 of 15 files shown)


guybedford:
Superseded by https://github.com/cloudflare/workers-rs/pull/753.

PR #761, opened 2025-06-20
Add name method to ObjectId, fixes #760

### LINKED ISSUE #760: [Feature] Make accessing Durable Object name more ergonomic

### Description

Durable Object IDs created with https://developers.cloudflare.com/durable-objects/api/namespace/#idfromname have a [name](https://developers.cloudflare.com/durable-objects/api/id/#name) property, which can be useful to be able to access within the Durable Object's `new()` function, for example for logging or to retrieve a specific configuration based on the name (e.g., https://github.com/cloudflare/azul/blob/56720294b68225bce8fba0273178639080fcba8f/crates/ct_worker/src/sequencer_do.rs#L24).

This can be done using a function like this:
```
/// Retrieve the
/// [name](https://developers.cloudflare.com/durable-objects/api/id/#name) that
/// was used to create a Durable Object Id with `id_from_name`. The signature of
/// this function is a little funny since the only way to access the `State`'s
/// inner `DurableObjectState` is via the `_inner()` method which takes
/// ownership of the state. Thus, we just re-derive the State from the inner
/// state and return it in case the calling function still needs it.
///
/// # Errors
///
/// Returns an error if the 'name' property is not present, for example if the
/// object was created with a random ID.
pub fn get_durable_object_name(state: State) -> Result<(State, String)> {
    let inner_state = state._inner();
    let id = inner_state.id()?;
    let obj = js_sys::Object::from(id);
    let name = js_sys::Reflect::get(&obj, &"name".into())?
        .as_string()
        .unwrap_or_default();
    Ok((State::from(inner_state), name))
}
```

The ask in this ticket is to either make this more ergonomic (e.g., add a `fn name(&self) -> Option<String>` method to the `State` struct, or just to document the above. Happy to file a PR if this seems useful to others.

lukevalenta:
- Fix up worker-sandbox/wrangler.toml to get 'npx wrangler dev' to work (on macOS)
- Add SOME_SECRET to .dev.vars to fix /durable/:id endpoint
- Add handlers for /hello and /storage endpoints in export_durable_object.rs in order to test ObjectId::name()

Happy to add more tests, etc. if you point me in the right direction. Running `npx wrangler dev` from worker-sandbox (with the fixes from this PR), I get the following outputs as expected:

```
% curl http://localhost:8787/durable/hello-unique
Hello from 85da127a4d082a143c886f91dd7f67b11bd831ce91dfa83d57feca3917a615f0!
% curl http://localhost:8787/durable/hello
Hello from your Durable Object!
```

### DIFF
--- a/worker/src/env.rs
+++ b/worker/src/env.rs
@@ -138,6 +138,7 @@ pub trait EnvBinding: Sized + JsCast {
     }
 }
 
+#[repr(transparent)]
 pub struct StringBinding(JsValue);
 
 impl EnvBinding for StringBinding {
@@ -154,6 +155,7 @@ impl JsCast for StringBinding {
     }
 
     fn unchecked_from_js_ref(val: &JsValue) -> &Self {
+        // Safety: Self is marked repr(transparent)
         unsafe { &*(val as *const JsValue as *const Self) }
     }
 }
@@ -199,7 +201,8 @@ impl JsCast for JsValueWrapper {
     }
 
     fn unchecked_from_js_ref(val: &JsValue) -> &Self {
-        unsafe { std::mem::transmute(val) }
+        // Safety: Self is marked repr(transparent)
+        unsafe { &*(val as *const JsValue as *const Self) }
     }
 }
 


worker-sandbox/src/test/durable.rs
```diff
@@ -5,6 +5,11 @@ use worker::*;
 pub async fn basic_test(env: &Env) -> Result<()> {
```
lukevalenta, line 5: I'm not sure how this actually gets called in tests (if at all). I can add another handler to the sandbox worker if that seems useful.

worker-sandbox/src/router.rs
```diff
@@ -452,6 +452,18 @@ pub fn make_router<'a>(data: SomeSharedData) -> Router<'a, SomeSharedData> {
             "/durable/auto-response",
             handler!(crate::test::auto_response::handle_auto_response),
         )
+        .get_async(
```
lukevalenta, line 455: I realized I didn't add the corresponding routes when the `http` feature is enabled. Addressed in https://github.com/cloudflare/workers-rs/pull/763.

worker-sandbox/src/test/durable.rs
lukevalenta: Addressed in https://github.com/cloudflare/workers-rs/pull/763.

worker/src/durable.rs
```diff
@@ -176,6 +176,15 @@ impl ObjectId<'_> {
             })
             .map_err(Error::from)
     }
+
+    /// The name that was used to create the `ObjectId` via [`id_from_name`](https://developers.cloudflare.com/durable-objects/api/namespace/#idfromname).
```
guybedford, line 180: Should we also implement `equals(otherId)` as well here?

guybedford (dismissed):

worker/src/durable.rs
```diff
@@ -176,6 +176,15 @@ impl ObjectId<'_> {
...
+    /// `None` is returned if the `ObjectId` was constructed using [`unique_id`](https://developers.cloudflare.com/durable-objects/api/namespace/#newuniqueid).
+    pub fn name(&self) -> Option<String> {
```
lukevalenta, line 182: Ooh looks like we can use a `getter` attribute instead: https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-js-imports/getter-and-setter.html. Just tested it and will push an update shortly.

worker/src/durable.rs
lukevalenta: Hmm good question. I don't need it for my particular use case, at least. If I naively derive `PartialEq` and `Debug` on ObjectId and all the structs it depends on, and then add this code within a Durable Object's `new()` function
```
        let namespace = env.durable_object("MY_CLASS").unwrap();
        let id = namespace.id_from_name("your Durable Object").unwrap();
        assert_eq!(id, state.id(), "Object IDs don't match");
```
then it fails with
```
  assertion `left == right` failed: Object IDs don't match
    left: ObjectId { inner: DurableObjectId { obj: Object { obj: JsValue(Object({"name":"your
  Durable Object"})) } }, namespace: Some(ObjectNamespace { inner: DurableObjectNamespace { obj:
  Object { obj: JsValue(Object({})) } } }) }
   right: ObjectId { inner: DurableObjectId { obj: Object { obj: JsValue(Object({"name":"your
  Durable Object"})) } }, namespace: None }
  ```

This is because the `state.id()` passed into the Durable Object's `new()` function has `namespace: None`. Looking at `ObjectId::get_stub()`, it looks like this is intentional to discourage having a DO calling `get_stub` on itself.

All that to say, I'm not sure what the expected behavior for an `equals` function should be given the optional namespace.

lukevalenta: Ah, I see there is already an `equals` that we can pull in from JS: https://developers.cloudflare.com/durable-objects/api/id/#equals. I'll add that in!

lukevalenta: Done! The equality check ignores the optional namespace field of the ObjectId, but still returns false when comparing two ObjectIds with the same name but from different namespaces, which is the behavior I'd expect.

guybedford (approved):

PR #762, opened 2025-06-21
Fix soundness issue in `StringBinding::unchecked_from_js_ref`

Vrtgs:
this may be common, or may not, future refcasting should probably use the [ref-cast](https://docs.rs/ref-cast/latest/ref_cast/) crate

### DIFF
--- a/worker/src/env.rs
+++ b/worker/src/env.rs
@@ -138,6 +138,7 @@ pub trait EnvBinding: Sized + JsCast {
     }
 }
 
+#[repr(transparent)]
 pub struct StringBinding(JsValue);
 
 impl EnvBinding for StringBinding {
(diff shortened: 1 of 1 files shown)


worker/src/env.rs
```diff
@@ -154,6 +155,7 @@ impl JsCast for StringBinding {
     }
 
     fn unchecked_from_js_ref(val: &JsValue) -> &Self {
+        // Safety: Self is marked repr(transparent)
         unsafe { &*(val as *const JsValue as *const Self) }
```
guybedford, line 159: Would this allow us to use transmute?

vrtgs: transmute works slightly worse with tools like miri, and this is equivalent to the generated code just more clear, the reason its safe due to `repr(transparent)` is because the layout of `repr(Rust)` is unspecified, yeah sure most of the time it works but its not guaranteed

Issue #207, opened 2022-07-28
Update matchit dependency


nuvacore:
Can you please update matchit dependency to 0.6 and release a new version of worker?

lukevalenta:
Looks like the current version in use is matchit=0.7 (after #450). However, we should update to matchit=0.8.6 (https://github.com/ibraheemdev/matchit/releases), which will align the parameters matching more closely with axum::Router.

PR #763, opened 2025-06-22
Clean up worker-sandbox

lukevalenta:
- Address or add allow rules for clippy::pedantic lints
- Add macros to avoid duplicate route configurations
- Move files from src/test/* to src/ and add test handlers and routes to
  make it possible to run the tests from the sandbox worker
- Move several handlers out of alarm.rs and into more suitable files
  (e.g., counter.rs where the corresponding Durable Objects live)
- Fix and add test handler for handle_basic_test. The test now passes
  except for the transactional storage API tests.

Depends on https://github.com/cloudflare/workers-rs/pull/761. I tested several (but not all) worker-sandbox routes both with the `http` feature enabled and disabled.

If we update the matchit dependency (https://github.com/cloudflare/workers-rs/issues/207) we could remove some of the new macros this PR adds, but that's a breaking change for how routes are defined.

### DIFF
--- a/examples/custom-req/src/lib.rs
+++ b/examples/custom-req/src/lib.rs
@@ -34,6 +34,6 @@ impl IntoResponse for MyResponse {
         impl Into<Box<dyn std::error::Error>>,
     > {
         crate::worker_sys::web_sys::Response::new_with_opt_str(Some(self.data))
-            .map_err(|e| Error::JsError(format!("{:?}", e)))
+            .map_err(|e| Error::JsError(format!("{e:?}")))
     }
 }
--- a/examples/rpc-client/build.rs
+++ b/examples/rpc-client/build.rs
@@ -7,7 +7,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("cargo::rerun-if-changed=wit/calculator.wit");
     let source = expand_wit_source("wit/calculator.wit")?;
     let out_dir = var("OUT_DIR")?;
-    let dest = format!("{}/calculator.rs", out_dir);
+    let dest = format!("{out_dir}/calculator.rs");
     std::fs::write(dest, source)?;
     Ok(())
 }
--- a/examples/rpc-client/src/lib.rs
+++ b/examples/rpc-client/src/lib.rs
@@ -14,5 +14,5 @@ async fn main(_req: Request, env: Env, _ctx: Context) -> Result<Response> {
 
     let num = service.add(1, 2).await?;
 
-    Response::ok(format!("{:?}", num))
+    Response::ok(format!("{num:?}"))
 }
--- a/worker-codegen/src/wit.rs
+++ b/worker-codegen/src/wit.rs
@@ -29,7 +29,7 @@ fn wit_type_to_str(ty: &wit_parser::Type) -> anyhow::Result<String> {
         wit_parser::Type::F64 => "f64".to_string(),
         wit_parser::Type::Char => "char".to_string(),
         wit_parser::Type::String => "String".to_string(),
-        wit_parser::Type::Id(t) => anyhow::bail!("Unsupported type: '{:?}'", t),
+        wit_parser::Type::Id(t) => anyhow::bail!("Unsupported type: '{t:?}'"),
     })
 }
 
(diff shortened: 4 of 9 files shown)


lukevalenta:
Looks like the `worker-sandbox/src/test/` directory is used after all (via `npm run test`). I'll revert the changes that broke the tests.

EDIT: The tests broke because of enabling debug symbols, not because we got rid of `worker-sandbox/src/test`.

lukevalenta:
Setting `dwarf-debug-info = false` at https://github.com/cloudflare/workers-rs/pull/763/files#diff-6743bc1f4fe1302279b625c1102a5e5a7992ad23c75941db65707d61918a5e6eR16 fixed the tests (since tests are run with the `dev` profile: https://github.com/cloudflare/workers-rs/blob/main/package.json#L26). This is now ready for review!

cc @guybedford

worker-sandbox/src/alarm.rs
```diff
@@ -1,6 +1,5 @@
 use std::time::Duration;
-use tokio_stream::{StreamExt, StreamMap};
-
+#[allow(clippy::wildcard_imports)]
 use worker::*;
```
guybedford: I'd prefer to either make this explicit or not have the clippy attribute.

worker-sandbox/src/cache.rs
```diff
@@ -7,7 +7,7 @@ use worker::{console_log, Cache, Date, Delay, Env, Request, Response, ResponseBu
 fn key(req: &Request) -> Result<Option<String>> {
     let uri = req.url()?;
     let mut segments = uri.path_segments().unwrap();
-    Ok(segments.nth(2).map(|s| s.to_owned()))
+    Ok(segments.nth(2).map(std::borrow::ToOwned::to_owned))
```
guybedford: Would be nicer to use an include of `std::borrow::ToOwned`.

worker-sandbox/wrangler.toml
```diff
@@ -75,9 +71,6 @@ preview_bucket_name = 'delete-bucket'
 [build]
 command = "worker-build --release"
 
-[package.metadata.wasm-pack.profile.dev.wasm-bindgen]
-dwarf-debug-info = true
```
guybedford, line 79: It would be really nice to know the reason for this. It doesn't seem to be a timeout. Any ideas?

guybedford (approved):
Thanks for working on this, would be great to get it landed.

worker-sandbox/wrangler.toml
lukevalenta: I managed to get the tests to pass with `dwarf-debug-info = true` when I run `vitest` with `--test-timeout=0 --no-file-parallelism` (i.e.., update the relevant scripts in package.json to the below):
```
    "test": "cd worker-sandbox && NO_MINIFY=1 worker-build --dev && NODE_OPTIONS=--experimental-vm-modules npx vitest run --test-timeout=0 --no-file-parallelism",
    "test-http": "cd worker-sandbox && NO_MINIFY=1 worker-build --dev --features http && NODE_OPTIONS=--experimental-vm-modules npx vitest run --test-timeout=0 --no-file-parallelism"
  ```

Without `--no-file-parallelism` I consistently get errors like this, but I'm not sure the root cause:
```
| ⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯
| 
| Vitest caught 1 unhandled error during the test run.
| This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.
| 
| ⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
| Error: Worker exited unexpectedly
|  ❯ ChildProcess.onUnexpectedExit ../node_modules/tinypool/dist/index.js:118:30
|  ❯ ChildProcess.emit node:events:529:35
|  ❯ ChildProcess._handle.onexit node:internal/child_process:292:12
```

worker-sandbox/src/alarm.rs
lukevalenta: Done! Also cleaned up this lint in other places I'd added it.

lukevalenta:
> Thanks for working on this, would be great to get it landed.

Thanks @guybedford, this is ready for another round of review.

guybedford (approved):

guybedford:
Really appreciated for the follow-up. Great to see these improvements.

Issue #408, opened 2023-11-07
[BUG] JS Snippets not working anymore.


MathieuTricoire:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What version of `workers-rs` are you using?

0.0.18

### What version of `wrangler` are you using?

3.15.0

### Describe the bug

Hi everyone,

I've recently started working with Cloudflare and Cloudflare Workers, and I've encountered what appears to be a regression with "JS Snippets". Upon reviewing the repository's history, it seems that this feature was previously functioning correctly, but I suspect changes, potentially around the "glue" code, might have affected this functionality.

Error when running `npm run dev`
<img width="1029" alt="Screenshot 2023-11-07 at 4 03 14 PM" src="https://github.com/cloudflare/workers-rs/assets/255766/a3272c44-a531-4ab3-8f63-af5946dd120d">

WebAssembly code:
<img width="1019" alt="Screenshot 2023-11-07 at 4 03 42 PM" src="https://github.com/cloudflare/workers-rs/assets/255766/228ee45e-8a90-4153-bd37-5ced8104a13a">

Shim.mjs code (glue code):
<img width="869" alt="Screenshot 2023-11-07 at 4 04 03 PM" src="https://github.com/cloudflare/workers-rs/assets/255766/7730f6ef-23b8-4f5a-8b61-d2daeeaf9d5d">

Let me know if I miss something or if this is a real issue.

Thanks,
Mat

### Steps To Reproduce

Use my example: https://github.com/MathieuTricoire/workers-rs/tree/fix-js-snippets/examples/js_snippets

If you clone my repo and want to be sure to reproduce the error do not forget to replace the `wrangler.toml` build command with something like this, since my repo contains a fix:

```
[build]
command = "worker-build --release"
```

MathieuTricoire:
FYI: I have a potential fix on my fork, let me know if you want me to create a PR.

kflansburg:
I'm not familiar with this mechanism, is this part of wasm-bindgen that we need to support in our custom glue code? I know that we treat WebAssembly imports different than other runtimes which may be the issue here. Are you using Miniflare 2 or 3?

MathieuTricoire:
Yes it's in wasm-bindgen: https://rustwasm.github.io/wasm-bindgen/reference/js-snippets.html and it was previously supported see this PR #142.

I'm using Miniflare 3.

PR #410, opened 2023-11-09
fix js snippets + example

kflansburg:
Closes #408 

This seems reasonable, but I don't know the history of this feature or why it stopped working. What do you think @zebp ?

PR #768, opened 2025-07-07
Fix wasm-bindgen JS Snippets

guybedford:
Reintroduces https://github.com/cloudflare/workers-rs/pull/410 adding support for Wasm Bindgen snippets (https://rustwasm.github.io/wasm-bindgen/reference/js-snippets.html) with a test in the test suite.

Resolves https://github.com/cloudflare/workers-rs/issues/408.


worker-sandbox/src/js_snippets.rs
hoodmane, line 1: Is worker-sandbox a test? I see there is a README in here but it is empty.

guybedford: Yeah its the test suite.

worker-sandbox/src/js_snippets.rs
```diff
@@ -0,0 +1,27 @@
+use wasm_bindgen::prelude::*;
+use worker::*;
+
+use crate::SomeSharedData;
+
+#[wasm_bindgen(inline_js = "export function js_performance_now() { return performance.now(); }")]
```
hoodmane, line 6: So if I'm understanding here, wasm-bindgen places this into `snippets/some_snippet.mjs` and then `js_performance_now` is declared to come from `imports["snippets/some_snippet.mjs"]["js_performance_now"]`? I think some comments in the added code in `worker-build/src/main.rs` explaining who put what where and what they expect us to do would be helpful. e.g., 
```
// This is for handling `inline_js`. Each inline_js file goes in a separate file
// in snippets called __. We're supposed to put the function defined there
// into an import with module ___ and name ___
```
also, what does the directory structure of snippets look like?

worker-sandbox/src/js_snippets.rs
hoodmane: Could you add to the readme.md in the folder saying that this is a test folder? Or better, rename the folder `test-worker` or something?

guybedford: Yup that is very much needed, will follow-up in a subsequent PR as I do these various refactorings of the codebase.

worker-sandbox/src/js_snippets.rs
guybedford: Yes, this is a wasm-bindgen output convention. Added a comment about the structure.

hoodmane (approved):

PR #771, opened 2025-07-09
Clarify that id.name is not available within DO

lukevalenta:
See https://github.com/cloudflare/workerd/pull/4351 for details, but id.name no longer works in local testing (and never worked in production). Update tests to reflect this, and demonstrate how the DO name can be passed as a query parameter instead.


guybedford (approved):
Thanks for following up on this.

PR #776, opened 2025-07-14
Make `as_organization` optional on `IncomingRequestCfProperties`

maxdeviant:
This PR updates the `IncomingRequestCfProperties` to reflect that `as_organization` is optional.

We ran into a case today in production where `cf.asOrganization` was `undefined`.

We're not using `worker` directly, but we are using bindings based on `worker-sys`, which resulted in the Worker panicking when trying to call the `as_organization` method.

I posted a [question](https://discord.com/channels/595317990191398933/1394398763119804528) about this in the Cloudflare Developers Discord and was told that this field can be optional and its presence should not be assumed:

Screenshot 2025-07-14 at 3 40 25 PM


guybedford (approved):
Thanks for posting the PR, yes this is definitely optional.

PR #784, opened 2025-07-18
Feat: Marked SendFuture and SendWrapper as unsafe types

parzivale:
Moved away from general unsafe impls to better communicate the usage style of SendFutures, removed unneeded Sync + Send impls from the Ai Module (Can instead implement SendWrapper).

This will probably constitute an api breakage and should (maybe?) go in the next release if accepted.


guybedford:
CI seems to be failing here after rebase.

Agreed this is the better user experience, would be happy to land it for the next major release after the current patch release.

parzivale:
@guybedford should be good to merge now

parzivale:
Hey what is missing from this for it to be mergeable?

guybedford:
I posted https://github.com/cloudflare/workers-rs/pull/861 here which just removes send entirely. Putting this one out there as a stake to question why we need this at all. At the very least I would have expected a failing test. Feedback and thoughts very much welcome.

parzivale:
As you mentioned in #861 this is due to Axum requiring send for it's handlers

parzivale:
Plus a lot of crates require send, I believe it is useful to expose a sensible default for implementing unsafe send

guybedford:
Ok, so just to get my head clear on the framing - `SendFuture` and `SendWrapper` are effectively wasm-bindgen helpers to allow single threaded apps to treat JS types as send.

Then working through still from fundamentals - why do we need to wrap the whole future? Can't we just make all JS types implement Send?

guybedford:
What if we just had a wasm bindgen flag like `wasm-bindgen --unsafe-single-threaded-send` that made all JS types unsafe send?

parzivale:
Hmm I mean it could work, you would probably would want both a crate feature and a cli feature as if threading ever came to wasm you would have to deactivate all threading code. Tbh Im nowhere near qualified enough to fully answer this

guybedford:
wasm-bindgen supports threading fine - so we would not allow the flag to exist alongside any use of threading.

Let me see if I can hack something together, will post back here shortly.

guybedford:
Yeah it all seems to work, implemented https://github.com/wasm-bindgen/wasm-bindgen/pull/4770 and applied this to https://github.com/cloudflare/workers-rs/pull/861. Further feedback welcome.

parzivale:
Yep I think yours is the better solution, unless workerd ever plans to move to a threaded execution model this makes everything considerably simpler.

parzivale:
Ill keep this pr open while https://github.com/wasm-bindgen/wasm-bindgen/pull/4770 is open

guybedford (approved):
861 on wasm bindgen has stalled since requiring a cfg for this would make it not work by default in the Rust linter.

Therefore we should land this at least in the mean time. Thanks for your patience I will include this in the release tomorrow.

Just to clarify, this is a non-breaking change correct?

Issue #757, opened 2025-06-19
[Feature] Support Secret Store Secrets


rawkode:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

When trying to consume a secret store secret, the error is `Binding cannot be cast to the type String from Fetcher`.

The TypeScript SDK requires an additional `.get()` on Secret Store secrets, which I suspect will be needed here also.

guybedford:
Can you share some more details on the exact situation that causes this issue?

Or can you suggest a simple way to replicate this with the existing test at https://github.com/cloudflare/workers-rs/blob/main/worker-sandbox/src/request.rs#L89-L91?

rawkode:
It fails when using this style of binding:

```
{
  "secrets_store_secrets": [
    {
      "binding": "<BINDING_VARIABLE>",
      "store_id": "<STORE_ID>",
      "secret_name": "<MY_SECRET_NAME>"
    }
  ]
}
```

I'll throw together a reproduction

guybedford:
PR now up for this in https://github.com/cloudflare/workers-rs/pull/786.

PR #786, opened 2025-07-24
feat: SecretStore implementation, Miniflare upgrade

guybedford:
Adds support for SecretStore bindings via `env.binding("SECRET_STORE").secret_store()`.

In order to support testing includes a full Miniflare upgrade from v3 to v4.

Resolves https://github.com/cloudflare/workers-rs/issues/757.


worker-sandbox/tests/auto_response.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, test, expect } from "vitest";
-import { mf } from "./mf";
+import { mf, mfUrl } from "./mf";
```
lukevalenta, line 2: Looks like instances of `http://fake.host/` have not yet been replaced (in this and a few other files).

worker/src/secret_store.rs
```diff
@@ -0,0 +1,95 @@
+use crate::{
+    send::{SendFuture, SendWrapper},
+    EnvBinding, Fetcher, Result,
+};
+use wasm_bindgen::JsCast;
+use wasm_bindgen_futures::JsFuture;
+
+mod sys {
```
lukevalenta: Should this go in the `worker-sys` crate for consistency with other features? There are a couple of other instances of `extern "C"` blocks in the `worker` crate, but many more in `worker-sys`.

worker-sandbox/tests/auto_response.spec.ts
guybedford: Thanks, updated.

worker/src/secret_store.rs
guybedford: Good point! appreciated for the quick review here.

worker-sys/src/types/secret_store.rs
```diff
@@ -0,0 +1,15 @@
+mod sys {
```
lukevalenta: Not sure it makes any difference, but I think you could remove the `mod sys` block. Most of the other types files don't have it (e.g., `worker-sys/src/types/context.rs`).

lukevalenta (approved):
Modulo the failing build, lgtm!

PR #794, opened 2025-08-30
fix: add BLOB support to SQL storage

connyay:
Fixes missing BLOB support in SQL storage functionality including ArrayBuffer conversion, iterator methods for BLOB data, and test coverage for various BLOB operations.

### DIFF
--- a/worker/src/sql.rs
+++ b/worker/src/sql.rs
@@ -154,6 +154,12 @@ impl TryFrom<JsValue> for SqlStorageValue {
             let mut bytes = vec![0u8; uint8_array.length() as usize];
             uint8_array.copy_to(&mut bytes);
             Ok(SqlStorageValue::Blob(bytes))
+        } else if js_val.is_instance_of::<js_sys::ArrayBuffer>() {
+            let array_buffer = js_sys::ArrayBuffer::from(js_val);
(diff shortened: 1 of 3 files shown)


worker/src/sql.rs
```diff
@@ -154,6 +154,12 @@ impl TryFrom<JsValue> for SqlStorageValue {
             let mut bytes = vec![0u8; uint8_array.length() as usize];
             uint8_array.copy_to(&mut bytes);
             Ok(SqlStorageValue::Blob(bytes))
+        } else if js_val.is_instance_of::<js_sys::ArrayBuffer>() {
+            let array_buffer = js_sys::ArrayBuffer::from(js_val);
```
guybedford: This can be done with a cast instead of `ArrayBuffer:from` - if we already know it is an instance of array buffer.

See `dyn_into` - https://github.com/wasm-bindgen/wasm-bindgen/blob/main/src/cast.rs#L43.

The checked cast also does an instance of check, so that could even replace the instance of call as well.

Might be worth refactoring the other cases in this direction too.

guybedford (dismissed):
Great to have the coverage here as well, thanks.

guybedford (approved):

Issue #796, opened 2025-09-01
[Feature] Containers Support


hr567:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

[Cloudflare Containers](https://developers.cloudflare.com/containers/) are in beta now. Currently, it appears that Containers cannot be used through the existing Rust bindings. Is there any plan to support Containers in worker-rs? Or is there currently any way to utilize Container capabilities? Thank you.

hr567:
I implemented a binding for Containers. I'm not remarkably familiar with JS and wasm-bindgen, could someone help review the code and offer some suggestions? Thank you very much! https://github.com/cloudflare/workers-rs/pull/797

guybedford:
Thanks for the PR, I will aim to review this soon.

guybedford:
This is going out in 0.6.3 now.

PR #797, opened 2025-09-02
Add Containers bindings

hr567:
#796


guybedford (changes requested):
Amazing work, we just need to ensure we have a test in the sandbox if possible.

hr567:
I add two test cases for container http and websocket. But in the test the `DurableObjectState::container` method cannot get the `Container`. I tried several different approaches, but they all failed. Since I'm not very familiar with JS, I might have missed something.
Could someone please help me figure out what's causing this? Is it due to the binding code or an issue with the environment setup? Thank you so much!

hr567:
Formatter with http feature is fixed

hr567:
Test by manually running `npx wrangler dev` in the worker-sandbox, then use `curl http://localhost:8787/container/echo --data 123abc`. The worker-sandbox runs correctly, and the Durable Object can correctly retrieve the container field.

hr567:
`npm run test{,-http}` currently passes locally when the container image is manually specified. However, issues with mirror building occur when using GitHub CI, which need to be resolved.
The issue with the durable object lacking a container member previously occurred due to incomplete configuration in Miniflare or specifying a Dockerfile instead of an image URL. CI testing also requires a properly built image, which may involve changes to the CI workflow, but those modifications are not yet ready.
A problem was discovered during the development of test cases. Enabling the `http` feature and performing try_into between worker::Request and http::Request causes a runtime error. The first patch fixes this issue, but the root cause remains unidentified, and it's unclear whether additional fixes are needed.

worker-sandbox/tests/mf.ts
```diff
@@ -61,6 +61,13 @@ const mf_instance = new Miniflare({
         PUT_RAW_TEST_OBJECT: "PutRawTestObject",
         AUTO: "AutoResponseObject",
         MY_CLASS: "MyClass",
+        ECHO_CONTAINER: {
+          className: "EchoContainer",
+          useSQLite: true,
+          container: {
+            imageName: "cloudflare-dev/echocontainer:69ea212d",
```
hr567: `imageName` needs to be compatible with the name of the image generated by CI

guybedford:
> Enabling the http feature and performing try_into between worker::Request and http::Request causes a runtime error.

Try using `JsCast` here, perhaps even as an unchecked cast. https://docs.rs/wasm-bindgen/latest/wasm_bindgen/trait.JsCast.html#method.unchecked_into

For the image name, can we make this an environment variable that we set for the test run? And then only run the container test when that environment variable is set so that the normal test suite skips this test otherwise.

Then we can do the container build and set this environment variable all in CI.

Let me know if that sounds like a good path forward.

Running a release cycle now, but we will get this one in for the next.

hr567:
> Try using `JsCast` here, perhaps even as an unchecked cast. https://docs.rs/wasm-bindgen/latest/wasm_bindgen/trait.JsCast.html#method.unchecked_into

`http` crate does not generate JS bindings via wasm-bindgen, so directly converting an `http::Request` appears to involve more effort and complexity.

hr567:
> For the image name, can we make this an environment variable that we set for the test run? And then only run the container test when that environment variable is set so that the normal test suite skips this test otherwise.

I think this is a good idea. By setting an environment variable, we can temporarily skip this test. Once we have a usable test image available, we can re-enable this test and integrate it into CI.

hr567:
I added the Docker image for testing in CI via AI consultation. Now, the CI will build the test container image every time it runs. The [CI should now pass](https://github.com/hr567/workers-rs/actions/runs/17791866162).

guybedford:
Thanks for the work here!

PR #819, opened 2025-09-18
worker-build fixes

guybedford:
Fixes various further worker-build issues:

* wasm-opt is now always used to the exact downloaded version
* wasm-opt flags use `--all-features` so we never have to worry about missing features again
* removes the console handler from tests, examples and templates now that this is no longer needed
* fixes panic hook reattachment
* integrates proper durable object class proxy wrappers for panic hook recovery
* includes a patch for heap reassignment in state recovery (pending upstream fix in https://github.com/wasm-bindgen/wasm-bindgen/pull/4665)
* removes the disabling of wasm-opt in examples, tests and templates


evmar:
Looks like this is the cause of https://github.com/cloudflare/workers-sdk/issues/10765

PR #821, opened 2025-09-21
doc: expand documentation for AI module

moosingin3space:
The Rustdoc for the Workers AI module was completely bare, and so this patch expands slightly on it.


guybedford (approved):
Thanks for posting.

Issue #722, opened 2025-04-14
[BUG] Memory leak in Rust Durable Object eviction


lukevalenta:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What version of `workers-rs` are you using?

0.5.0

### What version of `wrangler` are you using?

4.10.0

### Describe the bug

In Rust Workers, memory allocated for a Durable Object is not freed upon eviction. I'm able to reproduce locally with miniflare, and seem to have hit this same issue in a production worker (https://github.com/cloudflare/azul/issues/11).

### Steps To Reproduce

With the following files:
```
cat Cargo.toml
[package]
name = "memory-leak"
version = "0.1.0"
edition = "2021"

[package.metadata.release]
release = false

# https://github.com/rustwasm/wasm-pack/issues/1247
[package.metadata.wasm-pack.profile.release]
wasm-opt = false

[lib]
crate-type = ["cdylib"]

[dependencies]
worker = { version = "0.5.0" }

cat wrangler.jsonc
{
	"name": "memory-leak",
    "main": "build/worker/shim.mjs",
	"build": {
		"command": "cargo install -q worker-build && worker-build --release"
	},
	"compatibility_date": "2025-04-08",
	"migrations": [
		{
			"new_sqlite_classes": [
				"MyDurableObject"
			],
			"tag": "v1"
		}
	],
	"durable_objects": {
		"bindings": [
			{
				"class_name": "MyDurableObject",
				"name": "MY_DURABLE_OBJECT"
			}
		]
	},
	"observability": {
		"enabled": true
	}
}

cat src/lib.rs
use wasm_bindgen::prelude::*;
#[allow(clippy::wildcard_imports)]
use worker::*;

#[event(fetch, respond_with_errors)]
async fn main(_req: Request, env: Env, _ctx: Context) -> Result<Response> {
    let ns = env.durable_object("MY_DURABLE_OBJECT")?;
    let id = ns.id_from_name("foo")?;
    let stub = id.get_stub()?;
    stub.fetch_with_str("http://example.com").await
}

#[durable_object]
struct MyDurableObject {
    count: u64,
    _buffer: Vec<u8>,
}

#[durable_object]
impl DurableObject for MyDurableObject {
    fn new(state: State, _env: Env) -> Self {
        Self {
            count: 0,
            _buffer: Vec::with_capacity(100_000_000),
        }
    }
    #[allow(clippy::unused_async)]
[...]

PR #822, opened 2025-09-23
Fix instance memory leak

guybedford:
This fixes a memory leak when handling multiple instances of durable objects.

The panic handler proxy system stored the instances assuming one-time use, but we need to ensure GC'able reuse.

This resolves that handling by associating each of the proxies lazily with its instance identity, and then instead of reconstructing instances on the panic reset, instead reconstructing instances on the next entrypoint call.

This also inlines a large allocation for memory testing on the local test suite thanks to @lukevalenta's replication. In the process, including miniflare and compat flag updates to properly test finalization.


guybedford:
Resolves https://github.com/cloudflare/workers-rs/issues/722.

PR #829, opened 2025-09-25
Added custom_shim option to current worker build

parzivale:
Re-added support for custom shims, my current project requires that i defined Rpc methods on my workers which need both Ctx and Env provided


guybedford:
Thanks for posting this one, I've implemented an alternative in https://github.com/cloudflare/workers-rs/pull/831 will land this there instead.

PR #832, opened 2025-09-26
Support eager memory deallocation on hibernation

guybedford:
While formally not a memory leak, this reduces Wasm memory usage for durable objects subject to hibernation by eagerly freeing Wasm memory associated with the stale instance when a new instance is constructed.

### DIFF
--- a/worker-build/src/js/shim.js
+++ b/worker-build/src/js/shim.js
@@ -15,14 +15,14 @@ function registerPanicHook() {
 
 registerPanicHook();
 
-let instanceId = 0;
+let reinitId = 0;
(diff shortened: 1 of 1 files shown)


guybedford:
Unfortunately this fails because destroying the last instance when a new instance is constructed doesn't work since it seems we support constructing multiple in the same isolate of a given durable object.

guybedford:
It turns out this requires upstream work to properly support, given the multi-DO handling.

guybedford:
An upstream change for an unload event is posted in https://github.com/cloudflare/workerd/pull/5211.

PR #833, opened 2025-09-28
feat: create trait definitions for model and streamable model

parzivale:
The main objective behind this pr is to simplify working with streaming responses from text generation models, however the design should be flexible enough for other applications.

For now the Model and StreamableModel are left to the user to implement for their use case, but the ideal end goal for this pr is to define types for all the I/O interfaces workers-ai uses and then seal Model and StreamableModel.

### DIFF
--- a/test/src/ai.rs
+++ b/test/src/ai.rs
@@ -1,21 +1,9 @@
 use worker::{
-    worker_sys::{AiTextGenerationInput, AiTextGenerationOutput},
-    Env, Model, Request, Response, Result, StreamableModel,
+    models::llama_4_scout_17b_16e_instruct::Llama4Scout17b16eInstruct,
+    worker_sys::AiTextGenerationInput, Env, Request, Response, Result,
 };
 
 use crate::SomeSharedData;
-
-pub struct Llama4Scout17b16eInstruct;
-
-impl Model for Llama4Scout17b16eInstruct {
-    const MODEL_NAME: &str = "@cf/meta/llama-4-scout-17b-16e-instruct";
-
-    type Input = AiTextGenerationInput;
-
-    type Output = AiTextGenerationOutput;
-}
-
-impl StreamableModel for Llama4Scout17b16eInstruct {}
 
 const AI_TEST: &str = "AI_TEST";
 
--- a/test/src/ai.rs
+++ b/test/src/ai.rs
@@ -1,32 +1,18 @@
-use serde::{Deserialize, Serialize};
-use worker::{Env, Model, Request, Response, Result, StreamableModel};
+use worker::{
+    worker_sys::{AiTextGenerationInput, AiTextGenerationOutput},
+    Env, Model, Request, Response, Result, StreamableModel,
+};
 
 use crate::SomeSharedData;
 
 pub struct Llama4Scout17b16eInstruct;
 
-#[derive(Serialize)]
-pub struct DefaultTextGenerationInput {
-    pub prompt: String,
-}
-
-#[derive(Deserialize)]
-pub struct DefaultTextGenerationOutput {
-    pub response: String,
-}
-
-impl From<DefaultTextGenerationOutput> for Vec<u8> {
-    fn from(value: DefaultTextGenerationOutput) -> Self {
-        value.response.into_bytes()
-    }
-}
-
 impl Model for Llama4Scout17b16eInstruct {
     const MODEL_NAME: &str = "@cf/meta/llama-4-scout-17b-16e-instruct";
 
-    type Input = DefaultTextGenerationInput;
+    type Input = AiTextGenerationInput;
 
-    type Output = DefaultTextGenerationOutput;
+    type Output = AiTextGenerationOutput;
 }
 
 impl StreamableModel for Llama4Scout17b16eInstruct {}
@@ -41,11 +27,12 @@ pub async fn simple_ai_text_generation(
 ) -> Result<Response> {
     let ai = env
         .ai(AI_TEST)?
-        .run::<Llama4Scout17b16eInstruct>(DefaultTextGenerationInput {
-            prompt: "What is the answer to life the universe and everything?".to_owned(),
-        })
+        .run::<Llama4Scout17b16eInstruct>(
+            AiTextGenerationInput::new()
+                .set_prompt("What is the answer to life the universe and everything?"),
+        )
         .await?;
-    Response::ok(ai.response)
+    Response::ok(ai.get_response().unwrap_or_default())
 }
 
 #[worker::send]
@@ -56,9 +43,10 @@ pub async fn streaming_ai_text_generation(
 ) -> Result<Response> {
     let stream = env
         .ai(AI_TEST)?
-        .run_streaming::<Llama4Scout17b16eInstruct>(DefaultTextGenerationInput {
-            prompt: "What is the answer to life the universe and everything?".to_owned(),
-        })
+        .run_streaming::<Llama4Scout17b16eInstruct>(
+            AiTextGenerationInput::new()
+                .set_prompt("What is the answer to life the universe and everything?"),
+        )
         .await?;
 
     Response::from_stream(stream)
--- a/test/src/ai.rs
+++ b/test/src/ai.rs
@@ -1,0 +1,65 @@
+use serde::{Deserialize, Serialize};
+use worker::{Env, Model, Request, Response, Result, StreamableModel};
+
+use crate::SomeSharedData;
+
+pub struct Llama4Scout17b16eInstruct;
+
+#[derive(Serialize)]
+pub struct DefaultTextGenerationInput {
+    pub prompt: String,
+}
+
+#[derive(Deserialize)]
+pub struct DefaultTextGenerationOutput {
+    pub response: String,
+}
+
+impl From<DefaultTextGenerationOutput> for Vec<u8> {
+    fn from(value: DefaultTextGenerationOutput) -> Self {
+        value.response.into_bytes()
+    }
+}
+
+impl Model for Llama4Scout17b16eInstruct {
+    const MODEL_NAME: &str = "@cf/meta/llama-4-scout-17b-16e-instruct";
+
+    type Input = DefaultTextGenerationInput;
+
+    type Output = DefaultTextGenerationOutput;
+}
+
+impl StreamableModel for Llama4Scout17b16eInstruct {}
+
+const AI_TEST: &str = "AI_TEST";
+
+#[worker::send]
+pub async fn simple_ai_text_generation(
+    _: Request,
+    env: Env,
+    _data: SomeSharedData,
+) -> Result<Response> {
+    let ai = env
+        .ai(AI_TEST)?
+        .run::<Llama4Scout17b16eInstruct>(DefaultTextGenerationInput {
+            prompt: "What is the answer to life the universe and everything?".to_owned(),
+        })
+        .await?;
+    Response::ok(ai.response)
+}
+
+#[worker::send]
+pub async fn streaming_ai_text_generation(
+    _: Request,
+    env: Env,
+    _data: SomeSharedData,
+) -> Result<Response> {
+    let stream = env
+        .ai(AI_TEST)?
+        .run_streaming::<Llama4Scout17b16eInstruct>(DefaultTextGenerationInput {
+            prompt: "What is the answer to life the universe and everything?".to_owned(),
+        })
+        .await?;
+
+    Response::from_stream(stream)
+}
--- a/worker-sys/src/types/ai.rs
+++ b/worker-sys/src/types/ai.rs
@@ -3,7 +3,7 @@ use std::iter::FromIterator;
 use js_sys::{Array, Promise};
 use wasm_bindgen::prelude::*;
 
-use crate::scoped_chat::RoleScopedChatInput;
+use crate::typed_array;
 
 #[wasm_bindgen]
 extern "C" {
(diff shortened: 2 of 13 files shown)


guybedford:
Very interesting approach. It would be great to see some tests for the sorts of workflows you're seeking to enable here, and that would also help for review too.

parzivale:
Sure! should I create a new entry in the test folder for this? would there be a better place to put the tests?

guybedford:
Yes, in the tests folder, we have one big test app that is built altogether and tested as one.

parzivale:
There isn't currently an AI binding in the tests wrangler.toml, I can test locally on my account via a new binding but this will break CI right?

guybedford:
You should just be able to add a new binding there, if there's issues with that happy to look into it further.

parzivale:
@guybedford I've added new test cases to show a very simple example implementation

test/src/ai.rs
```diff
@@ -0,0 +1,65 @@
...
+pub async fn simple_ai_text_generation(
+    _: Request,
+    env: Env,
+    _data: SomeSharedData,
+) -> Result<Response> {
+    let ai = env
+        .ai(AI_TEST)?
+        .run::<Llama4Scout17b16eInstruct>(DefaultTextGenerationInput {
+            prompt: "What is the answer to life the universe and everything?".to_owned(),
+        })
```
guybedford: Instead of `DefaultTextGenerationInput`, could we implement [`AiTextGenerationInput`](https://workers-types.pages.dev/#AiTextGenerationInput) directly in core?

Along the lines of -

```rust
pub struct AiTextGenerationInput {
    pub prompt: Option<String>,
    pub raw: Option<bool>,
    pub stream: Option<bool>,
    pub max_tokens: Option<u32>,
    pub temperature: Option<f32>,
    pub top_p: Option<f32>,
    pub top_k: Option<u32>,
    pub seed: Option<u32>,
    pub repetition_penalty: Option<f32>,
    pub frequency_penalty: Option<f32>,
    pub presence_penalty: Option<f32>,
    pub messages: Option<Vec<RoleScopedChatInput>>,
    pub response_format: Option<AiTextGenerationResponseFormat>,
    pub tools: Option<serde_json::Value>, // For flexible union type
    pub functions: Option<Vec<AiTextGenerationFunctionsInput>>,
}
```

We don't even need all fields initially.

And similarly for https://workers-types.pages.dev/#AiTextGenerationOutput?

test/src/ai.rs
```diff
@@ -0,0 +1,65 @@
+use serde::{Deserialize, Serialize};
+use worker::{Env, Model, Request, Response, Result, StreamableModel};
+
+use crate::SomeSharedData;
+
+pub struct Llama4Scout17b16eInstruct;
```
guybedford: Agreed this seems like the right approach!

guybedford (commented):
Amazing work, this looks incredible. Thank you for working on this. Just a couple of points about standardizing the input and output options - I know this gets into the deeper bindings questions, but would be good to have a plan at least.

test/src/ai.rs
parzivale: Sure! that works. was trying to match the type definitions emitted by wrangler types, but that does seem more legible. Also would it make sense to mark the fields as private and use a builder pattern to construct the input?

parzivale: Also for tools I have a seperate pr I would like to make which is based on this one which handles tools as a new sub trait of model, the gist of it is you define tools like the following

```rust

type ToolRun = Rc<
    dyn Fn(
        &Env,
        Option<serde_json::Map<String, serde_json::Value>>,
    ) -> Pin<Box<dyn Future<Output = Result<String, Error>>>>,
>;

#[derive(Clone)]
pub struct Tool {
    pub name: &'static str,
    pub description: &'static str,
    pub run: ToolRun,
}

```

and then would would call them from ai with the following method

```rust

impl Ai {
  pub async fn run_with_tools<M: ToolsModel>(&self, input: &M::Input, tools: &[Tool]) -> Result<M::Ouput, Error>  {
  ...
  }
}
```

this would require a new trait called ToolsModel like this

```rust

pub trait ToolsModel: Model {}

```

The problem is I'm not sure this approach is generic enough to work for everyone. For instance I know normally working with axum  you are expected to wrap env in an Arc which mean you no loner have access to this functionality

guybedford: As much as possible it would be good to follow Wasm bindgen semantics. I know we wrap all types currently, but I'm trying to move to a model where when possible we should use worker-sys types as the high level types.

So the wasm-bindgen pattern would be to define an imported type, and define all properties as getters and setters:

```rust
#[wasm_bindgen]
extern "C" {
    # [wasm_bindgen (extends = :: js_sys :: Object , js_name = AiTextGenerationInput)]
    #[derive(Debug, Clone, PartialEq, Eq)]
    #[doc = "Ai Text Generation"]
    pub type FetchEventInit;
    #[doc = "Prompt for generation"]
    #[wasm_bindgen(method, setter = "prompt")]
    pub fn set_prompt(this: &AiTextgenerationInput, prompt: Option<String>);
    #[wasm_bindgen(method, getter = "prompt")]
    pub fn get_prompt(this: &AiTextgenerationInput) -> Option<String>;
}
```

Then functions would just take `&worker-sys::AiTextGenerationInput` input directly. And we'd likely alias it as `worker::ai::AiTextGenerationInput`.

parzivale: Right! gotcha! I didn't realise you were referring to types in js, thanks for clarifying

parzivale: I just looked into AiTextGeneration in the docs you provided and it seems AiTextGeneration is a typescript type not a class, so there is no imported type that we can bind to in the js namespace.

guybedford: That should be fine - it's a structural type not a formal import.

parzivale: Ah sorry I just saw your comment now, I just made a commit where I kept the AiTextGenerationInput struct in worker rather than binding to js, if you don't like the api I can change it

guybedford: I'm strongly trying to formalize our conventions, so let me think about this some more and follow-up soon. My initial sense is that serde perf is equivalent so we should prefer the direct interpretation convention, but I want to ensure I've got that right.

parzivale: gotcha personally I prefer being able to set options like this in the struct itself like

```rust
SomeConfig {
  option1: Some(123),
  option2: Some("Hello world"),
  ..default()
 }
 ```
rather than this
 ```rust
let config = SomeConfig::new();

config.option1(123);
config.option2("hello world");
 ```

but this is a personal preference, I'm open to amending this commit

Comment (guybedford) [reply test/src/ai.rs]: Hmm, still thinking this one through but I've posted up https://github.com/wasm-bindgen/wasm-bindgen/pull/4724 in the mean time. I still tend to think structural setter builder patterns on the base type is where we want to be for canonical representations in this project.

Comment (parzivale) [reply test/src/ai.rs]: I came up with a custom solution for chaining in my most recent commit, let me know if this matches what you are thinking for conventions

Comment (guybedford) [reply test/src/ai.rs]: Thanks. I'm continuing to think about what canonical bindgen looks like, and have decided a builder pattern would be fine here after all. I'm not going to land setter chaining for this reason upstream.

The important point for a builder though is that it should return the `worker-sys` type at the end of the build, so the wrapper is just for the builder validations only, and we don't have a wrapper type left afterwards.

So as long as we minimize wrappers I'm otherwise okay with it.

Then could your RoleScopedChat wrapper just be a custom impl on top of the base worker-sys type?

Comment (parzivale) [reply test/src/ai.rs]: What do you mean by custom impl? the current RoleScopedChat is a js object, would you mean dropping the custom constructors I made?

Comment (guybedford) [reply test/src/ai.rs]: I mean the inner access on RoleScopedChatInput - if we could unify that type as the base bindgen type rather than having an inner wrapping. That's the pattern I want to start moving towards overall in the codebase.

Comment (parzivale) [reply test/src/ai.rs]: Just to clarify, RoleScopedChatInput is already a JS object bound via wasm_bindgen, defined like this:

```rust

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(extends = js_sys::Object)]
    pub type RoleScopedChatInput;

    #[wasm_bindgen(constructor, js_class = Object)]
    fn new() -> RoleScopedChatInput;
}

```

I might be misunderstanding your suggestion, could you show what you’d expect the “custom impl on top of the base worker-sys type” to look like in this case?

guybedford: Ahh nice, yes that's exactly what I mean seems we no longer have the misdirection.

Note that I think `set_role_inner` should be a `set_role` directly?

parzivale: I didnt want to expose them as I want users to use the custom constructors I made as when generating conversations as I thought it would be easier to do this

```rust

let convo = vec![system("you are a helpful assistant"), user("hi")]

```

rather than 

```rust
let user_message = RoleScopedChatInput::new();
user_message.set_role("user");
user_message.set_content("hi");

let system_message = RoleScopedChatInput::new();
system_message.set_role("system");
system_message.set_content("you are a helpful assistant");

let convo = vec![system_message, user_message];

```

Though if this api is better suited to be in worker rather than worker-sys I can move this out and replace the old setters

parzivale: I also created the module scoped_chat as I didn't want to collide with any functions. So if a collision were to occur if users need to they can use this instead

```rust

let convo = vec![scoped_chat::system("you are a helpful assistant"), scoped_chat::user("hi")];

```

guybedford: As I said, I'd be happy with any kind of builder pattern though:

```rust
let convo = Conversation::builder()
   .system("foo")
   .user("hi")
   .build();
```

or however you think the API can be best designed / what seems most idiomatic.

The only important point here really is that the final `convo` is a `Vec<RoleScopedChatInput>` Wasm Bindgen native type directly.

parzivale: I have created a new macro for generating typed arrays in wasm bindgen and moved the builder funcs to be associated methods of the array. Basically we can now do the following

```rust

let convo = RoleScopedChatInputArray::new()
   .system("you are a helpful assistant")
   .user("hello");

```

I made the typed_array macro as I think this way of defining typed arrays will get used a lot if we decide to define types like this

Also if we do go with this method I might move it over to the proc macro crate as then I could implement the iterator traits on the array

parzivale:
Hey! I'm eager to merge this, do you have any feedback on what needs to be changed?

worker-sys/src/types/utils/typed_array.rs
```diff
@@ -0,0 +1,182 @@
...
+            ) -> ::js_sys::JsString;
+
+            #[wasm_bindgen(method, js_name = toString)]
+            pub fn to_string(this: &$name) -> ::js_sys::JsString;
+
+            #[wasm_bindgen(method)]
+            pub fn unshift(this: &$name, value: &$type) -> u32;
+        }
+    };
+}
```
guybedford: How is this file different to what is already in web_sys, and can't we just use that directly?

worker-sys/src/types/utils/typed_array.rs
```diff
@@ -0,0 +1,182 @@
+#[macro_export]
+macro_rules! typed_array {
```
guybedford: Can we not use the typed array already in `js_sys` here?

guybedford (commented):
@parzivale thanks for your patience here, the implementation looks really clean although I am a little bit concerned we may want to adapt the API in future.

We are okay with making breaking changes though so it would certainly be better to land and release than not.

I'm still not quite sure I follow why we need a custom Stream type for AI specifically over just improving the lower-level streaming primitives to support the use case. I'm also unclear on why we need a new type of typed array binding.

worker-sys/src/types/utils/typed_array.rs
parzivale: The typed array here is for any generic T not just the uint/int arrays from js_sys. It's just a typesafe wrapper around js_sys::Array. Its useful for dealing with collections that are js based.

parzivale:
> I'm still not quite sure I follow why we need a custom Stream type for AI specifically over just improving the lower-level streaming primitives to support the use case.

The idea behind that was to split out the different return types, for streaming replies we don't expect a type T that implements Deserialize/Serialize like we would normally do with a model response. Instead we expect a stream of T. I would be happy to break out the streaming implementation into something else, but I do think it is important to distinguish between streaming and non-streaming responses.

worker-sys/src/types/utils/typed_array.rs
guybedford: I'm working on a PR to add a backwards compatible `Array<T>` support to wasm bindgen for typed JS arrays.

It sounds like this crosses over with that use case, which will hopefully be released before the end of the year.

So we have two options, and I'll let you pick:

1. Figure out the bare minimum to ship now - perhaps using an untyped array, or custom function or some minimal approach we can make a breaking change on later, OR
2. Base this branch to the new generics work upstream in wasm bindgen, and land it when that lands in December sometime

guybedford:
> The idea behind that was to split out the different return types, for streaming replies we don't expect a type T that implements Deserialize/Serialize like we would normally do with a model response. Instead we expect a stream of T. I would be happy to break out the streaming implementation into something else, but I do think it is important to distinguish between streaming and non-streaming responses.

Would typed ReadableStream solve this as an alternative if we had that generic?

worker-sys/src/types/utils/typed_array.rs
parzivale: I mean workers rs is already using a forked version of wasm bindgen, of we use your pr will that postpone the use of stable wasm-bindgrn by a lot? Ie is there a plan to move to a non forked version before that?

I don't think keeping a fork of wasm-bindgen is a good idea as there seems to already be friction (ie #846)

parzivale:
> > The idea behind that was to split out the different return types, for streaming replies we don't expect a type T that implements Deserialize/Serialize like we would normally do with a model response. Instead we expect a stream of T. I would be happy to break out the streaming implementation into something else, but I do think it is important to distinguish between streaming and non-streaming responses.
> 
> Would typed ReadableStream solve this as an alternative if we had that generic?

No as there needs to be an additional wrapper as workers ai stream returns see events rather than just the T we need, so there needs to be a bit of postprocessing for things like tool calls when streaming.

worker-sys/src/types/utils/typed_array.rs
guybedford: We are not using a fork of wasm bindgen. The patches only apply locally to allow us to test wasm bindgen work in progress before releases (such as building branches on top of generics work before it is released). We cannot release anything that depends on unreleased wasm bindgen because we only release against released wasm bindgen. So the question here is about timelines.

guybedford:
> No as there needs to be an additional wrapper as workers ai stream returns see events rather than just the T we need, so there needs to be a bit of postprocessing for things like tool calls when streaming.

Could this be interpreted as a transform stream?

parzivale:
Yes it could! That would be a pretty good solution, the only problem is documentation, could we type alias it so in docs we can describe usage?

parzivale:
My normal first stop when dealing with return types is to go to definition and using a transform stream would make that harder

worker-sys/src/types/utils/typed_array.rs
parzivale: Ah my bad sorry, I misread the thread. Tbh I think just a temporary solution would be better and mark it under a unstable flag. What do you think?

guybedford: Sure, I'm happy to land something experimental if we document it as such and are open to changing later.

parzivale: Any changes needed to make? Other than the experimental feature flag

worker-sys/src/types/ai.rs
```diff
@@ -10,3 +14,228 @@ extern "C" {
...
+    pub fn system(self, content: &str) -> Self {
+        self.custom_role("system", content);
+        self
+    }
+
+    pub fn tool(self, content: &str) -> Self {
+        self.custom_role("tool", content);
+        self
+    }
+}
```
guybedford: Rather than putting these on the array, let's put these on the message builder itself.

So we would rather use directly:

```
messages.push(
    RoleScopedChatInput::builder()
        .role("user")
        .content("hello")
        .build()?
);
```

We may yet go either way on array generics - wrapper types or not. I don't want to assume one direction yet though. We definitely want builders at least.

worker/src/ai/models.rs
```diff
@@ -0,0 +1,55 @@
...
+    #[derive(Default, Debug, Serialize)]
+    #[serde(rename_all = "lowercase", untagged)]
+    pub enum Role {
+        #[default]
+        User,
+        Assistant,
+        System,
+        Tool,
+        Any(String),
+    }
```
guybedford, line 15: Can we not use `worker_sys` `Role` directly here instead?

worker/src/ai.rs
```diff
@@ -82,3 +96,75 @@ impl EnvBinding for Ai {
...
+    const MODEL_NAME: &str;
+    type Input: Into<JsValue>;
+    type Output: From<JsValue>;
+}
+
+pub trait StreamableModel: Model {}
+
+#[derive(Debug)]
+#[pin_project]
+pub struct AiStream<T: StreamableModel> {
```
guybedford, line 110: As I said I'm still not clear why AiStream is unique from a generic stream. Couldn't we just have TypedReadeableStream like you did for array?

guybedford (commented):
Thanks again for your patience. It's looking really great and so close to the ideal output that we want.

Feedback is all again for alignment with the ideal bindgen we are aiming for with upstream changes.

As much as we can get right now is a great help.

worker-sys/src/types/ai.rs
guybedford: Actually a generic array builder would be a useful pattern I suppose, but the trick with this would be constructing a builder API that can in theory support optimized creation of the whole array in a single buffer serialization (as an optimization, we don't actually have to do it yet).

That is, something more like:

```rust
messages.builder()
  .push_builder(|b| b.role("user").content("hello"))
  .push_builder(|b| b.role("user").content("hello"))
  .build()
```

where the `push_builder` method on `TypedArray<T>` is works for `T impl Builder` for some definition of a builder trait?

Or something along those lines. Let me know if that make sense where I'm trying to go with this?

worker/src/ai/models.rs
parzivale: I don't believe there is a Role struct in worker_sys, at least I couldn't find it

worker/src/ai.rs
parzivale: Yep we can I'll do what I said previously here and type alias it so we can document the output

worker/src/ai/models.rs
guybedford: Ahh I thought I saw it defined there as well here. No problem then.

worker-sys/src/types/ai.rs
parzivale: I'll create a generic array builder. I'm a little confused with the single buffer serialization optimization as the items in the array are jsvalues and are not serializeable

parzivale: Also currently the generic array isn't really a generic array just a helper macro which generates an array with typed helper methods, I can refactor to make a proper generic array. Though that might come with complications as I would need to cast every time an access is made rather than just use wasm-bindgen to define the return type

guybedford: In due course we will be getting a generic array upstream, just using that term here for now. I understand your macro is a custom type and I'm fine with this approach unflagged with the assumption that the change won't be too destructive down the line.

If we could do a generic array that would in theory be preferable (`TypedArray<T>` wrapping Array), but I've set enough requirements on this PR here I feel! Separately I would be interested to dig into your cast concern more as I didn't follow that, and it would help the upstream conversation there.

As for the optimization - the idea is that every call from Wasm to JS incurs a performance overhead. If we can batch those calls through an optimization as a single call that should be more performant, turning an array of N values from N calls out to JS to 1. Again not a hard constraint, just sharing the design space thinking.

guybedford: Specifically just that a Rust builder could theoretically implement such optimizations, but we don't actually need to do them now of course.

parzivale: For the casting concerns that was due to me misunderstanding what you wanted and assuming you wanted a new array wrapper which looked like this 

```rust
pub struct Array<T> {
   array: js_sys::Array,
   phantom: PhantomData<T>
}
```

Which wrapped all the associated methods on array with an extra casting step to T.

Regarding the optimization, I see what you mean I got a little confused due to the serialized wording and assumed you wanted something like a serde_wasm_bindgen call as part of the array builder build call.

parzivale: I've pushed an implementation of this, I tried to keep the bulk move to js in mind when constructing the builder, the push functionality to the array builder should probably have some refinement but I wanted to get it in front of your eyes first. Finally I would like to move the typedArray macro over to a proc macro before merging as I'm reaching for dirtier and dirtier hacks to avoid polluting the name space (that's why the exported typed array is referenced as RoleScopedChatInputArray::RoleScopedChatInputArray

worker/src/ai.rs
parzivale: I'll work on this tommorow

guybedford:
I'm still happy to review this for the release tomorrow if there is interest in finishing up the last change requests.

PR #837, opened 2025-10-10
Add client_accept_encoding method to Cf struct

alpgul:
This PR adds a wasm-bindgen getter to expose Cloudflare's clientAcceptEncoding property from the request cf object to Rust/WASM via IncomingRequestCfProperties::client_accept_encoding.

Reference: https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties


guybedford:
This is a good change, curious why it was closed?

PR #838, opened 2025-10-10
feat: add props support to Context

connyay:
Adds support for accessing props passed to workers via service bindings, enabling configuration to be passed based on invocation context.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -188,6 +188,9 @@ jobs:
       - name: Build Wasm Bindgen
         run: cd wasm-bindgen && cargo build -p wasm-bindgen-cli --bin wasm-bindgen --bin wasm-bindgen-test-runner
 
+      - name: Install wasm32 target
+        run: rustup target add wasm32-unknown-unknown
+
       - name: Run wasm-bindgen tests
         env:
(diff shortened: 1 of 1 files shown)


connyay:
I can't figure out how to add test coverage for this. 

Threads I pulled:

1. [miniflare](https://github.com/cloudflare/workers-sdk/tree/main/packages/miniflare) I dont see a way to stress context.props with serviceBindings 🤔 
2. basic unit tests in worker/src/context.rs ([example](https://gist.github.com/connyay/a3e594bbe529416a269c68ce6311e4fb)) but I cant figure out how to actually execute wasm_bindgen_test tests 🤔

guybedford:
Thank you for this one! Yes we don't currently run wasm_bindgen_test properly here... if you want to work on this it would be very much welcome, otherwise I think we are at the mercy of miniflare support.

I'm happy to land this without tests, but I'm also trying to figure out how this fits into the automated bindgen story as we think more about that too - so I may prefer to wait a few weeks on landing this to dwell on that a bit more if you don't mind.

connyay:
I missed `wasm-bindgen-test-runner` in my first pass. Seems to work locally. Will likely need some CI tweaks, but should be close

guybedford:
The generate failure seems unrelated here, I can aim to get that fixed.

For the test failure, seems as simple as adding the wasm32-unknown-unknown target so looks like we're almost good to land here.

PR #848, opened 2025-10-20
fix: respect `--mode no-install` and warn if `esbuild` is too old

lemmih:
I'm using `worker-build` in an environment without internet access.


guybedford:
Happy to provide this level of flexibility, but the extra version check is a perf cost we don't need to add.

We currently cache esbuild with pathing that informs the version, and assume we can obtain the correct version.

Perhaps the simplest approach is just to support an `ESBUILD_PATH` or `ESBUILD_BIN` environment var override?

guybedford:
I've implemented the suggestion here in https://github.com/cloudflare/workers-rs/pull/863.

PR #859, opened 2025-10-29
Make `Storage::get` return an option

rozbb:
This PR upstreams the fix in https://github.com/cloudflare/azul/pull/148 . The issue in that code base was that an error value from `Storage::get` was interpreted as a missing key _regardless of what the error was_. This turned out to be a huge footgun.

This PR changes the type signature to `get(key: &str) -> Result<Option<T>>`, returning `Ok(None)` when the key is not defined. This lets the caller distinguish between missing keys and proper errors.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


guybedford:
CI fix in https://github.com/cloudflare/workers-rs/pull/860, which I'll land now, then this can be based to that.

PR #861, opened 2025-10-30
Deprecate SendWrapper and #[worker::send] — JsValue is now Send

guybedford:
## Summary

`JsValue` (and all types wrapping it) is now `Send` in `wasm-bindgen`, so the vast majority of the `Send` ergonomic workarounds in this crate are no longer necessary.

The remaining `!Send` type is `JsFuture` (uses `Rc<RefCell<...>>` internally). This PR wraps all internal `JsFuture` usage with `SendFuture`, so every public async method now returns a `Send` future. Users no longer need to worry about `Send` at all.

### What changed

- **Removed all `unsafe impl Send/Sync`** on worker types (`Env`, `KvStore`, `D1Database`, `Request`, etc.) — they derive `Send` automatically now via `JsValue`.
- **Wrapped all internal `JsFuture` with `SendFuture`** — every public async method in the crate now returns a `Send` future. This is the key improvement: `Send`-ness is now the library's responsibility, not the user's.
- **Deprecated `SendWrapper`** — still exported from `worker::send` for backward compat, but no longer used internally. Users should use the inner type directly.
- **Deprecated `#[worker::send]`** — no longer needed since all public async methods return `Send` futures. The macro still works but emits a deprecation warning.
- **Removed all `#[worker::send]` usage** from tests and examples.

### Migration for users

- Remove `#[worker::send]` from your async handlers — it's no longer needed.
- Replace `send::SendWrapper::new(x)` with just `x`.
- `send::SendFuture` is still available if you work with `JsFuture` directly.

### Not a breaking change

Both `SendWrapper` and `#[worker::send]` continue to compile — they just emit deprecation warnings.

### DIFF
--- a/examples/axum/Cargo.lock
+++ b/examples/axum/Cargo.lock
@@ -342,9 +342,9 @@ checksum = "92ecc6618181def0457392ccd0ee51198e065e016d1d527a7ac1b6dc7c1f09d2"
 
 [[package]]
 name = "js-sys"
-version = "0.3.85"
+version = "0.3.90"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8c942ebf8e95485ca0d52d97da7c5a2c387d0e7f0ba4c35e93bfcaee045955b3"
+checksum = "14dc6f6450b3f6d4ed5b16327f38fed626d375a886159ca555bd7822c0c3a5a6"
 dependencies = [
  "once_cell",
  "wasm-bindgen",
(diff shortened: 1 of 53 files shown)


guybedford:
Ahh, found the failures here - I see this is because of Axum.

WERDXZ:
Cases like `leptos` ssr do want server functions to return a future that is `Send`, this (the `send` module) become useful when downstream users needs to make futures `Send` in workers environment, especially like the `JsFuture` type.

It might not be a great example of usage, but I am using it in my [website](https://github.com/WERDXZ/werdxz-info/blob/fa996e14bc010728791e3b2c9b560d0683257213/portfolio/src/worker_helpers.rs#L8).

guybedford:
I've updated this PR to use an experimental wasm bindgen feature for this in https://github.com/wasm-bindgen/wasm-bindgen/pull/4770, and this approach seems to work now.

guybedford:
All tests are passing - CI failure here is due to the examples build not using the forked wasm bindgen worker-build.

PR #862, opened 2025-10-31
feat: add support for lists in worker-codegen

sevki:
The current wit RPC system is somewhat limited to only being able to handle primitive types. These changes add experimental support for using list types, aka Vec in Rust.

An open question is, should list<u8> be handled as a special case to being a `::worker::js_sys::Uint8Array`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3498,17 +3498,6 @@ dependencies = [
 
 [[package]]
 name = "wasmparser"
-version = "0.205.0"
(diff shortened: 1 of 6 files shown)


guybedford:
Because RPC supports much more complex types such as functions and pipeline promises, steadily increasing surface area here doesn't actually get us to parity.

We are currently exploring TypeScript RPC generation that can fully capture the semantics of what we support today, and hope to have an update on this in the coming weeks.

PR #867, opened 2025-11-07
add benchmark suite

guybedford:
## Summary

Adds a benchmark suite and CodSpeed integration for continuous performance tracking of the workers-rs runtime.

## Benchmark

A self-contained Worker app in `benchmark/` that:
- Handles `/benchmark` requests by making 10 parallel sub-requests to `/stream`
- Each sub-request streams 1MB of data in 8KB chunks
- Timing uses `performance.now()` via `web-sys` for high-resolution measurement
- Runs against Miniflare locally via `node run.js`

## CodSpeed Integration

- `.github/workflows/codspeed.yml` runs on pushes to `main`, PRs, and manual dispatch
- Uses walltime mode on `ubuntu-latest`
- `codspeed-convert.js` transforms benchmark results into CodSpeed's walltime format
- Tracks two metrics: **worker internal time** and **end-to-end time**

## Review Feedback Addressed

- Updated miniflare to `^4.20260302.0`
- Replaced `npm run` with `node --run`
- Replaced `Date::now().as_millis()` with `performance.now()` via `web-sys::Performance`

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -45,18 +45,12 @@ jobs:
         working-directory: benchmark
         run: npm install
 
-      - name: Build and run benchmarks
+      - name: Build benchmark
         working-directory: benchmark
-        run: |
-          node --run build
-          node run.js
+        run: node --run build
 
-      - name: Convert results to CodSpeed format
-        working-directory: benchmark
-        run: node codspeed-convert.js $BENCH_RESULT
-
-      - name: Upload benchmark results
+      - name: Run benchmarks
         uses: CodSpeedHQ/action@v4
         with:
           mode: walltime
-          run: echo "Results already generated"
+          run: cd benchmark && node run.js && node codspeed-convert.js $BENCH_RESULT


benchmark/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "0.1.0",
+  "type": "module",
+  "description": "Performance benchmark suite for workers-rs",
+  "private": true,
+  "scripts": {
+    "build": "WASM_BINDGEN_BIN=../wasm-bindgen/target/debug/wasm-bindgen ../target/debug/worker-build --release",
+    "bench": "npm run build && node run.js"
+  },
+  "dependencies": {
+    "miniflare": "^4.20250923.0"
```
anonrig: it seems this is extremely old isn't it? can you also enable dependably on this path?

benchmark/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "workers-rs-benchmark",
+  "version": "0.1.0",
+  "type": "module",
+  "description": "Performance benchmark suite for workers-rs",
+  "private": true,
+  "scripts": {
+    "build": "WASM_BINDGEN_BIN=../wasm-bindgen/target/debug/wasm-bindgen ../target/debug/worker-build --release",
+    "bench": "npm run build && node run.js"
```
anonrig:
```suggestion
    "bench": "node --run build && node run.js"
```

benchmark/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+            let total_bytes = body.len() as u64;
+
+            Ok::<u64, String>(total_bytes)
+        };
+
+        tasks.push(task);
+    }
+
+    // Execute all tasks in parallel
+    let start = Date::now().as_millis();
```
anonrig: this seems to be a not good path. can we use rdtsc?

package.json
```diff
@@ -23,6 +23,7 @@
   },
   "scripts": {
     "build": "cd wasm-bindgen && cargo build -p wasm-bindgen-cli --bin wasm-bindgen && cd .. && cargo build -p worker-build",
+    "bench": "cd benchmark && npm run build && npm run bench",
```
anonrig:
```suggestion
    "bench": "cd benchmark && node --run build && node --run bench",
```

benchmark/src/lib.rs
guybedford: Using `Perfrmance` now.

Issue #871, opened 2025-11-15
[BUG] Can't define multiple Durable Objects in same file (again)


radix:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What version of `workers-rs` are you using?

0.6.7

### What version of `wrangler` are you using?

4.45.3

### Describe the bug

Original issue: https://github.com/cloudflare/workers-rs/issues/432
Fix was implemented here, by @kanarus: https://github.com/cloudflare/workers-rs/pull/675
Then, @guybedford caused a regression with this PR: https://github.com/cloudflare/workers-rs/pull/756

Which causes compile errors like this when you have multiple DurableObjects in the same file, because we now have two `use ::worker::DurableObject` in the same scope.

### Steps To Reproduce

1. try to define two structs annotated with #[durable_object] in one file
2. try to compile

This is the output:

```
$ cargo test
   Compiling worker-rust v0.1.0 (/home/radix/Projects/arpeggiorpg/worker)
error[E0252]: the name `DurableObject` is defined multiple times
  --> worker/src/durablegame.rs:52:1
   |
21 | #[durable_object]
   | ----------------- previous import of the trait `DurableObject` here
...
52 | #[durable_object]
   | ^^^^^^^^^^^^^^^^^ `DurableObject` reimported here
   |
   = note: `DurableObject` must be defined only once in the type namespace of this module
   = note: this error originates in the attribute macro `durable_object` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #872, opened 2025-11-17
fix: Properly support multiple durable objects

guybedford:
Follow-on to https://github.com/cloudflare/workers-rs/pull/756, where bringing the trait into the global scope would not allow multiple DOs.

This scopes DurableObject to the macro usage, requiring `DurableObject` to be explicitly imported for the impl block.

Resolves https://github.com/cloudflare/workers-rs/issues/871.

### DIFF
--- a/test/src/alarm.rs
+++ b/test/src/alarm.rs
@@ -1,5 +1,7 @@
 use std::time::Duration;
-use worker::{DurableObject, Env, Request, Response, Result, State, console_log, durable_object, wasm_bindgen};
+use worker::{
+    console_log, durable_object, wasm_bindgen, DurableObject, Env, Request, Response, Result, State,
+};
 
 use super::SomeSharedData;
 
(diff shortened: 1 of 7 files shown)


kanarus:
Hi @guybedford,

I'm the creator of the original PR for the multiple Durable Object issue (PR #675).

#675 originally introduced the #[DurableObject] attribute, which you renamed to #[durable_object].
However, considering the current difficulty around auto-DurableObject-imports with #[durable_object], how about we return to #[DurableObject]?

This would completely resolve the problem, in exchange for an intentional violation of Rust's standard naming convention.

guybedford:
@kanarus I don't think it's a major DX concern to have to import both names into scope to use durable objects. If there are other usability concerns you have with the pattern happy to talk them through. In general sticking to Rust idioms makes sense though, so for that reason I'd rather not go with a camel case naming convention again.

kanarus:
makes sense. Thanks!

PR #883, opened 2025-12-01
deps: update to wasm-bindgen@0.2.106

guybedford:


edwardzcn-decade (commented):
Thx for fixing it! Could we manage a tool to manage the consistency of wasm-bindgen version between worker-build and the submod ref in worker-rs?

guybedford:
worker-build and worker get major versioned together now, but the root cause here is that wasm bindgen makes breaking schema changes on minor releases. I think we need to figure that out upstream.

PR #888, opened 2025-12-15
fix(templates): remove invalid pinned version in generated wrangler.toml build command

Sabourifar:
Templates currently generate:
command = "cargo install -q worker-build@^0.7 && worker-build --release"

Modern versions of Cargo reject the incomplete version specifier `@0.7`, causing: error: invalid value 'worker-build@0.7' ... unexpected end of input while parsing minor version number

Change to unpinned install to use the latest compatible worker-build: command = "cargo install -q worker-build && worker-build --release"

This resolves Cargo syntax errors on fresh Windows (and other) projects without affecting existing functionality.

### DIFF
--- a/templates/hello-world-http/wrangler.toml
+++ b/templates/hello-world-http/wrangler.toml
@@ -3,4 +3,4 @@ main = "build/index.js"
 compatibility_date = "{{ "now" | date: "%Y-%m-%d" }}"
 
 [build]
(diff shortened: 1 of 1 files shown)


guybedford:
I believe it is only `cargo install worker-build@0.7` that doesn't work in Cargo and not `cargo install worker-build@^0.7`.

So perhaps you were hitting a case of `cargo install worker-build@0.7` somewhere instead?

PR #894, opened 2025-12-20
fix(d1): use Function.apply for bind() to fix WASM crash (#893)

### LINKED ISSUE #893: D1 .bind() causes wasm crash: o.__wbindgen_start is not a function
## Description

Using D1's `.bind()` function for INSERT/UPDATE/DELETE operations causes the worker to crash with a wasm-bindgen error.

## Environment

- workers-rs version: 0.7.1
- worker-macros version: 0.7.1
- wrangler version: 4.56.0
- Platform: macOS

## Error

```
TypeError: o.__wbindgen_start is not a function
    at G (index.js:437:286)
    at D (index.js:870:57)
    at Object.construct (index.js:891:5)
```

With log message: `Reinitializing Wasm application`

## Reproduction

```rust
use worker::*;

#[event(fetch, respond_with_errors)]
pub async fn main(req: Request, env: Env, _ctx: Context) -> Result<Response> {
    Router::new()
        .post_async("/test", |_req, ctx| async move {
            let d1 = ctx.env.d1("DB")?;
            
            // This crashes:
            let stmt = d1.prepare("INSERT INTO resources (id, name) VALUES (?1, ?2)");
            let query = stmt.bind(&["id123".into(), "test".into()])?;
            query.run().await?;
            
            Response::ok("done")
        })
        .run(req, env)
        .await
}
```

## What Works

- D1 SELECT queries (with or without `.bind()`)
- D1 INSERT with literal SQL values (no `.bind()`):
  ```rust
  let stmt = d1.prepare("INSERT INTO resources (id, name) VALUES ('id123', 'test')");
  stmt.all().await?;  // Works!
  ```

## What Doesn't Work

- D1 INSERT/UPDATE/DELETE with `.bind()` parameters - crashes worker

## Workaround

Using literal SQL formatting (less secure):
```rust
let sql = format!("INSERT INTO resources (id, name) VALUES ('{}', '{}')", 
    id, name.replace("'", "''"));
let stmt = d1.prepare(&sql);
stmt.all().await?;
```

## Cargo.toml

```toml
[dependencies]
worker = { version = "0.7.1", features = ["d1"] }
worker-macros = { version = "0.7.1" }
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
```

ak9024:
The wasm-bindgen `variadic` attribute wasn't properly spreading array arguments when calling D1's bind method, causing WASM initialization errors for INSERT/UPDATE/DELETE operations.

This fix uses JavaScript's native Function.apply() to properly spread the arguments, bypassing the problematic variadic handling.

- Modified bind() and bind_refs() to use js_sys::Reflect and Function.apply
- Added test case for INSERT/UPDATE/DELETE with raw bind()

Fixes #893

### DIFF
--- a/test/src/d1.rs
+++ b/test/src/d1.rs
@@ -299,3 +299,51 @@ pub async fn retrive_first_none(
 
     Response::ok("ok")
 }
+
+/// Test for GitHub issue #893: D1 .bind() WASM crash with INSERT statements
+/// This test verifies that direct bind() with JsValue works for INSERT/UPDATE/DELETE
+#[worker::send]
+pub async fn insert_with_raw_bind(
+    _req: Request,
+    env: Env,
+    _data: SomeSharedData,
+) -> Result<Response> {
+    let db = env.d1("DB")?;
+
+    // Test INSERT with direct bind() using JsValue (the failing pattern from issue #893)
+    let stmt = db.prepare("INSERT INTO people (id, name, age) VALUES (?1, ?2, ?3)");
+    let query = stmt.bind(&[
+        JsValue::from_f64(1000.0),
+        JsValue::from_str("Test User"),
+        JsValue::from_f64(25.0),
+    ])?;
+    query.run().await?;
+
+    // Verify the insert worked
+    let verify_stmt = worker::query!(&db, "SELECT * FROM people WHERE id = 1000");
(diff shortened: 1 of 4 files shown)


ak9024:
close #893

Issue #891, opened 2025-12-18
[BUG] Unable to build


Shua-github:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What version of `workers-rs` are you using?

0.7.1

### What version of `wrangler` are you using?

4.55.0

### Describe the bug

unable to build
output:
```
PS D:\dev\rust\{name}> npx wrangler dev

 ⛅️ wrangler 4.55.0
───────────────────
[custom build] Running: cargo install -q worker-build@^0.7 && worker-build --release
[custom build] error: invalid value 'worker-build@0.7' for '[CRATE[@<VER>]]...': unexpected end of input while parsing minor version number
[custom build]
[custom build]   tip: if you want to specify SemVer range, add an explicit qualifier, like '^0.7'
[custom build]
[custom build] For more information, try '--help'.
[custom build]
✘ [ERROR] Error: Command failed with exit code 1: cargo install -q worker-build@^0.7 && worker-build --release

  error: invalid value 'worker-build@0.7' for '[CRATE[@<VER>]]...': unexpected end of input while
  parsing minor version number

    tip: if you want to specify SemVer range, add an explicit qualifier, like '^0.7'

```

### Steps To Reproduce

1. 
```bash
cargo generate cloudflare/workers-rs
```
2.
```bash
cd {name}
```
3.
```bash
npx wrangler dev
```

guybedford:
This was resolved in https://github.com/cloudflare/workers-rs/pull/889, the next release will be going out today or tomorrow.

guybedford:
Actually, this should be working already on 0.7.1. Which version of Cargo are you using here?

Shua-github:
> Actually, this should be working already on 0.7.1. Which version of Cargo are you using here?

PS D:\dev\rust\> cargo -V
cargo 1.94.0-nightly (2c283a9a5 2025-12-04)

Shua-github:
> > Actually, this should be working already on 0.7.1. Which version of Cargo are you using here?
> 
> PS D:\dev\rust> cargo -V cargo 1.94.0-nightly (2c283a9a5 2025-12-04)

But I have the same issue using the stable version. After removing `cargo install -q worker-build@^0.7 && `, it reports an error:
```
npx wrangler dev

⛅️ wrangler 4.56.0
───────────────────
[custom build] Running: worker-build --release
[custom build] [INFO]: 🎯 Checking for the Wasm target...
[custom build]
[...]

PR #897, opened 2025-12-22
fix: windows unable to build

Shua-github:
Resolve https://github.com/cloudflare/workers-rs/issues/891

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -9,8 +9,12 @@ permissions:
 
 jobs:
   worker-build:
-    name: Test & Build worker-build
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: ['ubuntu-latest', 'windows-latest']
+
+    name: Build worker-build
+    runs-on: ${{ matrix.platform }}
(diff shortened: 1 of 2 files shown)


guybedford:
This is definitely an oversight, and I'll land this, but it's not clear to me what this has to do with any of the problems reported in #891, and I couldn't find any replicable bug caused by this issue (new Windows CI in https://github.com/cloudflare/workers-rs/pull/898 passes just fine without this PR!).

Shua-github:
> This is definitely an oversight, and I'll land this, but it's not clear to me what this has to do with any of the problems reported in #891, and I couldn't find any replicable bug caused by this issue (new Windows CI in #898 passes just fine without this PR!).

on my computer, it tries to execute the folder, but it worked after I deleted the `.exe`

guybedford:
Strange, the cache function should still return something from within the folder both before and after the change, and I would have expected https://github.com/cloudflare/workers-rs/pull/889 to be the main required Windows fix. Did you definitely validate the local build before and after? I would of course have been happier with a failing test case. Regardless this will make it into the next release and thanks for contributing.

Issue #604, opened 2024-07-28
[Feature] Support incremental backoffs in cloudflare workers


Starttoaster:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Since most of tokio isn't supported (since it does not compile to wasm) I'm not certain if there's a real pattern to support a Cloudflare worker, written in Rust using the workers-rs package, to forward a request to another server using reqwest, with an incremental backoff in the case of a 429 rate limit.

kflansburg:
We do support `reqwest`, `async`,  I think this would be pretty trivial to implement with a loop:

```rust
let mut wait = 1000; // milliseconds
let res = loop {
    let res = reqwest::get(url).await?;
    if res.status() != 429 {
        break res;
    }
    sleep(1000).await
    wait *= 2;
};
```

The only tricky thing is sleeping, but I think you can do something like this:
```rust
pub async fn sleep(delay: i32) {
    let mut cb = |resolve: js_sys::Function, reject: js_sys::Function| {
        web_sys::window()
            .unwrap()
            .set_timeout_with_callback_and_timeout_and_arguments_0(&resolve, delay);
    };

    let p = js_sys::Promise::new(&mut cb);

    wasm_bindgen_futures::JsFuture::from(p).await.unwrap();
}
```

Starttoaster:
Thanks! This does get me a bit closer to something that actually works, by way of this actually compiling and running on `npx run dev`, but I'm getting an issue that I'm maybe not experienced enough with js_sys to understand. I'm running this locally, and trying to run a simple bash script that just slams my local worker with a curl request, one after another (not concurrently.) 

```
[wrangler:inf] POST /test 500 Internal Server Error (99ms)
✘ [ERROR] Uncaught (in response) Error: The script will never generate a response.

✘ [ERROR] A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.
```
And I can tell that not all of the requests actually completed successfully, I believe I'm getting this every time the worker's loop is receiving a 429 from the upstream service.

If this makes sense to you, some help would be appreciated. But I'm still looking in case I can figure it out myself..

kflansburg:
This is reported by the Workers runtime when all futures have executed but not response has returned, so I suspect there is something slightly wrong with my hacky `sleep` method. 

[...]

PR #901, opened 2026-01-06
[Draft] Idea: add tokio-workers compatibility shim crate

irvinebroque:
## Summary

- Add `tokio-workers` crate providing tokio-compatible APIs that work on Cloudflare Workers
- Users can swap `use tokio::*` with `use tokio_workers::*` to run tokio-based code on Workers
- Add `examples/tokio-workers-demo` demonstrating all supported APIs

## Supported APIs

| Tokio API | tokio-workers Implementation |
|-----------|------------------------------|
| `spawn`, `spawn_local` | `wasm_bindgen_futures::spawn_local` with custom `JoinHandle` |
| `time::sleep` | JavaScript `setTimeout` |
| `time::timeout` | Race sleep against future |
| `sync::*` | Re-exported from tokio (already runtime-agnostic) |
| `join!`, `try_join!`, `select!` | Re-exported from futures-util |
| `spawn_blocking` | Feature-gated: `spawn-blocking-panic` or `spawn-blocking-sync` |

## Example Usage

```rust
// Before (won't work on Workers):
use tokio::{spawn, time::sleep, sync::mpsc};

// After (works on Workers!):
use tokio_workers::{spawn, time::sleep, sync::mpsc};
```

## Testing

All 18 unit tests pass, plus the demo example has 7 endpoints that all work correctly:

- `/spawn` - Concurrent task spawning (~11ms for 3x10ms tasks)
- `/sleep` - Timer works as expected
- `/timeout` - Fast ops succeed, slow ops timeout
- `/channels` - mpsc channels work with multiple producers
- `/mutex` - Async mutex correctly synchronizes
- `/concurrent` - `join!` runs operations in parallel (~30ms not 90ms)
- `/spawn-blocking` - Sync computation works

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2503,6 +2503,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "tokio-workers"
+version = "0.1.0"
+dependencies = [
+ "futures-util",
+ "js-sys",
+ "pin-project",
+ "tokio",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "wasm-bindgen-test",
+ "web-sys",
+]
+
+[[package]]
+name = "tokio-workers-demo"
+version = "0.1.0"
+dependencies = [
+ "js-sys",
+ "serde",
+ "serde_json",
+ "tokio-workers",
+ "worker 0.7.2",
+]
+
+[[package]]
 name = "toml"
 version = "0.8.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "worker-macros",
     "test",
     "worker-sys",
+    "tokio-workers",
     "examples/*",
     "test/container-echo",
 ]
--- a/examples/tokio-workers-demo/Cargo.toml
+++ b/examples/tokio-workers-demo/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "tokio-workers-demo"
+version = "0.1.0"
+edition = "2021"
+
+[package.metadata.release]
+release = false
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+worker.workspace = true
+tokio-workers = { path = "../../tokio-workers", features = ["spawn-blocking-sync"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+js-sys.workspace = true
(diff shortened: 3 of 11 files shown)


irvinebroque:
@danlapid @kflansburg @guybedford is something of this flavor a terrible idea? _Some_ flavor of drop-in shim for Tokio things, to make it easier for someone to take an app built for Tokio, drop in a replacement library for equivalent behavior, and then get up and running? Ex:

```rs
// Before (won't work on Workers):
use tokio::{spawn, time::sleep, sync::mpsc};

// After (works on Workers!):
use tokio_workers::{spawn, time::sleep, sync::mpsc};
```

Point of PR is to show idea not to land anything. Came out of conversation where smelled like someone had code that _probably_ would work with `workers-rs` but came across as too tricky.

Just seems like at minimum there is some guidance what concretely you can/can't use from Tokio, how you might go about adapting, etc.

We talk a bit about that here: https://developers.cloudflare.com/workers/languages/rust/crates/

But there is more to say probably? Or more guidance? Or an agent that provides guidance to give people?

danlapid:
What's the benefit of diverging from the ecosystem here?
What is workers specific about this crate?

tokio already has built in wasm support - https://docs.rs/tokio/latest/tokio/#wasm-support

time works on wasm32-wasi but that will be solved by our work to enable a wasm-bindgen target backed by emscripten.

FWIW Ingvar had an experiment to enable wasm32-wasi with a js ffi middleware to implement wasi, IIRC we didn't go with that because wasi actually had worst crate compatibility than wasm32-unknown

In the meantime, there's https://github.com/cunarist/tokio-with-wasm which I guess is similar to what you did here?

kflansburg:
I do think that `sleep` and having to use `wasm_bindgen_futures::spawn_local` is kind annoying for users ([see](https://github.com/cloudflare/workers-rs/issues/604)). I think there are a number of things from Tokio though that do just work out of the box as @danlapid mentioned, and I'm not sure re-exporting them is valuable.

guybedford:
This was really inspiring, thank you for posting it.

My conclusion like @danlapid is that this belongs in upstream Tokio, and the question is simply how to get there.

It probably makes the most sense to support this under an emscripten target, and aligning Wasm Bindgen with the emscripten target, as it defines an actual execution machine.

As soon as we have the Wasm Bindgen emscripten target support (hopefully coming in the next few months), I will post a Tokio PR to support these features upstream on the combined Wasm Bindgen test suite already available for Tokio.

If we can achieve such a direction we align platform support between Python and Rust workers as well, it will just take a few more months to come together first I think.

PR #904, opened 2026-01-14
feat: add CacheMode to RequestInit

danlapid:
Add support for the `cache` option in RequestInit, allowing control over how requests interact with the HTTP cache. Supports NoStore, NoCache, and Reload modes matching the Cloudflare Workers runtime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -163,9 +163,9 @@ dependencies = [
 
 [[package]]
 name = "axum-core"
-version = "0.5.5"
+version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "59446ce19cd142f8833f856eb31f3eb097812d1479ab224f54d72428ca21ea22"
+checksum = "08c78f31d7b1291f7ee735c1c6780ccde7785daae9a9206026862dab7d8792d1"
 dependencies = [
  "bytes",
  "futures-core",
@@ -333,9 +333,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.2.50"
+version = "1.2.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9f50d563227a1c37cc0a263f64eca3334388c01c5e4c4861a9def205c614383c"
+checksum = "cd4932aefd12402b36c60956a4fe0035421f544799057659ff86f923657aada3"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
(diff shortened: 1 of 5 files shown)


worker/src/request_init.rs
```diff
@@ -538,3 +549,33 @@ impl From<RequestRedirect> for web_sys::RequestRedirect {
...
+
+impl From<CacheMode> for web_sys::RequestCache {
+    fn from(mode: CacheMode) -> Self {
+        match mode {
+            CacheMode::NoStore => web_sys::RequestCache::NoStore,
+            CacheMode::NoCache => web_sys::RequestCache::NoCache,
+            CacheMode::Reload => web_sys::RequestCache::Reload,
+        }
+    }
+}
```
guybedford, line 570: This is already defined as `RequestCache` - https://docs.rs/crate/web-sys/0.3.83/source/src/features/gen_RequestCache.rs.

Would be nice to use the web-sys definition directly if we can.

guybedford: Ahh, because we don't support `force-cache` and `only-if-cached`, that makes sense.

danlapid: Yeah that's why I cloned it

danlapid: But I made it convertible

worker/src/request_init.rs
```diff
@@ -538,3 +549,33 @@ impl From<RequestRedirect> for web_sys::RequestRedirect {
         }
     }
 }
+
+/// Cache mode for controlling how requests interact with the cache.
+/// Corresponds to JavaScript's `RequestInit.cache` property.
+#[derive(Clone, Copy, Debug, Serialize)]
+#[serde(rename_all = "kebab-case")]
```
guybedford: We don't usually derive Serialize for all types, is there a reason for needing this?

worker/src/request_init.rs
```diff
@@ -538,3 +549,33 @@ impl From<RequestRedirect> for web_sys::RequestRedirect {
...
+
+impl From<CacheMode> for &str {
+    fn from(mode: CacheMode) -> Self {
+        match mode {
+            CacheMode::NoStore => "no-store",
+            CacheMode::NoCache => "no-cache",
+            CacheMode::Reload => "reload",
+        }
+    }
+}
```
guybedford: What's the use case for this one? Again, not typically implemented for string enums in Wasm Bindgen, which is what this is representing. We usually implement `From<CacheMode> for JsValue` though to convert things into `JsValue` in generic APIs. With the generics work this gets superceded now by `upcast()` though.

worker/src/request_init.rs
```diff
@@ -538,3 +549,33 @@ impl From<RequestRedirect> for web_sys::RequestRedirect {
         }
     }
 }
+
+/// Cache mode for controlling how requests interact with the cache.
+/// Corresponds to JavaScript's `RequestInit.cache` property.
+#[derive(Clone, Copy, Debug, Serialize)]
```
guybedford: It can be worth implementing `Eq` and `PartialEq` as well, we do that for all web-sys types.

guybedford (commented):
Posted some suggestions, let me know if you're okay with me making those changes to land.

guybedford (approved):

PR #911, opened 2026-01-16
Reproduce bug

danlapid:
This should work but seems to fail by hitting
https://github.com/cloudflare/workerd/blob/2286f2a2c7c50ad82caf199825af101a863af9e0/src/workerd/api/http.c++#L753


danlapid:
Found out this is a workerd bug

PR #913, opened 2026-01-20
fixed injecting default headers and cf options

### LINKED ISSUE #865: [BUG] worker::RequestInit, it's always error 'invalid URL or   options for Request'.

0.6

4.45.3

### Describe the bug

use request_init always faield. run the code.

any way catch js err?

### Steps To Reproduce

 ```rust
// work
Request::new("https://httpbin.org/post", worker::Method::Post).unwrap();  


// not work  called `Result::unwrap()` on an `Err` value: JsError("invalid URL or options for Request")
let mut request_init = RequestInit::new();
request_init.method = worker::Method::Post;
Request::new_with_init("https://httpbin.org/post", &request_init).unwrap();
```

SATVIKsynopsis:
Fixes #865

Fixes the issue where `RequestInit` could inject invalid default headers and cf options, causing `Request::new_with_init` to fail even for minimal requests.

Only set headers when non empty and only attach cf when non default.  and also added a test for `RequestInit`.

### DIFF
--- a/worker/src/request_init.rs
+++ b/worker/src/request_init.rs
@@ -83,16 +83,16 @@ impl From<&RequestInit> for web_sys::RequestInit {
         // set the Cloudflare-specific `cf` property on FFI RequestInit
         if !req.cf.is_default() {
             let r = ::js_sys::Reflect::set(
-            inner.as_ref(),
-            &JsValue::from("cf"),
-            &JsValue::from(&req.cf),
-        );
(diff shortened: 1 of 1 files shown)


guybedford (approved):

SATVIKsynopsis:
Thanks a lot for the review and merge.
Appreciate the guidance and looking forward to contributing more to workers-rs.

Issue #663, opened 2024-10-25
Workflows support


lemarier:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Hi guys,

Would be great if we could trigger [workflows](https://developers.cloudflare.com/workflows/) from workers-rs 

Would be even better if we could build workflows with workers-rs at some point as well.

Thanks!

sidharthachatterjee:
Thank you for opening this. It's on my mind. We're working on prioritising this after we hit some ground for stability. We'll most likely ship this for GA.

jeastham1993:
@sidharthachatterjee is there any update on this out of interest?

PeterMHammond:
I just built an entire Workflow in workers-rs using standard Durable Object just because it's not available.  
(It is really difficult to get right).  Argued with Gemini for a long time about it.
```
As of the current state of workers-rs and Cloudflare's documentation, there is no direct equivalent to the WorkflowEntrypoint and WorkflowStep API from the TypeScript Cloudflare Workflows within workers-rs.

The way Cloudflare Workflows are defined and executed with the WorkflowEntrypoint class and the step API is a higher-level abstraction that is primarily available in the JavaScript/TypeScript environment of Workers.
```
`In conclusion, while you can indeed use wasm-bindgen to manually create bindings to JavaScript APIs within the Worker environment, directly replicating the WorkflowEntrypoint and WorkflowStep abstraction from TypeScript might be very challenging due to the internal nature of how Workflows are executed. It's more likely that you could use wasm-bindgen to interact with lower-level Workflows management functions (if exposed) or continue to build your workflow logic using Durable Objects as demonstrated earlier.`

Which is what I'm doing and this is a real pain to recreate the OOB features of what Workflows offer.
I'm not going back to JavaScript/TypeScript if at all possible.
Looking forward to RUST being a first class citizen in the Cloudflare architecture.

spigaz:
@PeterMHammond I'm not sure if I have already explained this before to you.

But its not that hard to extend the missing functionality as workers-rs are just a shim over the js/ts implementation.

Most likely if you point Gemini to the documentation and the workers-rs repo, it will guide you in the process.

What might be harder is to get some feedback from Cloudflare.

I'm not sure why, but as people aren't organised and keep moving away, workers-rs is pretty much on hold in practice.

PeterMHammond:
[...]

PR #918, opened 2026-01-28
feat: add Workflows support

connyay:
Add bindings for Workflows including step primitives (do, sleep, sleepUntil), event/instance management, and a proc macro for defining workflow entrypoints. Includes example worker, sys-level bindings, and integration tests.


connyay:
Fixes #663 👏

guybedford:
Amazing, I will take a look soon. Hopefully we can get this in for the next release!

worker/src/workflow.rs
```diff
@@ -0,0 +1,538 @@
...
+
+impl From<std::time::Duration> for WorkflowDuration {
+    fn from(d: std::time::Duration) -> Self {
+        Self::Milliseconds(d.as_millis() as u64)
+    }
+}
+
+/// Error type for non-retryable workflow errors.
+#[derive(Debug)]
+pub struct NonRetryableError {
```
guybedford, line 1120: For this to actually be non-retryable I believe it must be an instance of `NonRetryableError`.

Therefore this would need to be:

```
#[wasm_bindgen(module = "cloudflare:workflows")]
extern "C" {
    #[wasm_bindgen(extends = js_sys::Error)]
    #[derive(Debug, Clone)]
    pub type NonRetryableErrorSys;
    #[wasm_bindgen(constructor, js_class = "NonRetryableError")]
    pub fn new(message: &str) -> NonRetryableErrorSys;
}
```

And the esbuild in worker-build would need to be updated with the `--external cloudflare:workflows` external import.

worker-sys/src/types/workflow.rs
```diff
@@ -0,0 +1,87 @@
...
+        callback: &js_sys::Function,
+    ) -> Result<js_sys::Promise, JsValue>;
+
+    #[wasm_bindgen(method, catch, js_name = "do")]
+    pub fn do_with_config(
+        this: &WorkflowStep,
+        name: &str,
+        config: JsValue,
+        callback: &js_sys::Function,
+    ) -> Result<js_sys::Promise, JsValue>;
```
guybedford: Why not type these as async with return typing?

guybedford (commented):
Left some initial comments. Let me know your thoughts.

worker/src/workflow.rs
connyay: great catch. updated and added some code paths that stress it in the example

worker-sys/src/types/workflow.rs
connyay: good call. cleaned it up quite a bit

test/src/durable.rs
```diff
@@ -23,8 +23,6 @@ pub struct QueryParams {
 
 impl DurableObject for MyClass {
     fn new(state: State, _env: Env) -> Self {
-        // Unfortunately we can't access the `name` property within the Durable Object (see <https://github.com/cloudflare/workerd/issues/2240>). Instead, we can pass it as a request parameter.
```
connyay: plucked to https://github.com/cloudflare/workers-rs/pull/959

examples/workflow/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+impl WorkflowEntrypoint for MyWorkflow {
+    fn new(_ctx: Context, env: Env) -> Self {
+        Self { env }
+    }
+
+    async fn run(
+        &self,
+        event: WorkflowEvent<serde_json::Value>,
```
guybedford: Can we come up with a better typing solution here? It is quite restrictive to rely so heavily on serde. Can we instead make this generic with the bound being something something as simple as IntoWasmAbi?

worker/src/workflow.rs
```diff
@@ -0,0 +1,532 @@
...
+
+impl EnvBinding for Workflow {
+    const TYPE_NAME: &'static str = "Workflow";
+
+    fn get(val: JsValue) -> Result<Self> {
+        let obj = Object::from(val);
+        let constructor_name = obj.constructor().name();
+        if constructor_name == Self::TYPE_NAME
+            || constructor_name == "WorkflowImpl"
+            || constructor_name == "Fetcher"
```
guybedford: Why must we support `Fetcher` here as well?

worker/src/workflow.rs
```diff
@@ -0,0 +1,532 @@
...
+
+impl WorkflowInstance {
+    fn from_js(val: JsValue) -> Self {
+        Self {
+            inner: val.unchecked_into(),
+        }
+    }
+
+    /// The unique ID of this workflow instance.
+    pub fn id(&self) -> Result<String> {
```
guybedford: When can this fail? Does it need to be Result?

worker/src/workflow.rs
```diff
@@ -0,0 +1,532 @@
...
+    Constant,
+    Linear,
+    Exponential,
+}
+
+/// Options for waiting for an external event.
+#[derive(Debug, Clone, Serialize)]
+pub struct WaitForEventOptions {
+    #[serde(rename = "type")]
+    pub event_type: String,
```
guybedford: Note that `type_` is how we usually handle keyword renaming.

worker/src/workflow.rs
```diff
@@ -0,0 +1,532 @@
...
+}
+
+/// Retention policy for workflow instances.
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RetentionOptions {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub success_retention: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub error_retention: Option<String>,
```
guybedford: Could we support a `WorkflowSleepDuration` type say via `WorkflowSleepDuration::new(5, WorkflowDuration::Seconds)`, which just wraps a JsString underneath? Might be easier to do if `WorkflowSleepDuration` is an imported type wrapper itself extended with a custom impl block.

guybedford (commented):
This is very close to a perfect implementation, some small suggestions. Probably worth carefully working through this before landing so I won't rush it out in todays release actually.

examples/workflow/src/lib.rs
connyay: Does JsValue work here?

worker/src/workflow.rs
connyay: I think it was something weird with miniflare, but working fine now without it

worker/src/workflow.rs
connyay: love it. I used this throughout. 

it is a bit awkward in a few spots. places like `pub timeout: Option<WorkflowSleepDuration>,` or `pub delay: WorkflowSleepDuration,`

Almost added some aliases, but not sure if it is worth the api surface area. let me know what you think
```rust
pub type WorkflowDelayDuration = WorkflowSleepDuration;
pub type WorkflowTimeoutDuration = WorkflowSleepDuration;
pub type WorkflowRetentionDuration = WorkflowSleepDuration;
```

worker/src/workflow.rs
connyay: 🤷 I would be surprised if this failed. removed result and added an expect

examples/workflow/src/lib.rs
guybedford: What I mean specifically is making this a generic argument `T` so that `run<T>` can take any type with the IntoWasmAbi trait from wasm bindgen?

guybedford: Actually with the Wasm Bindgen generics you could just make `T: JsGeneric` on the new `JsGeneric` trait.

connyay: `JsGeneric` is a bit awkward to work with here.

```rust
/// The event passed to a workflow's run method.
#[derive(Debug, Clone)]
pub struct WorkflowEvent<T = JsValue> {
    pub payload: T,
    pub timestamp: crate::Date,
    pub instance_id: String,
}

impl<T: JsGeneric> WorkflowEvent<T> {
    pub fn from_js(value: JsValue) -> Result<Self> {
        let payload = get_property(&value, "payload")?;
        Ok(Self {
            // SAFETY: JsGeneric guarantees T has ErasableGeneric<Repr = JsValue>,
            // so T and JsValue have an identical memory layout.
            payload: unsafe {
                core::mem::transmute_copy(&core::mem::ManuallyDrop::new(payload))
            },
            timestamp: get_timestamp_property(&value, "timestamp")?,
            instance_id: get_string_property(&value, "instanceId")?,
        })
    }
}
```

`JsCast` is slightly less awkward:

```rust
/// The event passed to a workflow's run method.
#[derive(Debug, Clone)]
pub struct WorkflowEvent<T = JsValue> {
    pub payload: T,
    pub timestamp: crate::Date,
    pub instance_id: String,
}

impl<T: JsCast> WorkflowEvent<T> {
    pub fn from_js(value: JsValue) -> Result<Self> {
        Ok(Self {
            payload: get_property(&value, "payload")?
                .dyn_into()
                .map_err(|_| crate::Error::JsError("payload is not the expected type".into()))?,
            timestamp: get_timestamp_property(&value, "timestamp")?,
            instance_id: get_string_property(&value, "instanceId")?,
        })
    }
}
```

thoughts/opinions? or am I missing something/doing something wrong with JsGeneric?

guybedford: I likely need to properly follow the API design details here, I will aim to do a thorough review next week after this current release. Sorry for the delay and thanks for the patience as always!

connyay: All good! Thank you!

connyay: I dog fooded this branch a bit more in another project and 30f2d4e2a21774f592dc7f911e41eb4bce1a3cd9 is feeling pretty dang ergonomic to me. I know this doubles down on serde, when you were trying to not rely so heavily on it :-/ 

I'm still happy to try other directions.

sam-rusty:
Just checking in on this PR. Is anyone currently reviewing or working on it?

iamralch:
@guybedford any plans this PR to be accepted

PR #920, opened 2026-01-29
Implement `run_bytes` method on AI binding for models that return binary data.

### LINKED ISSUE #783: [BUG] AI binding fails to handle binary responses from image generation  models

0.6.0

4.25.0

### Describe the bug

The `worker::Ai` binding cannot properly handle binary responses from AI image
  generation models like Stable Diffusion. When calling `ai.run()` with an image
  generation model, the binary image data gets lost and converted to an empty object
  `{}`.

  ## Current Behavior

  When using the AI binding to generate images, the response is always an empty
  object instead of the actual image bytes:

  ```rust
  let ai = env.ai("AI")?;
  let request = ImageGenRequest {
      prompt: "a beautiful sunset".to_string(),
  };

  // This returns an empty object {} instead of image bytes
  let response: serde_json::Value =
  ai.run("@cf/stabilityai/stable-diffusion-xl-base-1.0", &request).await?;
  // response is Object({}) - the actual image data is lost

  Expected Behavior

  The AI binding should return the raw image bytes. The JavaScript version works
  correctly:

  const imageResponse = await
  env.AI.run("@cf/stabilityai/stable-diffusion-xl-base-1.0", { prompt });
  // imageResponse contains the actual PNG image data
  return new Response(imageResponse, {
      headers: { "content-type": "image/png" }
  });

### Steps To Reproduce

Reproduction

  Minimal example that demonstrates the issue:

  use worker::*;
  use serde::{Deserialize, Serialize};

  #[derive(Serialize)]
  struct ImageGenRequest {
      prompt: String,
  }

  #[event(fetch)]
  async fn fetch(_req: Request, env: Env, _ctx: Context) -> Result<Response> {
      let ai = env.ai("AI")?;

      let request = ImageGenRequest {
          prompt: "test image".to_string(),
      };

      // This fails - returns empty object instead of image bytes
      let response: serde_json::Value =
  ai.run("@cf/stabilityai/stable-diffusion-xl-base-1.0", &request).await?;

      console_log!("Response: {:?}", response); // Logs: Response: Object({})

      Response::error("Image data lost", 500)
  }

Impact

  This makes it impossible to use Rust workers for AI image generation, speech
  synthesis, or any other AI models that return binary data. This is a significant
  limitation for developers wanting to use Rust for AI-powered applications on
  Cloudflare Workers.

  Console Output

  Polished prompt: <prompt text>
  Response from AI: Object({})
  Error generating image: Error: invalid type: JsValue(Object({})), expected a
  sequence

  The error shows the worker is trying to deserialize to Vec<u8> but receives an
  empty object instead.
  ```

kflansburg:
The method returns a `ByteStream` which implements `Stream`.

This can be passed directly to the response or collected to `Vec<u8>`, as shown in the docstring.

Also added an example worker to demonstrate / test this behavior.

Closes #783

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,7 +82,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -93,7 +93,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -493,7 +493,7 @@ dependencies = [
  "cookie",
  "document-features",
  "idna",
- "indexmap 2.13.0",
+ "indexmap",
  "log",
  "serde",
  "serde_derive",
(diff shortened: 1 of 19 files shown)


guybedford:
Amazing, thanks for this. I'll make sure to review it for the next release hopefully next week.

PR #926, opened 2026-02-07
Added example: Using serde to populate structures from an API

noobosapien:
This is a demo of using serde Serialize and Deserialize to populate structs from JSON from an API and return a response from that struct serialised as JSON. It also uses http:Request instead of worker::Request.
It is useful to get values tied to Rust variables from an API.
This fills a gap in the current examples, where using external APIs to get values and manipulate them.

### DIFF
--- a/examples/serde-rust-structs/Cargo.toml
+++ b/examples/serde-rust-structs/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "serde-rust-structs"
+version = "0.1.0"
+edition = "2021"
+
+[package.metadata.release]
+release = false
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+worker = { version = "0.7", features = ['http'] }
+worker-macros = { version = "0.7", features = ['http'] }
+http = "1.3"
+serde = "1.0.228"
+serde_json = "1.0.149"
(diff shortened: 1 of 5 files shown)


examples/serde-rust-structs/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[package.metadata.release]
+release = false
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+worker = { version = "0.7", features = ['http'] }
+worker-macros = { version = "0.7", features = ['http'] }
```
jakubadamw, line 14: This is an unused dependency. 🙂

examples/serde-rust-structs/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+http = "1.3"
+serde = "1.0.228"
+serde_json = "1.0.149"
```
jakubadamw: This is an unused dependency. This is important in so far as using `serde_json` in code compiled to Wasm and deployed to Cloudflare Workers is normally discouraged in favour of [`serde-wasm-bindgen`](https://crates.io/crates/serde-wasm-bindgen) and using the native `JSON.*()` functions (through, say, js-sys) as that reduces final code bloat. But anyway, your example doesn't actually use `serde_json` in any way, so this can be removed.

examples/serde-rust-structs/Cargo.toml
noobosapien: Will remove it soon!

examples/serde-rust-structs/Cargo.toml
noobosapien: Thanks for letting me know.

PR #929, opened 2026-02-07
fix: use platform-native TLS for binary downloads to support corporate proxies

### LINKED ISSUE #928: [BUG] worker-build binary downloads fail with UnknownIssuer on systems with TLS-inspecting proxies

0.7.4

4.63.0

### Describe the bug

`worker-build` fails to download `wasm-bindgen` (and potentially `esbuild`/`wasm-opt`) on systems where HTTPS traffic is inspected by a corporate TLS proxy (e.g., Netskope).

The `download()` function in `worker-build/src/binary.rs` uses `ureq::get(url).call()`, which relies on ureq's default TLS configuration: `rustls` with `RootCerts::WebPki`. This uses a statically compiled bundle of Mozilla root CAs (`webpki-roots` crate) and completely ignores the operating system's certificate trust store.

On systems with TLS inspection, the proxy re-signs all HTTPS traffic with a custom CA certificate that is trusted by the OS (installed in the macOS system keychain, Windows certificate store, etc.) but is **not** in the `webpki-roots` bundle. This causes `rustls` to reject the proxy's certificate with `UnknownIssuer`, making all binary downloads fail:

```
[INFO]: ⬇️ Downloading Wasm Bindgen@0.2.106...
Error: Failed to fetch URL https://github.com/wasm-bindgen/wasm-bindgen/releases/download/0.2.106/wasm-bindgen-0.2.106-aarch64-apple-darwin.tar.gz
Caused by:
io: invalid peer certificate: UnknownIssuer
```

`curl` works fine because they use the OS trust store.

**Expected behavior:** `worker-build` should use the platform's native TLS certificate verification (or at minimum respect the OS certificate store) so that downloads work in corporate environments with TLS inspection.

**Proposed fix:** Enable the `native-tls` feature on `ureq` and configure the `download()` function to use `TlsProvider::NativeTls` with `RootCerts::PlatformVerifier`, which delegates certificate verification to the OS:

### Steps To Reproduce

_No response_

aquilacf:
Binary downloads (wasm-bindgen, esbuild, wasm-opt) fail with `invalid peer certificate: UnknownIssuer` in environments where HTTPS traffic is intercepted by a corporate security proxy (e.g. Netskope).

The root cause is that `ureq` defaults to `rustls` with `webpki-roots`, a static bundle of Mozilla root CAs compiled into the binary. When a corporate proxy re-signs TLS certificates with its own CA (installed in the OS keychain but absent from the Mozilla bundle), `rustls` rejects the certificate chain.

### Changes

- Added `native-tls` feature to the `ureq` dependency in `worker-build/Cargo.toml`
- Changed the `download()` function in `binary.rs` to create a `ureq::Agent` configured with `TlsProvider::NativeTls` and `RootCerts::PlatformVerifier`, delegating certificate verification to the OS (SecureTransport on macOS, SChannel on Windows, OpenSSL on Linux)

This ensures downloads respect the system certificate store, including any corporate or custom root CAs, while remaining transparent to users not behind a proxy.

Fixes #928

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,9 +592,9 @@ checksum = "26bf8fc351c5ed29b5c2f0cbbac1b209b74f60ecd62e675a998df72c49af5204"
 
 [[package]]
 name = "deranged"
-version = "0.5.5"
+version = "0.3.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ececcb659e7ba858fb4f10388c250a7252eb0a27373f1a72b8748afdd248e587"
+checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
 dependencies = [
  "powerfmt",
 ]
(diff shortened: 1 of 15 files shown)


guybedford (approved):
Thanks for posting the fix.

aquilacf:
Thank you!

PR #930, opened 2026-02-19
Minor R2/Hyperdrive/Service Bindings related changes

fisherdarling:
Hey! We've been running a fork of workers-rs internally for 18 months and realized we should probably upgrade and upstream our changes. These should be the relevant changes that aren't upstreamed yet.

1) we remove a debug log for `Delay` executed every time it's dropped. That ends up adding a lot of noise to our WOBS logs.
2) then we fix HTTP version conversion failing for service bindings. We found that if we don't provide a version fallback, incoming requests over a service binding fail.
3) we also added Hyperdrive's `connect` method.
4) For R2 we  we added an undocumented R2 list option, `startAfter` and conditional inserts to the `PutOptionsBuilder`. This is a BREAKING change when executing PUTs, `Result<Object>` -> `Result<Option<Object>>`.

### DIFF
--- a/worker-sys/src/types/hyperdrive.rs
+++ b/worker-sys/src/types/hyperdrive.rs
@@ -8,7 +8,7 @@ extern "C" {
     #[derive(Debug, Clone, PartialEq, Eq)]
     pub type Hyperdrive;
 
-    #[wasm_bindgen(method, getter, js_name=connectionString, catch)]
+    #[wasm_bindgen(method, catch)]
     pub fn connect(this: &Hyperdrive) -> Result<Socket, JsValue>;
 
     #[wasm_bindgen(method, getter, js_name=connectionString)]


guybedford:
Thanks for sharing this! Unfortuantely I didn't have time to fully review for this release, but will make sure to include it with the next one.

worker-sys/src/types/hyperdrive.rs
```diff
@@ -1,11 +1,16 @@
...
 
+use crate::Socket;
+
 #[wasm_bindgen]
 extern "C" {
     #[wasm_bindgen(extends=js_sys::Object)]
     #[derive(Debug, Clone, PartialEq, Eq)]
     pub type Hyperdrive;
 
+    #[wasm_bindgen(method, getter, js_name=connectionString, catch)]
```
connyay:
```suggestion
    #[wasm_bindgen(method, catch)]
```

PR #934, opened 2026-02-24
fix: use staging directory with build lock to prevent concurrent build conflicts

### LINKED ISSUE #927: [BUG] wrangler crashed on code changes, but only on consecutive build

0.7.4

4.63.0

### Describe the bug

when trying the cargo generate template following the tutorial, running npx wrangler have some weird behaviour, it crashed after the build process and complains about unexpected output

my suspect is that the worker-build command produces some stdout that are not expected by wrangler and while the command is stdout without stderr, it's picked and stopped the process

### Steps To Reproduce

1. run cargo generate to download the template
2. run npx wrangler dev
3. the first pass is ok
4. made some code changes
5. wrangler rebuild, and it crashed
6. also tried disabling wasm-opt via Cargo.toml, or supply wasm-opt and wasm-bindgen from env-var, problem still persist

relevant part of the log, it said something about wasm-opt, but the code compiled on running wrangler dev, i think the logic on checking for a rebuild is different on first try

```
      'warning: `test-cf-axum` (lib) generated 2 warnings\n' +
      '    Finished `release` profile [optimized] target(s) in 3.89s\n' +
      '[INFO]: Optimizing wasm binaries with `wasm-opt`...\n' +
      'Error: No such file or directory (os error 2)\n' +
      'To disable `wasm-opt`, add `wasm-opt = false` to your package metadata in your `Cargo.toml`.',
```

---

current workaround to suppress this
 
1. I try to hide the stdout and only redirect it if it failed
2. change wrangler.toml
```
[build]
command = "out=$(cargo install -q worker-build@^0.7 && worker-build --release 2>&1) || echo '$out'"
```
3. any further code changes only forwarded to warngler if worker-build can't build the code

guybedford:
## Summary

Fixes #927

When `wrangler dev` triggers overlapping `worker-build` processes (e.g. a rebuild starts while the previous one is still running), they race on the `build/` directory — one process's `remove_dir_all` can delete files the other is still working with, causing `No such file or directory (os error 2)` during the wasm-opt step.

## Changes

- **New `build_lock.rs`**: Introduces `BuildLock`, which uses a `build/.tmp/` staging directory with an mtime-based heartbeat as a lightweight lock:
  - A background thread touches `.tmp` mtime every 1 second
  - On startup, if `.tmp` exists with mtime < 5s, another build is active — waits for it to go stale
  - `finish()` moves entries from `.tmp/` into `build/` and cleans up

- **Modified `main.rs`**: Replaces the `remove_dir_all(build/) + build directly` pattern with:
  1. `BuildLock::acquire()` — wait for concurrent builds, create fresh `.tmp/`
  2. All build steps write into `.tmp/`
  3. `lock.finish()` — swap entries into `build/`, remove `.tmp/`

- **New dependency**: `filetime = "0.2"` for portable mtime manipulation

## How it works

```
Process A                          Process B
─────────                          ─────────
acquire() → creates build/.tmp/
heartbeat: touch .tmp every 1s
building into .tmp/ ...            acquire() → sees .tmp mtime < 5s
                                   waits...
finish() → moves .tmp/* → build/
heartbeat stops                    sees .tmp stale (>5s) or gone
                                   creates fresh .tmp/, proceeds
```

### DIFF
--- a/worker-build/src/build_lock.rs
+++ b/worker-build/src/build_lock.rs
@@ -55,7 +55,8 @@ impl BuildLock {
         let oldtmp_dir = out_dir.join(".oldtmp");
 
         // Ensure the parent out_dir exists
-        fs::create_dir_all(out_dir)?;
+        fs::create_dir_all(out_dir)
+            .with_context(|| format!("Failed to create output directory {}", out_dir.display()))?;
 
         // Wait for any active build to finish
         Self::wait_for_stale(&tmp_dir)?;
@@ -77,7 +78,8 @@ impl BuildLock {
         }
 
         // Create fresh .tmp
-        fs::create_dir_all(&tmp_dir)?;
+        fs::create_dir_all(&tmp_dir)
+            .with_context(|| format!("Failed to create staging directory {}", tmp_dir.display()))?;
 
         // Start heartbeat
         let stop = Arc::new(AtomicBool::new(false));
@@ -115,7 +117,12 @@ impl BuildLock {
         self.stop_heartbeat();
 
         // Move each entry from .tmp/ into out_dir/
-        for entry in fs::read_dir(&self.tmp_dir)? {
+        for entry in fs::read_dir(&self.tmp_dir).with_context(|| {
+            format!(
+                "Failed to read staging directory {}",
+                self.tmp_dir.display()
+            )
+        })? {
             let entry = entry?;
             let name = entry.file_name();
             let dest = self.out_dir.join(&name);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3963,6 +3963,7 @@ dependencies = [
  "console",
  "dirs-next",
  "env_logger",
+ "filetime",
  "flate2",
  "glob",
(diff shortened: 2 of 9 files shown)


worker-build/src/build_lock.rs
```diff
@@ -0,0 +1,185 @@
...
+            fs::rename(entry.path(), &dest).with_context(|| {
+                format!(
+                    "Failed to move {} from staging to output",
+                    name.to_string_lossy()
+                )
+            })?;
+        }
+
+        // Remove the now-empty .tmp directory
+        let _ = fs::remove_dir_all(&self.tmp_dir);
```
jakubadamw: Seeing it is expected that the directory is empty by now (as it is), then I would suggest using `fs::remove_dir()` to make it clearer. Also, I wonder if the silencing of the errors here and above is desired? At the very least, they could be logged?

guybedford: Good suggestion, added.

PR #935, opened 2026-02-24
Fix a few issues in worker-build

### LINKED ISSUE #739: [BUG] `worker-build -h` is confusing, because it prints help text for `wasm-pack build` instead

n/a

n/a

### Describe the bug

with `worker-build` 0.1.2 installed:

```plaintext
$ worker-build -h
Error: Everything required to configure and run the `wasm-pack build` command

Usage: --no-typescript [OPTIONS] [PATH] [EXTRA_OPTIONS]...

Arguments:
  [PATH]              The path to the Rust crate. If not set, searches up the path from the current directory
  [EXTRA_OPTIONS]...  List of extra options to pass to `cargo build`

Options:
  -s, --scope <SCOPE>        The npm scope to use in package.json, if any
  -m, --mode <MODE>          Sets steps to be run. [possible values: no-install, normal, force] [default: normal]
      --no-typescript        By default a *.d.ts file is generated for the generated JS file, but this flag will disable generating this TypeScript file
      --weak-refs            Enable usage of the JS weak references proposal
      --reference-types      Enable usage of WebAssembly reference types
  -t, --target <TARGET>      Sets the target environment. [possible values: bundler, nodejs, web, no-modules, deno] [default: bundler]
      --debug                Deprecated. Renamed to `--dev`
      --dev                  Create a development build. Enable debug info, and disable optimizations
      --release              Create a release build. Enable optimizations and disable debug info
      --profiling            Create a profiling build. Enable optimizations and debug info
  -d, --out-dir <OUT_DIR>    Sets the output directory with a relative path [default: pkg]
      --out-name <OUT_NAME>  Sets the output file names. Defaults to package name
      --no-pack              Option to not generate a package.json
      --no-opt               Option to skip optimization with wasm-opt
  -h, --help                 Print help
```

It looks as though `worker-build` propagated the `-h` flag to a `wasm-pack build` command, which then printed *its* help text.

This is confusing and not helpful. For example, it led me to believe that I could run `worker-build --out-dir foo`, which doesn't work because `worker-build` already supplied that argument (and it can't be specified twice).

### Steps To Reproduce

1. Run `worker-build -h` (in any directory; no project needed); produces the help text above.
2. After reading the help text, get confused and try `worker build --out-dir foo`, which returns an error.

jakubadamw:
Closes #739 by correcting the output of `--help` – I hid options that aren't meant to be configured by the user as well as corrected the program's name (from `--no-typescript` (!) to `worker-build`). The `--no-typescript` argument seemingly meant to be passed downstream but was actually ignored (it occupied the program name's position). Now that it's enabled, we no longer expect there to be an index_bg.wasm.d.ts file in the output directory, so I updated the cleanup accordingly.

This also cleans up the handling of mutually exclusive build profiles such as `--dev` and `--release`, seeing Clap can do this by default through an argument group with `multiple` set to `false`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,7 +82,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -93,7 +93,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


guybedford (approved):
These are great, thank you.

PR #946, opened 2026-03-10
Use `impl IntoIterator` consistently

logan-gatlin:
Use `impl IntoIterator` consistently in all APIs. This replaces two other patterns that were in use:
* `Vec<T>` parameters
* `impl Into<Option<Vec<T>>>`
Using a `Vec<T>` parameter when only an iterable collection is an anti-pattern. Using `IntoIterator` instead allows users to pass arrays by value instead since the function is now generic over all array sizes (among many other types). 

`impl Into<Option<Vec<T>>>` is also a poor practice. In every case where it was used, `Some(vec![])` and `None` where treated the same, making the `Option` wrapper redundant. Since `Option` implements iterator, users passing `None` will see no change.

This change is not backwards compatible, but most users will have to make little to no change to existing code.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -18,6 +18,9 @@ jobs:
 
     steps:
       - uses: dtolnay/rust-toolchain@1.87.0
+      - name: Enable symlinks on Windows
+        if: runner.os == 'Windows'
+        run: git config --global core.symlinks true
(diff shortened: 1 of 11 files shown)


guybedford:
As a breaking change, I will leave this open to reconsider as part of a comprehensive redesign of the API rather than doing incremental breaks sporadically.

PR #947, opened 2026-03-10
Move re-initialization logic into Rust/wasm-bindgen

logan-gatlin:
This PR moves (most of) the responsibility for re-initializing Rust workers away from our custom JS shim, and into Rust/wasm-bindgen.

Latest wasm-bindgen handles re-initialization automatically when building for the panic=unwind target. This means that most of the shim we provide for WASM modules is unnecessary. This PR creates a new minimal shim for panic=unwind builds.

Durable objects require slightly more work since their state must be reset on an abort. Previously, durable objects were heap allocated Rust structs managed by JS objects which kept pointers into Rust's memory. This is problematic when the instance resets, as these pointers become dangling. 

The new design is to keep a global `HashMap` in Rust containing all live DO instances, keyed by a unique `u32`. In our JS shim, we maintain a list of live DO ids and the arguments used to construct them. When re-initialization happens, every entry in this list is re-initialized.

*Notes*:
This requires a new feature in `wasm-bindgen` allowing for a post-reinitialization hook, right now I am just pointing the submodule to my feature branch. Both Rust and `wasm-bindgen` needed to be bumped to their latest version so I could write panic=unwind tests. This makes the diff messy because many doc tests subsequently failed to build, and CI had to be modified to support this new setup. When the wasm-bindgen reinit hook is merged I will probably split this up into two PRs


worker-build/src/main.rs
```diff
@@ -230,10 +230,18 @@ fn add_export_wrappers(out_dir: &Path) -> Result<()> {
     let shim_path = output_path(out_dir, "shim.js");
     let mut output = fs::read_to_string(&shim_path)
         .with_context(|| format!("Failed to read {}", shim_path.display()))?;
-    for class_name in class_names {
-        output.push_str(&format!(
-            "export const {class_name} = new Proxy(exports.{class_name}, classProxyHooks);\n"
-        ));
+    for class_name in &class_names {
+        if panic_unwind {
```
jakubadamw: I think this would read nicer as:

```rust
let export = if panic_unwind {
    // In panic=unwind mode, wasm-bindgen handles reinit natively.
    // Re-export classes directly without Proxy wrapping.
    format!("exports.{class_name}")
} else {
    format!("new Proxy(exports.{class_name}, classProxyHooks)")
};
output.push_str(&format!("export const {class_name} = {export};\n"));
```

or even:

```
write!(&mut output, "export const {class_name} = {export};\n")?;
```

for the last line (will require `use std::fmt::Write;`).

🙂

worker-build/src/main.rs
```diff
@@ -230,10 +230,18 @@ fn add_export_wrappers(out_dir: &Path) -> Result<()> {
...
+            // In panic=unwind mode, wasm-bindgen handles reinit natively.
+            // Re-export classes directly without Proxy wrapping.
+            output.push_str(&format!(
+                "export const {class_name} = exports.{class_name};\n"
```
guybedford: I believe all exported classes still still need recovery given they are all treated as singletons in the worker with the same auto-reinit of the class singleton functionality expected.

guybedford (commented):
Would be nice to see how this integrates with the upstream branch of supporting StaleError and if that is sufficient to support the instance checks.

We also need to ensure unwind test coverage here.

logan-gatlin:
This PR broke doctests somehow, probably because I bumped the versions of a lot of things. Had to go in and add `ignore` and `norun` in places

worker-build/src/main.rs
logan-gatlin: fixed

logan-gatlin:
I updated this to use the new fork of reinit-hooks https://github.com/wasm-bindgen/wasm-bindgen/pull/5059

PR #959, opened 2026-04-01
deps: bump miniflare to 4.20260329.0 and remove undici dependency

connyay:
Use miniflare's built-in createFetchMock instead of undici's MockAgent. Update durable object tests to use state.id().name() directly now that workerd#2240 is fixed, removing the QueryParams workaround.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -64,7 +64,7 @@ jobs:
       - uses: cargo-bins/cargo-binstall@main
 
       - name: Install cargo-generate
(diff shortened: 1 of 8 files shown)


guybedford:
It seems this is breaking the containers tests, so I'm unable to land it unfortunately. We might need to rework how those behave further.

guybedford:
I've gone ahead and just disabled the containers tests for now, we can follow-up separately with figuring out the containers test coverage.

guybedford (dismissed):

guybedford (approved):

PR #962, opened 2026-04-01
Integrate re-init hook

logan-gatlin:
Simplifies the shim code for the panic=unwind path using the new wasm-bindgen reinit hook. I created a new shim file for panic unwind builds that handles incrementing the instanceId and setting the panic error hook in Rust.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,21 +22,21 @@
       }
     },
     "node_modules/@babel/runtime": {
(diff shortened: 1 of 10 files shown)


guybedford:
In theory this should now fix this test - https://github.com/cloudflare/workers-rs/blob/main/test/tests/panic.spec.ts#L183.

Can we try that out on this PR?

PR #974, opened 2026-04-12
Fix the `#[event(start)]` regression from v0.8

### LINKED ISSUE #973: [BUG] Regression in 0.8.0: Can’t use `#[event(start)]` without a direct `wasm-bindgen` dependency

0.8.0

n/a

### Describe the bug

After upgrading to v0.8, a start handler like this:

```rust
#[event(start)]
pub fn setup_hook() {
    // …
}
```

in a crate that does not have `wasm-bindgen` as a dependency, will fail to compile:

```
error[E0433]: cannot find module or crate `wasm_bindgen` in this scope
 --> lib.rs:LL:1
  |
LL | #[event(start)]
  | ^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `wasm_bindgen`
  |
  = note: this error originates in the attribute macro `event` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0433`.
error: could not compile `generate-newspaper-entry` (lib) due to 1 previous error
```

This is a consequence of [this change](https://github.com/cloudflare/workers-rs/commit/e363ba6#diff-5fbccf7aad8fecebdb79b59bcdf993ad2236c34b9360b7f6cc2862374497a610) where the import of `::worker::wasm_bindgen` was dropped.

### Steps To Reproduce

Cargo.toml
```toml
[package]
name = "issue-973"
version = "0.1.0"
edition = "2024"

[lib]
crate-type = ["cdylib"]

[dependencies]
# Only `worker` — no direct `wasm-bindgen` dependency.
# The `#[event(start)]` macro should re-export what it needs
# via `::worker::wasm_bindgen`, but since 0.8.0 it doesn't.
worker = "0.8.0"
```

src/lib.rs
```rust
use worker::event;

// Fails to compile with worker 0.8.0 because the `event` macro
// emits a bare `wasm_bindgen` reference instead of `::worker::wasm_bindgen`.
#[event(start)]
pub fn setup_hook() {}
```

jakubadamw:
Fixes a regression from https://github.com/cloudflare/workers-rs/commit/e363ba6#diff-5fbccf7aad8fecebdb79b59bcdf993ad2236c34b9360b7f6cc2862374497a610L244-R234 that made any use of `#[event(start)]` fail to compile unless the user crate had a direct dependency on `wasm-bindgen`.

Closes #973.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -113,7 +113,7 @@ jobs:
 
       - name: Build all examples
         run: |
-          sed -i 's/, "examples\/axum"//g' Cargo.toml
+          sed -i '/"examples\/axum",/d' Cargo.toml
           for example in examples/*/; do
               example_name=$(basename "$example")
               if [ "$example_name" = "coredump" ]; then
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,6 +201,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "axum-on-workers"
+version = "0.1.0"
+dependencies = [
+ "axum",
+ "axum-macros",
+ "serde",
+ "tower-service",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "worker",
+ "worker-macros",
+]
+
+[[package]]
 name = "base64"
 version = "0.22.1"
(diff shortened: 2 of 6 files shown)


guybedford:
Thanks so much!

PR #978, opened 2026-04-19
feat(durable object): expose `SyncKvStorage`

chise0713:
- add wasm bindings for `SyncKvStorage`, [reference](https://github.com/cloudflare/workerd/blob/v1.20260418.1/src/workerd/api/sync-kv.h)
- expose `kv()` on `DurableObjectStorage`
- add missing `startAfter` to `ListOptions` (present in workerd, both [async kv](https://github.com/cloudflare/workerd/blob/v1.20260418.1/src/workerd/api/actor-state.h#L69) and [sync kv](https://github.com/cloudflare/workerd/blob/v1.20260418.1/src/workerd/api/sync-kv.h#L22)) 

Related: #967


aDifferentJT:
Hi, I'm interested in using this API, it looks like status stalled on this, is there anything that can be done to help this along?

PR #979, opened 2026-04-20
feat(worker): add Flagship binding

connyay:
Adds first-class support for the Flagship feature-flag binding announced on 2026-04-17. Users can now evaluate flags via `env.flagship("FLAGS")` with a fluent `EvaluationContext` builder and typed `EvaluationDetails<T>` responses.

- worker-sys: raw wasm_bindgen externs for all 9 Flagship JS methods
- worker: Flagship wrapper, EvaluationContext builder, EvaluationDetails<T>
- test: mini-flagship miniflare mock + 10 handlers + 16 vitest specs
- examples/flagship: runnable worker demoing value/details/context usage


connyay:
@guybedford I took an initial swing at ts-gen for flagship. 

Mostly worked, but a few maybe controversial things:

<details>
<summary>(resolved) - promise values that returned primitives had pretty bad ergonomics:</summary>
1. promise values that returned primitives had pretty bad ergonomics:
```rust
let value = env
        .flagship(BINDING)?
        .get_string_value(&flag, "fallback")
        .await?
        .as_string()
        .unwrap_or("fallback")
```
~Tried this https://github.com/wasm-bindgen/ts-gen/pull/11 which improved consumer ergonomics at the cost of generation complexity and ugliness. let me know what you think.~

EDIT: less bad version https://github.com/wasm-bindgen/ts-gen/pull/14

not too bad:

```rust
let value: bool = env
        .flagship(BINDING)?
        .get_boolean_value(&flag, false)
        .await?
        .value_of();
```

```rust
let value = String::from(
        env.flagship(BINDING)?
            .get_string_value(&flag, "fallback")
            .await?,
);
```
</details>

2. I continue to fight with object generic / jsgeneric ergonomics in wasm
with ootb ts-gen the .get_object_value call looked like:

```rust
pub async fn handle_object(req: Request, env: Env, _data: SomeSharedData) -> Result<Response> {
    let flag = last_segment(&req)?;
    let default = serde_wasm_bindgen::to_value(&default_theme())?;
    let raw = env
        .flagship(BINDING)?
        .get_object_value(&flag, &default)
        .await?;
    let value: Theme = serde_wasm_bindgen::from_value(raw)?;
    Response::from_json(&serde_json::json!({ "flag": flag, "value": value }))
}
```

yuck. fix for this was to comment out the type for generation and handroll. open to ideas to improve this. hit similar stuff in workflows as we've discussed before https://github.com/cloudflare/workers-rs/pull/918#discussion_r3025038518

3. the ts-gen `_with_record_N` generation is pretty nasty. 
what do you think about https://github.com/wasm-bindgen/ts-gen/pull/13

guybedford:
Thanks for working through the concerns here! This gives a lot of interesting feedback to think about - I will look into this next week and see if I can come up with some further suggestions as well.

examples/flagship/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    let url = req.url()?;
+    let flag = query(&url, "flag").unwrap_or_else(|| "checkout-flow".into());
+    let flagship = env.flagship(BINDING)?;
+    let value = String::from(match query(&url, "userId") {
+        Some(user_id) => {
+            let ctx = EvaluationContext::new()
+                .string("userId", &user_id)
+                .string("country", "US");
+            flagship
+                .get_string_value_with_record(&flag, "control", ctx.as_ref())
```
connyay: unshippable, imo

guybedford: What should it be, or just the naming? Is `get_string_value_with_context` better? Note you can just use `&ctx`.

connyay: `get_string_value_with_context` is better, or `get_string_value` with an Option for context. the `_record_1` and `_record_2` are effectively useless/uncallable currently.

guybedford:
I ran out of time again this week to include this one - but there are a number of major updates to ts-gen that might update this output considerably.

It would be good to continue working through the outstanding bindgen edges here, perhaps we can sync on this via Discord or otherwise further.

guybedford:
I updated this to the latest ts-gen which now outputs correct generics. @connyay please take a look.

I agree with the use of wrappers to support custom structs for details and object types - at least until we can support first-class structs in wasm-bindgen itself. This is great feedback to motivate that work.

Overall, the current approach looks good to land to me, and we now no longer name methods like `record2` etc in ts-gen rather using descriptive names based on param positions.

Please let me know if you spot any more rough edges!

connyay:
This looks great to me! Ship it

guybedford:
Putting some more thought to this the `JsString` etc generics just don't work nicely. Instead I posted https://github.com/wasm-bindgen/wasm-bindgen/pull/5167 and have based the bindings to that which support proper primitive generic types in the functions and FlagshipEvaluationDetails.

After the next release of wasm-bindgen will re-evaluate then. There are likely further improvements that can be made as well.

guybedford:
Unfortunately https://github.com/wasm-bindgen/wasm-bindgen/pull/5167 isn't going to land, instead I'm hopeful https://github.com/wasm-bindgen/wasm-bindgen/pull/5180 can give us proper generics instead, but that will still take a while to review.

Therefore this PR would need to instead patch the remaining generic usability cases such as for `JsString` returns to not use automated bindgen.

guybedford:
@connyay would you be able to update this PR to use a manual bindings for `String` and `Number` again? I think that would be better to land in the mean time while we wait for proper ABI-level generics.

connyay:
@guybedford should be good to go. ptal

ajharhussain250-jpg:
You are workr but not with me coopretive?why?

connyay:
👉  https://github.com/cloudflare/workers-rs/pull/1053

guybedford:
@connyay thanks again for your patience on this one! Any feedback on the ergonomics of proper generics in that other PR very much welcome too.

PR #999, opened 2026-05-14
wasm-bindgen@0.2.122 and unwind safety updates

guybedford:
This updates to wasm-bindgen@0.2.122.

In the process, it adopts the unwind safety enforcement landed in wasm-bindgen/wasm-bindgen#5128, which adds `MaybeUnwindSafe` bounds on `#[wasm_bindgen]` exports and closure types so that types crossing the JS/Rust boundary under `panic=unwind` are explicitly unwind safe.

Picking up that change required updates in two dependencies, which propagate the bound through their public APIs:

* `wasm-streams` (https://github.com/MattiasBuelens/wasm-streams/pull/35) - implements `UnwindSafe`/`RefUnwindSafe` for the streams wrappers
* `gloo-timers` (https://github.com/rustwasm/gloo/pull/562) - propagates unwind safety to user callbacks

Both are added here as submodules pointing at the corresponding fork branches and patched via `[patch.crates-io]` until the upstreams land and are published.

The durable object test fixtures (`test/src/counter.rs`, `test/src/durable.rs`) had `RefCell<T>` fields holding `Copy` primitives. These were refactored to `Cell<T>` wrapped in `AssertUnwindSafe`:

```rs
// before
count: RefCell<usize>,
*self.count.borrow_mut() += 10;
```

```rs
// after
count: AssertUnwindSafe<Cell<usize>>,
self.count.set(self.count.get() + 10);
```

`Cell<T: Copy>` is genuinely unwind safe in single-threaded wasm - every operation is an atomic move with no borrow guard to leave dangling - but the type system rejects it because `UnsafeCell` has a blanket negative `RefUnwindSafe` impl. `AssertUnwindSafe` is the narrow assertion scoped to just the interior-mutability fields rather than a blanket `impl RefUnwindSafe` on the durable object structs (which would also be asserting unwind safety on the `State` and `Env` fields we don't own).

Also adds a `.cargo/config.toml` URL replacement so transitive git deps fetched via SSH fall back to HTTPS, useful for `cargo generate` users without GitHub SSH set up.

Tested via `npm run test-panic-unwind` - the panic-unwind build now compiles cleanly and the existing test suite passes (one pre-existing rate_limit test failure unrelated to this change).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,5 @@
 [target.wasm32-unknown-unknown]
 rustflags = ["--cfg", "getrandom_backend=\"wasm_js\""]
+
+[url."https://github.com/"]
+insteadOf = ["git@github.com:", "ssh://git@github.com/"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,2 @@
 [target.wasm32-unknown-unknown]
 rustflags = ["--cfg", "getrandom_backend=\"wasm_js\""]
-
-[url."https://github.com/"]
-insteadOf = ["git@github.com:", "ssh://git@github.com/"]
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,14 @@
 [submodule "wasm-bindgen"]
 	path = wasm-bindgen
 	url = https://github.com/wasm-bindgen/wasm-bindgen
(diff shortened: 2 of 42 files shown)


guybedford:
This release is currently pending https://github.com/MattiasBuelens/wasm-streams/pull/35.

Issue #774, opened 2025-07-14
[Feature] Add `[must_use]` attribute and implement `IntoFuture` for `D1PreparedStatement`


lexoliu:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

`D1PreparedStatement` is just a statement, which has not been executed actually. However, user could make mistake by writing code like this:
```rust
query!(&self.db, "DELETE FROM users WHERE id = ?", id)?; // correct version: query!(&self.db, "DELETE FROM users WHERE id = ?", id)?.all().await?
```
This code does not work, `D1PreparedStatement` just like a `Future` in Rust. It wouldn't  been executed automatically, it must been called explicitly .

I suggest that we could implement `IntoFuture` for `D1PreparedStatement`, so that we can execute a statement by `await` it directly. What's more, `D1PreparedStatement` need to been attributed as `#[must_use]`, to avoid misuse statically.

lexoliu:
I'm confused about the functionality of `D1PreparedStatement::run`, I looked up the [documents](https://developers.cloudflare.com/d1/worker-api/prepared-statements/) of Cloudflare. It says the `run` method

 > Runs the prepared query (or queries) and returns results. The returned results includes metadata.

While [`D1PreparedStatement::run`](https://docs.rs/worker/latest/worker/d1/struct.D1PreparedStatement.html#method.run)

>  Executes a query against the database but only return metadata.

Why the behaviour between Rust and Javascript is different?

<img width="765" height="446" alt="Image" src="https://github.com/user-attachments/assets/675fd0b2-bc1c-4281-9924-b93c421d6cd8" />

PR #1007, opened 2026-06-12
feat(d1): add IntoFuture and #[must_use] to D1PreparedStatement

eyupcanakman:
`D1PreparedStatement` holds a query that has not run yet. With no terminal call it compiles fine and does nothing, e.g. `query!(&db, "DELETE FROM users WHERE id = ?", id)?;`. Adding `#[must_use]` to the type makes that a compile warning, matching the KV builders (`GetOptionsBuilder`, `PutOptionsBuilder`).

#774 also asks for `impl IntoFuture` so a statement runs when you `.await` it. That would be the first `IntoFuture` in the crate, and `await stmt` has to pick one terminal method. I can follow up with one delegating to `run()` to match the JS `stmt.run()`, or `all()` if you'd prefer.

Part of #774.

### DIFF
--- a/test/src/d1.rs
+++ b/test/src/d1.rs
@@ -297,7 +297,8 @@ pub async fn insert_and_retrieve_optional_none(
         &None::<String>,
         &None::<u32>
     )?;
-    query.run().await?;
+    // Run the insert by awaiting the statement directly (exercises IntoFuture).
+    query.await?;
 
     let stmt = worker::query!(&db, "SELECT * FROM nullable_people WHERE id = 3");
     let person = stmt.first::<NullablePerson>(None).await?.unwrap();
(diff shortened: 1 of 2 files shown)


guybedford:
For the `must_use` I'm open to either way on this one. Formally it's not a correctness issue if it isn't used though.

For the `IntoFuture` on the query itself that would be great though definitely.

eyupcanakman:
Added the `IntoFuture` impl so awaiting a statement runs it through `run()`. Kept the `must_use` alongside it, but can drop it if you'd rather this PR stay scoped to just `IntoFuture`.

guybedford (approved):

ajharhussain250-jpg (commented):
Hi i am a workr and i hop you cansider me on x twitter.  best wishise for you.  your  freind Azher Hussain Khuwaja.

PR #1015, opened 2026-06-16
fix: re-add reference-types target feature stripped by `strip = true`

### LINKED ISSUE #1014: [BUG] Can no longer compile with `strip = true`

0.8.5

4.101.0

### Describe the bug

`wasm-bindgen 0.2.125` hard requires externref table in wasm binaries and this table is removed with `strip = true` in cargo `[profile.release]`

It fails to compile with the following error:
```
error: failed to generate catch wrappers

Caused by:
    externref table required for catch wrappers
Error: Running the wasm-bindgen CLI
```

### Steps To Reproduce

_No response_

connyay:
`strip = true` (alias for `strip = "symbols"`) drops the `target_features` custom section from the compiled wasm, which removes the `reference-types` advertisement that `wasm-bindgen` needs to run its externref transform. Without it, builds fail with "externref table required for catch wrappers".

Generalize the wasm custom-section helpers in `producers.rs` so they can be reused, and add a new `reference_types` module that patches the feature back in after cargo and before `wasm-bindgen` runs.

Fixes #1014

### DIFF
--- a/worker-build/src/build/mod.rs
+++ b/worker-build/src/build/mod.rs
@@ -510,6 +510,11 @@ pub fn wasm_bindgen_build(
         .join(data.crate_name())
         .with_extension("wasm");
 
+    // `strip = true` drops the target_features section that wasm-bindgen
+    // needs to enable its externref transform. Patch it back in.
+    crate::reference_types::ensure_reference_types_feature(&wasm_path)
+        .context("Failed to ensure the reference-types target feature is advertised")?;
+
     let dts_arg = if disable_dts {
         "--no-typescript"
     } else {
--- a/worker-build/src/main.rs
+++ b/worker-build/src/main.rs
@@ -19,6 +19,7 @@ mod emoji;
 mod lockfile;
 mod main_legacy;
 mod producers;
+mod reference_types;
 mod versions;
 
 use build::{Build, BuildOptions};
--- a/worker-build/src/producers.rs
+++ b/worker-build/src/producers.rs
@@ -9,11 +9,14 @@ struct ProducersEntry<'a> {
     version: &'a str,
 }
 
-struct ProducersSection {
-    start: usize,
-    end: usize,
-    content_start: usize,
-    content_end: usize,
+/// Byte range of a wasm custom section (id 0).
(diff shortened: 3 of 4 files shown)


worker-build/src/build/mod.rs
```diff
@@ -510,6 +510,11 @@ pub fn wasm_bindgen_build(
         .join(data.crate_name())
         .with_extension("wasm");
 
+    // `strip = true` drops the target_features section that wasm-bindgen
+    // needs to enable its externref transform. Patch it back in.
+    crate::reference_types::ensure_reference_types_feature(&wasm_path)
+        .context("Failed to ensure the reference-types target feature is advertised")?;
```
connyay, line 516: I'm torn on if patching is the right call here. https://rustwasm.github.io/docs/book/reference/code-size.html#optimizing-builds-for-code-size doesn't mention `strip = true`, so I am not sure how widespread `strip = true` or `strip = "symbols"` is in the wild.

Might be safer to detect `strip = true` or `strip = "symbols"` and raise an error instead. Open to suggestions/feedback here.

guybedford:
Could we just ungate the reference types feature check for externref transforms in wasm bindgen? Surely it can just be assumed now?

connyay:
> Could we just ungate the reference types feature check for externref transforms in wasm bindgen? Surely it can just be assumed now?

yes. I'll take that approach

connyay:
> > Could we just ungate the reference types feature check for externref transforms in wasm bindgen? Surely it can just be assumed now?
> 
> yes. I'll take that approach

hm. actually. would need to bump the wasm-bindgen rust version to 1.82 - is there appetite for that?
https://github.com/wasm-bindgen/wasm-bindgen/blob/6ff7f1cd2f636d6f19d29b45680885088cc9e653/Cargo.toml#L15

guybedford:
We can't bump the library MSRV for another 4 months. The CLI MSRV (the one required by the actual processor), should already be above 1.82 though?

connyay:
👉  https://github.com/wasm-bindgen/wasm-bindgen/pull/5205

PR #660, opened 2024-10-22
Refactor according to Vectorize V2 spec

garvit-gupta:
Update the Vectorize interface according to the V2 schema. 

This PR is built on top of @kflansburg's PR: #609, but includes the changes needed for the V2 spec.

For details about Vectorize, refer to https://developers.cloudflare.com/vectorize/.

API Reference: https://developers.cloudflare.com/api/operations/vectorize-list-vectorize-indexes

PR #1019, opened 2026-06-23
feat(vectorize): add Vectorize (V2) binding

connyay:
Adds the Cloudflare Vectorize binding. The bulk is generated from a curated `types/vectorize.d.ts` via ts-gen (same approach as the `email` binding), with a thin hand-written layer for the parts ts-gen can't infer: the `EnvBinding` impl, `env.vectorize(binding)`, and some `f32`/serde conveniences.

The curated types are a V2-only subset of upstream. Filters and metadata serialize as plain objects rather than JS `Map`s, which Vectorize silently ignores.

Tests are a `/vectorize` route that hits the binding end to end. That's compile coverage on wasm32 only; there's no vitest spec because Vectorize has no local Miniflare emulator. See examples/vectorize for a semantic-search example using Workers AI embeddings.

Regenerating the bindings needs the ts-gen string-enum codegen fixes (pending upstream, plus a submodule bump). The committed output already includes them, so the crate compiles against the pinned wasm-bindgen as-is.


connyay:
Supersedes https://github.com/cloudflare/workers-rs/pull/660 and https://github.com/cloudflare/workers-rs/pull/609 with ts-gen plumbing

Needs https://github.com/wasm-bindgen/ts-gen/pull/30 to land properly

PR #1022, opened 2026-07-07
perf: remove redundant copies and JS boundary crossings on hot paths

connyay:
- http/body.rs: Bytes::from(vec) instead of copy_from_slice, removing a second copy of every body chunk in the http-feature body bridge
- d1: bind D1Type::Blob as an ArrayBuffer via Uint8Array (one memcpy) instead of serializing element-by-element into a JS number array
- sql.rs: use js_sys::IteratorNext accessors instead of per-row Reflect::get with fresh key strings; drop the per-row Array.from copy in the raw cursor path
- response.rs: Response::json() parses the buffered body in place instead of cloning it into a String via text()
- response.rs: Response::cloned() duplicates fixed/empty bodies on the Rust side, reserving the JS Response.clone() tee for stream bodies

Adds a D1 blob round-trip integration test (previously uncovered).

### DIFF
--- a/worker/src/d1/mod.rs
+++ b/worker/src/d1/mod.rs
@@ -286,7 +286,10 @@ impl<'a> From<&'a D1Type<'a>> for JsValue {
             D1Type::Integer(i) => JsValue::from_f64(i as f64),
             D1Type::Text(s) => JsValue::from_str(s),
             D1Type::Boolean(b) => JsValue::from_bool(b),
-            D1Type::Blob(a) => serde_wasm_bindgen::to_value(a).unwrap(),
+            // D1's documented BLOB representation is an `ArrayBuffer`; copying
+            // through a `Uint8Array` crosses the JS boundary once instead of
+            // serializing element-by-element into a number array.
+            D1Type::Blob(a) => js_sys::Uint8Array::from(a).buffer().into(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


worker/src/d1/mod.rs
```diff
@@ -286,7 +286,10 @@ impl<'a> From<&'a D1Type<'a>> for JsValue {
             D1Type::Integer(i) => JsValue::from_f64(i as f64),
             D1Type::Text(s) => JsValue::from_str(s),
             D1Type::Boolean(b) => JsValue::from_bool(b),
-            D1Type::Blob(a) => serde_wasm_bindgen::to_value(a).unwrap(),
+            // D1's documented BLOB representation is an `ArrayBuffer`; copying
+            // through a `Uint8Array` crosses the JS boundary once instead of
+            // serializing element-by-element into a number array.
+            D1Type::Blob(a) => js_sys::Uint8Array::from(a).buffer().into(),
```
connyay, line 292: hmph. this is a good change, but might break some consumers. Worth a break here IMO

worker/src/http/body.rs
```diff
@@ -74,7 +74,7 @@ impl HttpBody for Body {
             stream
                 .poll_next_unpin(cx)
                 .map_ok(|buf| {
-                    let bytes = Bytes::copy_from_slice(&js_sys::Uint8Array::from(buf).to_vec());
+                    let bytes = Bytes::from(js_sys::Uint8Array::from(buf).to_vec());
```
connyay, line 77: #1 win in the PR. Maybe worth breaking to its own PR if the rest of it is spooky/controversial

guybedford (approved):
These are really great, thank you!

PR #1025, opened 2026-07-10
feat(wrangler-switch): switchinfg from wrangler.toml to wrangler.jsonc

AOrps:
# switching from wrangler.toml -> wrangler.jsonc on hello-world template in workers-rs

## Testing
```sh
# generating the template use local repo
pwd && mkdir -p generated &&  cargo generate --path templates/hello-world --name test-hello --destination ./generated --force --define panic_unwind=false

# change dir to test worker
cd templates/hello-world 

# testing view on dev
npx wrangler dev 

# ensuring that wrangler takes
npx wrangler deploy --dry-run
```

Let me know if there is anything else, I should add or it I should add it to all of the rust templates?

### DIFF
--- a/templates/hello-world/wrangler.jsonc
+++ b/templates/hello-world/wrangler.jsonc
@@ -1,0 +1,9 @@
+{
+    "$schema": "node_modules/wrangler/config-schema.json",
+    "name": "{{project-name}}",
+    "main": "build/index.js",
+    "compatibility_date": "{{ "now" | date: "%Y-%m-%d" }}",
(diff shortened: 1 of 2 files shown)


guybedford:
Thanks for the PR. While jsonc is recommended in the JS ecosystem, I think it would be fine to retain the toml configuration format in the Rust ecosystem, given its extensive use, at least for now. If we reach a point where wrangler features start to heavily favor jsonc over TOML configuration we can certainly reconsider in future.

PR #1033, opened 2026-07-17
Re-export url::ParseError alongside url::Url

### LINKED ISSUE #380: [BUG] ParseError is not exported from url crate when implementing From<ParseError> for CustomError

0.0.18

3.5.1

### Describe the bug

When using `Fetch::url`, we first need to have the Url ready before we can use it.

```rust
async fn fetch_price(ticker_1: &str, ticker_2: &str) -> Result<Response, CustomError> {
    let path = format!(
        "{ENDPOINT}?symbol={}{}",
        ticker_1.to_uppercase(),
        ticker_2.to_uppercase()
    );

    let url_ = Url::parse(&path);
    // rest code
}
```

By calling `Url::parse` it returns a `Result<Url, ParseError>`, so we need to do a `match`.

```rust
    match url_ {
        Ok(url) => {
            let response = Fetch::Url(url).send().await?;
            return Ok(response);
        }
        Err(_) => Err(CustomError::new("Failed to parse.".into(), 404u16)),
    }
```

The above could be simplified if `url::ParseError` is exported along with `url::Url`.

```rust
    let url_ = Url::parse(&path)?;
```

But the above will not compile, it says `the trait From<url::parser::ParseError> is not implemented for CustomError`, now this should end because even if I implemented `From<ParseError> for CustomError`, `ParseError` is not exported.

We can fix this by exporting ParseError in:
https://github.com/cloudflare/workers-rs/blob/main/worker/src/lib.rs line 11
```rust
pub use url::{ParseError, Url};
```

Now this works!
```rust
#[derive(Debug)]
struct CustomError {
    status: u16,
    message: String,
}

impl From<ParseError> for CustomError {
    fn from(value: ParseError) -> Self {
        Self {
            status: 404u16,
            message: value.to_string(),
        }
    }
}

async fn fetch_price(ticker_1: &str, ticker_2: &str) -> Result<Response, CustomError> {
    let path = format!(
        "{ENDPOINT}?symbol={}{}",
        ticker_1.to_uppercase(),
        ticker_2.to_uppercase()
    );

    let url_ = Url::parse(&path)?;
    let response = Fetch::Url(url_).send().await?;
    return Ok(response);
}

```

### Solution:
Since `url::Url` is re-exported, `url::ParseError` should be exported too.

### Steps To Reproduce

_No response_

eyupcanakman:
`Url::parse` returns `Result<Url, ParseError>` and `Error` already has a `From<url::ParseError>` impl, but only `Url` is re-exported. Naming the error type means taking a direct dependency on `url`.

Re-exports it alongside `Url` (`pub use url::{ParseError, Url};`), with a trybuild `pass` test next to the `start-handler.rs` one from #974.

Adding `ParseError` at the root can conflict under `use worker::*` if `chrono`'s `ParseError` is also in scope via `*`, though I didn't find a public crate doing that and an explicit import clears it. Given #890, I can re-export the `url` module for `worker::url::ParseError` instead if you'd prefer.

Closes #380.

### DIFF
--- a/worker/src/lib.rs
+++ b/worker/src/lib.rs
@@ -149,7 +149,8 @@ use std::result::Result as StdResult;
 #[doc(hidden)]
 pub use async_trait;
 pub use js_sys;
-pub use url::{ParseError, Url};
+pub use url;
+pub use url::Url;
 pub use wasm_bindgen;
 pub use wasm_bindgen_futures;
 pub use web_sys;
--- a/worker/src/lib.rs
+++ b/worker/src/lib.rs
@@ -149,7 +149,7 @@ use std::result::Result as StdResult;
 #[doc(hidden)]
 pub use async_trait;
 pub use js_sys;
-pub use url::Url;
+pub use url::{ParseError, Url};
 pub use wasm_bindgen;
 pub use wasm_bindgen_futures;
 pub use web_sys;
--- a/worker/tests/pass/url-parse-error.rs
+++ b/worker/tests/pass/url-parse-error.rs
@@ -2,7 +2,7 @@
 // `Url::parse` returns `Result<Url, ParseError>`, and `Error` has a
 // `From<url::ParseError>` impl, so the error type must be nameable through
 // `::worker::` without `url` as a direct dependency.
-use worker::{ParseError, Url};
+use worker::{url::ParseError, Url};
 
 fn parse(input: &str) -> Result<Url, ParseError> {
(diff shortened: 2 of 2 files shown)


guybedford:
Thanks for the PR! I changed this to reexport `url` directly rather, so it becomes `url::ParseError` instead - clarifying the namespacing and better supporting further similar issues.

guybedford (approved):

Issue #177, opened 2022-05-09
async initializer for durable objects


cometkim:
I'm not sure it is already supported in workers-rs. But hopefully there would be an equivalent binding to `blockConcurrencyWhile` in JavaScript. Something like `async fn new(...) -> Self`

I use a pattern that refers to configs stored in KV when initializing DOs through it, but I don't know how to do the same in a workers-rs project.

andrewdavidmackenzie:
Same need here.

I think an `async new()` implementation could handle a failure to read from storage and in that case return a new instance created from default values - and avoid the need for `new()` to return `Result`

PR #1039, opened 2026-07-31
Add blockConcurrencyWhile binding for Durable Objects

lukevalenta:
NOTE: I used AI to help draft this PR, but I understand and approve the code and think this would be a useful feature for the Rust bindings.

Bind the runtime's blockConcurrencyWhile via State::block_concurrency_while (errors reset the object) and block_concurrency_while_infallible (errors returned as values). This mostly addresses #177, but since a literal async constructor isn't supported we instead document a lazy-init handler pattern.

### DIFF
--- a/.changeset/do-block-concurrency-while.md
+++ b/.changeset/do-block-concurrency-while.md
@@ -2,7 +2,10 @@
 "workers-rs": minor
 ---
 
-Add `State::block_concurrency_while` and `State::block_concurrency_while_infallible` for
-Durable Objects, binding the runtime's `blockConcurrencyWhile`. Both run an async closure
-while blocking delivery of other events until it completes; the `_infallible` variant
-returns application errors as values instead of resetting the object.
+Add `State::block_concurrency_while` for Durable Objects, binding the runtime's
+`blockConcurrencyWhile` with the same call-time semantics as JavaScript: the event delivery
+gate closes eagerly at the call, and the returned future yields the closure's value when
+awaited. Discarding the future instead supports the constructor idiom, where `new()` gates
+delivery of all events until async initialization completes. A closure returning `Err`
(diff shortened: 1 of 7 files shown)


guybedford:
Thanks for posting this one. To try to better align with the JS semantics I just posted up a change. AI summary below - 

The original wrapper deferred the blockConcurrencyWhile call to first poll of the returned future, which silently diverged from JS semantics: in JS the gate closes as a side effect of the call, and awaiting the promise is optional. That call-time/await-time fusion is what made the constructor idiom impossible and forced the lazy-init pattern (which needs an initialized check replicated across every handler — fetch, alarm, all websocket callbacks — and breaks if any await precedes the check).

The reworked API is a single eager method:

```rust
pub fn block_concurrency_while<F, Fut, T>(&self, closure: F) -> impl Future<Output = Result<T>>
```

The JS call happens in the fn body (closure via once_into_js, so its lifetime is decoupled from the future). Both JS idioms then fall out of one signature: await it in a handler for an atomic critical section returning T, or call it un-awaited from new() to gate all event delivery during async init — exactly matching un-awaited blockConcurrencyWhile in a JS constructor. This also let me drop _infallible: errors-as-values is just returning them inside Ok (T = Result<_, E>), now covered by a doc note and a test.

Tests carried over with one strengthened: the lazy-init test became a constructor-gating test — new() fires the init (two awaits, 0-sentinel field), and the spec asserts the very first request already observes the loaded value with the closure having run exactly once. The RMW and reset-on-Err tests are unchanged and still pass against real workerd via miniflare.

guybedford (dismissed):

guybedford:
@lukevalenta would value your confirmation the changes I added work for your use case before landing further.

guybedford (dismissed):

guybedford (approved):

PR #1041, opened 2026-08-03
feat: add bindings for inbound connect() TCP handler

ThomasRubini:
Dependent on https://github.com/cloudflare/workers-sdk/pull/14995 because I'd love to add a wrangler config with the TCP handler in the example

### DIFF
--- a/worker/src/lib.rs
+++ b/worker/src/lib.rs
@@ -195,7 +195,7 @@ pub use crate::response::{EncodeBody, IntoResponse, Response, ResponseBody, Resp
 pub use crate::router::{RouteContext, RouteParams, Router};
 pub use crate::schedule::*;
 pub use crate::secret_store::SecretStore;
-pub use crate::socket::*;
+pub use crate::socket::{ConnectionBuilder, FromSocket, SecureTransport, Socket, SocketAddress, SocketInfo, SocketOptions};
 pub use crate::streams::*;
 pub use crate::version::*;
 pub use crate::websocket::*;
--- a/worker/src/lib.rs
+++ b/worker/src/lib.rs
@@ -195,7 +195,7 @@ pub use crate::response::{EncodeBody, IntoResponse, Response, ResponseBody, Resp
 pub use crate::router::{RouteContext, RouteParams, Router};
 pub use crate::schedule::*;
 pub use crate::secret_store::SecretStore;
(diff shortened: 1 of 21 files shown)


worker/src/lib.rs
```diff
@@ -194,7 +194,7 @@ pub use crate::response::{EncodeBody, IntoResponse, Response, ResponseBody, Resp
 pub use crate::router::{RouteContext, RouteParams, Router};
 pub use crate::schedule::*;
 pub use crate::secret_store::SecretStore;
-pub use crate::socket::*;
+pub use crate::socket::{ConnectionBuilder, FromSocket, SecureTransport, Socket, SocketAddress, SocketInfo, SocketOptions};
```
guybedford: It seems this is missing `postgres_tls`, either keep the glob or include that explicitly.

guybedford (commented):
It would be great to include this when ready. Please just make sure to include a full test once the SDK change has landed.

worker/src/lib.rs
ThomasRubini: Hum, thank you I'm not sure how this even happened

ThomasRubini:
@guybedford AFAIK tests use miniflare and not wrangler, and miniflare is currently experimenting with a v5 alpha release, which means the only way to test connect() is to switch all tests to use an unstable (?) API
Would you accept a follow-up PR for a test once miniflare v5 releases (I'm not sure when that'll be) ?

guybedford (approved):
Thanks for adding the full Miniflare test. Verified locally: the new inbound connection test passes, and the example works end-to-end under \`wrangler dev\` with the \`[[connect]]\` config (\`nc 127.0.0.1 5432\` echoes back).

I pushed a small commit fixing the README link (rustdoc intra-doc path in a plain markdown file) and switching the handler error path to \`console_error!\` to match the fetch handler.

Two notes, neither blocking:

* The example's build command installs \`worker-build@^0.8\` from crates.io, which won't include the \`connect\` shim until the next \`worker-build\` release, so the example will only work once that ships.
* The example comment says the socket is closed automatically when the handler returns. The test destroys the client side itself so this isn't exercised. Have you confirmed workerd closes the server side on promise resolution? If so it would be worth documenting that on \`Socket\`.

guybedford (approved):

PR #1048, opened 2026-08-20
fix: add body_json and content_type to queue messages

### LINKED ISSUE #1013: [BUG] Queue: typed `MessageBatch<T>` deserialize via `serde_wasm_bindgen::from_value` mangles payloads — offer a JSON-based path

0.8.4

4.100.0

### Summary

The typed queue-consumer path (`MessageBatch<T>` / `Message::try_from(RawMessage)`) deserializes each body with `serde_wasm_bindgen::from_value`. For real-world payloads we found its V8↔serde type coercion **unreliable** — it produces wrong data or an opaque deserialize error. Bypassing it with `JSON.stringify(body)` + `serde_json::from_str` works reliably for the exact same payloads.

### Environment
- `worker` **0.8.4**, `wasm32-unknown-unknown`, Rust 1.96.
- Payload: a struct with `serde_json::Value`s and tuples, e.g. `substitutes: Vec<(String, serde_json::Value)>`, plus `Vec<String>`, strings.

### Reliable workaround (consumer side)
```rust
use js_sys::JSON;
fn parse<T: serde::de::DeserializeOwned>(body: &JsValue) -> Result<T, String> {
    let s = JSON::stringify(body).ok().and_then(|s| s.as_string())
        .ok_or("JSON.stringify produced no string")?;
    serde_json::from_str(&s).map_err(|e| e.to_string())
}
```
Round-trips through plain JSON and Just Works where `serde_wasm_bindgen::from_value` mangled the data. (This is also what the previous generation of our consumer hand-rolled — the footgun is long-standing.)

### Asks
- Document the limitation + this `JSON.stringify` + `serde_json` escape hatch in the Queues guide — it was the single biggest footgun we hit.
- Consider a `content_type`-aware / JSON-based deserialize helper on `MessageBatch` / `Message`, e.g. `msg.body_json::<T>()`, that uses `JSON.stringify` + `serde_json` for `json`-typed bodies instead of `serde_wasm_bindgen::from_value`.
- More generally: surface the message's `content_type` to the consumer so it can choose the right decoder.

### Steps To Reproduce

```rust
// worker-0.8.4/src/queue.rs
impl<T> TryFrom<RawMessage> for Message<T> where T: DeserializeOwned {
    fn try_from(value: RawMessage) -> Result<Self, Error> {
        let body = serde_wasm_bindgen::from_value(value.body())?; // <-- V8 coercion
        ...
    }
}
```

`serde_wasm_bindgen`'s coercion misbehaves on `serde_json::Value`, tuples (`Vec<(String, Value)>`), maps, and number / `undefined` edge cases.

Related to #1012

SameerVers3:
Closes #1013

serde_wasm_bindgen::from_value is unreliable for types like serde_json::Value and tuples. This adds an escape hatch:

```rust
for msg in batch.raw_iter() {
    let body: MyPayload = msg.body_json()?;
}
```

Also exposes content_type() on Message and RawMessage so consumers can inspect the producer's format.

### DIFF
--- a/worker/src/queue.rs
+++ b/worker/src/queue.rs
@@ -99,6 +99,16 @@ impl<T> Message<T> {
     pub fn raw_body(&self) -> JsValue {
         self.inner().body().unwrap()
     }
+
+    /// The content type of the message.
+    pub fn content_type(&self) -> Option<QueueContentType> {
+        self.inner.content_type().unwrap().map(|s| {
+            s.as_string()
+                .unwrap()
+                .parse()
+                .unwrap_or(QueueContentType::V8)
+        })
+    }
 }
 
 impl<T> TryFrom<RawMessage> for Message<T>
@@ -126,6 +136,24 @@ impl RawMessage {
     /// The body of the message.
     pub fn body(&self) -> JsValue {
         self.inner.body().unwrap()
+    }
+
+    /// The content type of the message.
+    pub fn content_type(&self) -> Option<QueueContentType> {
+        self.inner.content_type().unwrap().map(|s| {
+            s.as_string()
+                .unwrap()
+                .parse()
+                .unwrap_or(QueueContentType::V8)
+        })
+    }
+
+    /// deserialize the body as json using serde_json
+    pub fn body_json<T: DeserializeOwned>(&self) -> Result<T> {
+        let json_str = js_sys::JSON::stringify(&self.body())?
+            .as_string()
+            .ok_or_else(|| Error::RustError("failed to stringify body".into()))?;
+        Ok(serde_json::from_str(&json_str)?)
     }
 }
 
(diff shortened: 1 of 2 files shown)


SameerVers3:
@guybedford Can you review this?

worker/src/queue.rs
```diff
@@ -99,6 +99,19 @@ impl<T> Message<T> {
...
         self.inner().body().unwrap()
     }
+
+    /// The content type of the message.
+    pub fn content_type(&self) -> Option<QueueContentType> {
+        self.inner
+            .content_type()
+            .unwrap()
+            .map(|s| match s.as_string().unwrap().as_str() {
+                "json" => QueueContentType::Json,
```
jakubadamw: Could use `strum::EnumString` here much like it’s done [here](https://github.com/cloudflare/workers-rs/blob/main/worker-macros/src/event.rs#L5-L14). But I suppose this repetitive pattern is already spread across the module, so it’s a matter of taste.

SameerVers3: yes you are correct, I missed it. updating it now

PR #1053, opened 2026-09-01
Add flagship bindings with per-monomorphization generics

logan-gatlin:
Amends the existing flagship work to use per-monomorphization generics. Eliminates the `into_value()` conversions. Also adds `impl (De)Serialize` for the types.

Thanks to @connyay for doing most of the work on this


guybedford (changes requested):
This is an excellent proof of concept proving real generics solves the issue we were having!

Let's integrate `ts-gen` generation into this PR, by adding support for generic-mono in ts-gen under an option, which when set outputs similar output to what you had here, but automatically.

logan-gatlin:
Regenerated bindings using this branch of ts-gen: https://github.com/wasm-bindgen/ts-gen/pull/31

worker/src/flagship.rs
```diff
@@ -0,0 +1,241 @@
+use crate::{send::SendFuture, EnvBinding, Result};
+use js_sys::{Object, Promise};
+use serde::{de::DeserializeOwned, Deserialize, Deserializer, Serialize, Serializer};
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::JsFuture;
+
+// Hand-written companion to the ts-gen output in `bindings/flagship.rs`.
+// Object values still need serde conversion because arbitrary Rust structs are
+// not wasm-bindgen ABI types. `EvaluationContext` and `EnvBinding` also remain
+// here because ts-gen does not synthesize them.
```
guybedford, line 10: Why can't ts-gen synthesize `EvaluationContext`. isn't that what `FlagshipEvaluationContext` is?

It seems like we could support this in ts-gen if we had something along the lines of:

* first-class Record<K, V> lowering. Emit a #[wasm_bindgen(extends = Object)] newtype for the alias with new() plus setters derived from V — for a primitive union, per-variant set_string/set_number/set_bool (same machinery the dictionary builder already uses for fixed keys, just keyed by runtime &str). Then the generated methods take &FlagshipEvaluationContext and EvaluationContext in workers-rs is deleted. This is the direct fix for the "union erased to JsValue" warning too.
* keep getObjectValue<T extends object> in the .d.ts and let per-mono generics emit it as get_object_value<T: IntoWasmAbi + FromWasmAbi> / FlagshipEvaluationDetails<T>. T is a JS-side type (Object/JsValue), so the serde layer becomes a one-line serde_wasm_bindgen::from_value at the call site or a tiny generic helper — not four hand-rolled externs and a parallel EvaluationDetails<T> struct. One details type, not two.

PR #1058, opened 2026-09-08
Regenerate email bindings with per-mono generics

logan-gatlin:
## Summary

- Enable experimental per-monomorphization generics for both generated Email binding copies
- Regenerate Email APIs with `JsStringLike` parameters and `JsString` getter variants
- Advance `ts-gen` to include the typed-array dictionary-helper ABI fix

## Stack

- Depends on #1053, which adds the Flagship per-mono bindings

## Validation

- `cargo test -p worker --target wasm32-unknown-unknown`
- `cargo check -p worker-sys -p worker --target wasm32-unknown-unknown`
- Email and Flagship Vitest suites pass; full suite passed 150 tests with one unrelated R2 timeout that passed on targeted rerun

### DIFF
--- a/chompfile.toml
+++ b/chompfile.toml
@@ -18,6 +18,14 @@ deps = ['install:ts-gen']
 # implicit "every input is its own export" default is disabled.
 run = '''ts-gen --input types/email.d.ts --output worker/src/bindings/email.rs \
   --errors-as-error \
+  --experimental-generic-mono \
+  --export types/email.d.ts \
+  --export cloudflare:email \
+  --external "Env=crate::Env" \
+  --external "ExecutionContext=crate::Context"
+ts-gen --input types/email.d.ts --output worker-sys/src/types/email.rs \
+  --errors-as-error \
+  --experimental-generic-mono \
   --export types/email.d.ts \
   --export cloudflare:email \
   --external "Env=crate::Env" \
--- a/examples/receive-email/src/lib.rs
+++ b/examples/receive-email/src/lib.rs
@@ -44,7 +44,7 @@ async fn email(message: ForwardableEmailMessage, _env: Env, _ctx: Context) -> Re
         .write_to_string()
         .map_err(|e| Error::RustError(e.to_string()))?;
 
-    let reply = email::EmailMessage::new(&to, &from, &raw)?;
+    let reply = email::EmailMessage::new(to, from, raw)?;
     message.reply(&reply).await?;
     Ok(())
 }
--- a/worker/src/bindings/email.rs
+++ b/worker/src/bindings/email.rs
@@ -10,7 +10,7 @@ use ::web_sys::ReadableStream;
 use js_sys::*;
 #[allow(unused_imports)]
 use wasm_bindgen::prelude::*;
-#[wasm_bindgen]
+#[wasm_bindgen(experimental_generic_mono)]
 extern "C" {
     # [wasm_bindgen (extends = Event , extends = Object)]
     #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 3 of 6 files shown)


logan-gatlin:
Oh I thought ts-gen was used for way more than just email and flagship, I should probably just fold this into the first PR

chompfile.toml
```diff
@@ -18,6 +18,14 @@ deps = ['install:ts-gen']
 # implicit "every input is its own export" default is disabled.
 run = '''ts-gen --input types/email.d.ts --output worker/src/bindings/email.rs \
   --errors-as-error \
+  --experimental-generic-mono \
+  --export types/email.d.ts \
+  --export cloudflare:email \
+  --external "Env=crate::Env" \
+  --external "ExecutionContext=crate::Context"
+ts-gen --input types/email.d.ts --output worker-sys/src/types/email.rs \
```
guybedford, line 26: `worker-sys/src/types/email.rs` isn't wired into `worker-sys` at all (no `mod email;` in `worker-sys/src/types.rs`) — it's an orphan from #996. Delete the file instead of adding this second invocation.

examples/receive-email/src/lib.rs
```diff
@@ -44,7 +44,7 @@ async fn email(message: ForwardableEmailMessage, _env: Env, _ctx: Context) -> Re
         .write_to_string()
         .map_err(|e| Error::RustError(e.to_string()))?;
 
-    let reply = email::EmailMessage::new(&to, &from, &raw)?;
+    let reply = email::EmailMessage::new(to, from, raw)?;
```
guybedford, line 47: This change is required because `&String` no longer satisfies the generic bound — which is a source break for downstream users of the released `EmailMessage::new`. See the review summary: this needs `JsStringLike for &String` upstream in wasm-bindgen first, after which this diff (and the `send-email` / `test` ones) can be reverted.

worker/src/bindings/email.rs
```diff
@@ -115,31 +135,37 @@ extern "C" {
...
+    pub fn set_content_id<S: ::wasm_bindgen::JsStringLike>(this: &EmailAttachment, val: S);
     #[wasm_bindgen(method, getter)]
     pub fn filename(this: &EmailAttachment) -> String;
+    #[wasm_bindgen(method, getter, js_name = "filename")]
+    pub fn filename_js_string(this: &EmailAttachment) -> JsString;
     #[wasm_bindgen(method, setter)]
-    pub fn set_filename(this: &EmailAttachment, val: &str);
+    pub fn set_filename<S: ::wasm_bindgen::JsStringLike>(this: &EmailAttachment, val: S);
     #[wasm_bindgen(method, getter, js_name = "type")]
     pub fn type_(this: &EmailAttachment) -> String;
```
guybedford, line 160: Nit (ts-gen): `type__js_string` has a double underscore from `type_` + `_js_string`. The keyword-escape underscore should be stripped before suffixing so this becomes `type_js_string`.

guybedford (changes requested):
Verified locally: `chomp build:types` is a clean no-op on this commit (deterministic, and the ts-gen bump leaves `flagship.rs` unchanged), `send_email.spec.ts` + `flagship.spec.ts` pass, and the glue collapses `(&str, &str, String)` into a single `new EmailMessage(...)` import as expected.

Requesting changes for two things:

1. `worker-sys/src/types/email.rs` is a dead file — it is never declared (`worker-sys/src/types.rs` has no `mod email;`), it was left behind by #996. Rather than adding a second ts-gen invocation to keep regenerating it, delete the file and drop the second command.

2. This is the first per-mono regeneration of an already-released binding, and it breaks `&String` callers. `JsStringLike` is sealed to `String`, `&str`, `JsString`, `&JsString`; with a concrete `&str` parameter `&String` deref-coerced, through a generic bound it does not:

```
error[E0277]: the trait bound `&std::string::String: JsStringLike` is not satisfied
  --> EmailMessage::new(s, s, s)   // s: &String
```

The example diffs in this PR (`new(&to, &from, &raw)` -> `new(to, from, raw)`) are exactly that break, and `worker::EmailMessage::new`, `SendEmailBuilderBuilder::text(&body)` etc. are public API. This needs a wasm-bindgen PR adding `IntoWasmAbi` / `OptionIntoWasmAbi` / `JsStringLike` for `&String` (delegating to `&str`) before we regenerate existing bindings with per-mono generics, otherwise every binding we convert will hit the same regression. This PR should then land on top of that submodule bump with the example changes reverted.

PR #1060, opened 2026-09-08
feat(d1): add raw_with_column_names for raw({ columnNames: true })

### LINKED ISSUE #1052: [Feature] support .raw({columnNames: true}) for D1PreparedStatement

### Description

So there currently are bindings for calling `.raw()` on a prepared statement. 

https://github.com/cloudflare/workers-rs/blob/bb141710ebcf38b5a61c6e8f1d1c20c3dd2f3f6d/worker-sys/src/types/d1.rs#L98-L99

However, the option to fetch column names too via `.raw({columnNames: true})` is not supported right now.

https://developers.cloudflare.com/d1/worker-api/prepared-statements/#raw

It would be great for writing interfaces to existing ORMs like diesel.

I'm currently forking & updating an old diesel-d1 codebase I found (fork: https://github.com/hydrogen602/diesel-d1) and diesel indexes into a row via an `usize` so returning arrays of arrays are perfect for that.

mqmalagris:
Closes #1052

## Problem

`worker-sys` binds `D1PreparedStatement.raw()` with no arguments, so the `raw({ columnNames: true })` form of the D1 API is unreachable from Rust. `raw()` gives back rows as positional arrays with no way to learn what those positions mean.

That blocks the case in the issue: an ORM mapping layer (the reporter is working on a `diesel-d1` fork) indexes a row by `usize`, so arrays-of-arrays are the right shape — but it still needs the column names to map them.

## Fix

Two layers, both mirroring what's already there.

**`worker-sys/src/types/d1.rs`** — bind `raw` a second time, as `raw_with_options`, taking a `js_sys::Object`. A separate Rust name is required because wasm-bindgen can't overload on arity; `js_name=raw` keeps it pointing at the same JS method. The existing zero-arg `raw` binding is untouched.

**`worker/src/d1/mod.rs`** — add `raw_with_column_names<T>() -> Result<(Vec<String>, Vec<Vec<T>>)>`. It builds `{ columnNames: true }`, calls the new binding, then splits the header off the front.

The split is the reason this is a separate method rather than an option on `raw()`. D1 returns the column names as the *first array of the result* — typed upstream as `Promise<[string[], ...T[]]>` — so a single return value would be heterogeneous and force every caller to deal with a first row that isn't a row. Returning `(names, rows)` keeps rows uniformly `T`.

Also non-breaking: `raw()` and `raw_js_value()` keep their exact signatures and behavior.

## Test

Extended the existing `prepared_statement` integration test in `test/src/d1.rs`, next to the current `raw()` assertions:

- the same single-row query through `raw_with_column_names` yields `["id", "name", "age"]` plus one row whose positional values match what `raw()` returns;
- a query matching no rows still yields the column names, with an empty row set.

That second case is the one I would have got wrong by reading the code alone, so I checked it against a real runtime rather than assuming. With miniflare:

```
raw({columnNames:true}), 1 row  -> [["id","name","age"],[6,"Ryan Upton",21]]
raw({columnNames:true}), 0 rows -> [["id","name","age"]]
raw(),                   0 rows -> []
```

So the header is always present when `columnNames` is set, which matches the upstream tuple type. The implementation still guards the empty-array case, but as a defensive branch rather than an expected one, and the doc comment says the names are reported even with no rows.

Checks run: `cargo check -p worker -p worker-sandbox --target wasm32-unknown-unknown`, `cargo clippy -p worker -p worker-sys --target wasm32-unknown-unknown -- -D warnings` (clean), `cargo fmt --check` (clean).

I did not add a changeset — no `.changeset/*.md` has been filed per-PR in this repo's history, only the README and config. Happy to add one if that's wrong.

## Out of scope

`raw_js_value()` has no `columnNames` counterpart. It seemed better to add the one method the issue asks for than to speculatively double the surface; say the word if you'd like `raw_js_value_with_column_names` for symmetry.

### DIFF
--- a/test/src/d1.rs
+++ b/test/src/d1.rs
@@ -56,6 +56,24 @@ pub async fn prepared_statement(
     assert_eq!(columns[1].as_str(), Some("Ryan Upton"));
     assert_eq!(columns[2].as_u64(), Some(21));
 
+    // The same rows, plus the column names they are positioned by.
+    let (column_names, rows) = stmt.raw_with_column_names::<serde_json::Value>().await?;
+    assert_eq!(column_names, vec!["id", "name", "age"]);
+    assert_eq!(rows.len(), 1);
+    let columns = &rows[0];
+
+    assert_eq!(columns[0].as_u64(), Some(6));
+    assert_eq!(columns[1].as_str(), Some("Ryan Upton"));
+    assert_eq!(columns[2].as_u64(), Some(21));
+
+    // Column names are still reported when the statement matches no rows.
+    let (column_names, rows) = worker::query!(&db, "SELECT * FROM people WHERE name = ?")
+        .bind_refs(&D1Type::Text("Nobody At All"))?
+        .raw_with_column_names::<serde_json::Value>()
+        .await?;
+    assert_eq!(column_names, vec!["id", "name", "age"]);
+    assert!(rows.is_empty());
+
     let stmt_2 = unbound_stmt.bind_refs([&D1Type::Text("John Smith")])?;
     let person = stmt_2.first::<Person>(None).await?.unwrap();
     assert_eq!(person.name, "John Smith");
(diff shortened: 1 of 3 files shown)


guybedford:
Let's rather add the options object as typed, with getters/setters for props. See the patterns in `worker/src/bindings` folder for examples here.

mqmalagris:
@guybedford Got it! Done and pushed. D1PreparedStatementRawOptions follows the pattern.

PR #1062, opened 2026-09-14
fix: ai run json compatible serializer

### LINKED ISSUE #1050: [BUG] ai.run misbehaves with serde_json::Value, should probably use json_compatible serializer

0.8.5

4.118.0

### Describe the bug

Here, in workers-rs ai.run, it calls `serde_wasm_bindgen::to_value`.

https://github.com/cloudflare/workers-rs/blob/bb141710ebcf38b5a61c6e8f1d1c20c3dd2f3f6d/worker/src/ai.rs#L26

Unfortunately, if your input is or contains a serde_json::Value, those get converted to Javascript *Map* values, not objects.

The downstream javascript `ai.run` implementation does not seem to tolerate Map objects, and seems to just ignore them (it's kind of hard to tell what's going on but it looks like maybe it was treating them as empty objects).

I believe the workers-rs `ai.run` should be serializing its input using the [json_compatible](https://docs.rs/serde-wasm-bindgen/latest/serde_wasm_bindgen/struct.Serializer.html#method.json_compatible) serializer from serde_wasm_bindgen. This would allow it to pass on the objects appropriately to ai.run. I wouldn't be surprised if this affects multiple APIs in workers-rs.

### Steps To Reproduce

```rust
#[derive(Serialize)]
struct Tool {
    name: String,
    description: String,
    parameters: serde_json::Value,
}

#[derive(Serialize)]
struct Message {
    role: &'static str,
    content: &'static str,
}

#[derive(Serialize)]
struct ChatRequest {
    messages: Vec<Message>,
    tools: Vec<Tool>,
}

let tool = Tool {
    name: "CreateCreature".into(),
    description: "Create a creature".into(),
    parameters: serde_json::json!({
        "type": "object",
        "properties": {
            "creature": {
                "type": "object",
                "properties": {
                    "name": { "type": "string" },
                    "size": {
                        "type": "object",
                        "properties": {
                            "x": { "type": "integer" },
                            "y": { "type": "integer" },
                            "z": { "type": "integer" }
                        }
                    }
                }
            }
        }
    }),
};

let req = ChatRequest {
    messages: vec![Message {
        role: "user",
        content: "Create a creature named Goblin.",
    }],
    tools: vec![tool],
};

ai.run(model, req).await?;
```

When I run code like the following, it doesn't seem the model actually gets the content of the `parameters` in the input schema.

SameerVers3:
Fixes #1050

`ai.run` and `ai.run_bytes` were using the default `serde_wasm_bindgen` serializer, which turns json objects into js maps. since the workers ai runtime expects plain js objects, nested fields just got dropped and passed as an empty `{}`.

Here is what I did to fix this:

* patched it: switched to `Serializer::json_compatible()` so it produces plain js objects instead of maps.
* add a test: added a regression test (/serde-repro and serde_repro.spec.ts) to make sure it serializes to `Object`.
* [x] ran fmt
* [x] ran all the test
* [x] no breaking change


PeterMHammond:
Thanks for picking this up. We hit the same `Map` encoding with Workers AI and Vectorize and route around it with `serialize_maps_as_objects(true)`, so an upstream fix is welcome.

One suggestion: `json_compatible()` changes more than map encoding. In serde-wasm-bindgen 0.6.5 it is:

```rust
pub const fn json_compatible() -> Self {
    Self {
        serialize_missing_as_null: true,
        serialize_maps_as_objects: true,
        serialize_large_number_types_as_bigints: false,
        serialize_bytes_as_arrays: true,
    }
}
```

`serialize_missing_as_null: true` serializes every `Option::None` (and `()` / unit structs) as `null` instead of `undefined`. Today a `None` field without `skip_serializing_if` is simply absent once the input is stringified. After this PR it would be sent as an explicit `null`. That is a behavior change for existing callers, and a receiving schema can validate an absent key and an explicit `null` differently. `serialize_bytes_as_arrays: true` also changes how `serde_bytes` fields cross.

The minimal fix for #1050 is just the map flag:

```rust
input.serialize(&serde_wasm_bindgen::Serializer::new().serialize_maps_as_objects(true))?
```

For the broader sweep you mentioned: `Queue::send` / `send_batch` have the same issue. `From<T> for SendMessage<T>` defaults to `QueueContentType::Json`, and `into_raw_send_message` encodes with `serde_wasm_bindgen::to_value`, so a map field in a queue message arrives as `{}`. Durable Object storage and WebSocket attachments go through structured clone, so a `Map` survives there.

SameerVers3:
thanks for the context on the json_compatible() internals, but since the codebase already uses it in [`kv/builder.rs`](https://github.com/cloudflare/workers-rs/blob/main/worker/src/kv/builder.rs), I think it's worth keeping consistent here. although I am open to review again if maintainers have preference

PR #1065, opened 2026-09-15
worker-build: retry binary downloads on failure

op:
Downloads of tool binaries would fail the entire build on any transient network error. Retry the download up to 5 attempts with exponential backoff (2s, 4s, 8s, 16s), warning between attempts. Also truncate extracted files on write so a retry after a partial extraction cannot leave trailing garbage.


guybedford (approved):
Nice, thank you.

